PR #27, opened 2024-10-23
Build system for lind-wasm

yashaswi2000:
## Description

This introduces the practice of working branches and dev branches for have a reproducible build structure for jlind-wasm which uses
* wasmtime
* rawposix
* glib

all these git modules will have a stable main branch called lind-wasm-main which will contain the most latest compatible working version. and the build system will pull from this branch to build lind_wasm

* Adds a Makefile for build system
* Adds a updated Dockerfile for building lind_wasm docker images


JustinCappos:
I don't have enough context to give this a proper review.  However, having a simplified build system is a big need for the project.

PR #41, opened 2024-11-19
port rawposix wait PR

qianxichen233:
Ported from https://github.com/Lind-Project/RawPOSIX/pull/94


rennergade:
I think you mentioned you need to change a few lines in libc to patch that in. Now that we have the monorepo we can include that here.

rennergade (approved):

Issue #38, opened 2024-11-18
Support longjmp


qianxichen233:
It seems like bash is trying to use `longjmp` for syntax like `if`, `while`. `longjmp` is currently a missing feature in lind-wasm since the original implementation of `longjmp` in glibc is in asm language instead of c. So we probably want to implement this feature in lind-wasm.
Wasmer already has a support of `longjmp` using Asyncify so we can implement the same thing in wasmtime. The basic flow of longjmp goes as follows:
1. When a process calls `setjmp`, unwind the process, which will generate a unwind data. Save this data somewhere and rewind the process back.
2. When a process calls `longjmp`, unwind the process, then replace the unwind data with the saved unwind data from `setjmp` and do the rewind. This can make the process back to the previous state when `setjmp` is called.

PR #44, opened 2024-11-19
setjmp/longjmp implementation

qianxichen233:
This PR implement setjmp and longjmp within lind-wasm (issue https://github.com/Lind-Project/lind-wasm/issues/38)


JustinCappos (commented):
Can you add some basic tests with this addition?   Even just the simple example from:  https://en.cppreference.com/w/c/program/setjmp would be helpful.

qianxichen233:
> Can you add some basic tests with this addition? Even just the simple example from: https://en.cppreference.com/w/c/program/setjmp would be helpful.

Added a longjmp testcase. The longjmp is also already tested againest bash and bash is able to run things like `if` upon it

src/glibc/target/include/bits/setjmp.h
```diff
@@ -27,13 +27,7 @@
 
 #ifndef _ASM
 
-# if __WORDSIZE == 64
```
JustinCappos, line 30: Why is this change made?   Shouldn't you at least assert the __WORDSIZE is set as expected?

JustinCappos: Okay, this is all fine, but please explain this in comments in the code, so that this is clear to us ~3 months / years from now...   We will stumble across this code later and wonder why it was done this way in many cases.

src/RawPOSIX/RawPOSIX
```diff
@@ -0,0 +1 @@
+../../RawPOSIX
```
rennergade, line 1: what is this change?

qianxichen233: Deleted. This is something I accidentally created and committed

src/wasmtime/crates/rawposix
```diff
@@ -1 +1 @@
-../../RawPOSIX
\ No newline at end of file
+../../RawPOSIX/
```
rennergade, line 1: also this?

qianxichen233: This is a symlink. The RawPOSIX is currently soft-linked under crates of wasmtime

rennergade (approved):

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -1110,6 +1110,331 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
...
+
+            if let Some(unwind_data) = data {
+                // replace the unwind data
+                unsafe { std::ptr::copy_nonoverlapping(unwind_data.as_ptr(), rewind_start_parent, unwind_data.len()); }
+            } else {
+                // if the hash does not exist
+                // according to standard, calling longjmp with invalid jmp_buf would
+                // cause undefined behavior and may lead to crash. Since invalid jmp_buf is not able to be detected,
+                // so it will not return any kind of error.
+                // However, our approach of using Asyncify to implement longjmp do be able to detect it
```
rennergade, line 1416: "is able to detect it"

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -1270,6 +1595,32 @@ pub fn exit_syscall<T: LindHost<T, U> + Clone + Send + 'static + std::marker::Sy
     0
 }
 
+pub fn setjmp_syscall<T: LindHost<T, U> + Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static + std::marker::Sync>
```
rennergade, line 1598: neither of these are syscalls

src/wasmtime/crates/wasmtime/src/runtime/store.rs
```diff
@@ -1326,6 +1336,38 @@ impl<'a, T> StoreContextMut<'a, T> {
         self.0.rewinding = state;
     }
 
+    // store the unwind data, return its hash
+    pub fn store_unwind_data(&mut self, ptr: *const u8, len: usize) -> u64 {
+        // Allocate a vector with enough capacity to hold the data.
```
yzhang71, line 1341: Do we need to check the ptr is not null here?

qianxichen233: This is invoked by wasmtime and ptr is something created by wasmtime instead of a user-passed argument. ptr is not supposed to be zero since it is basically the base_address of the liear memory adding some potential offsets.

rennergade: Leaving that explanation in a comment would be nice

yzhang71 (approved):

rennergade (approved):

PR #64, opened 2024-12-10
Mod: Fixing issue #35

yzhang71:
This PR resolves [issue #35](https://github.com/Lind-Project/lind-wasm/issues/35). Only the `crt1.c` file has been modified.

### DIFF
--- a/src/glibc/lind_syscall/crt1/crt1.c
+++ b/src/glibc/lind_syscall/crt1/crt1.c
@@ -2,6 +2,9 @@
 #include <sysexits.h>
 
 extern char** environ;
+// environ is a global variable that holds the environment variables for a program.
+// It is typically an array of strings, where each string represents an environment variable in the format KEY=VALUE.
+// The environ variable is used to store and access environment variables passed to the program by the runtime or operating system.
 
 static char *empty_environ[1] = { NULL };
 
@@ -66,6 +69,8 @@ void __wasi_initialize_environ(void) {
     size_t environ_count;
     size_t environ_buf_size;
     __wasi_environ_sizes_get(&environ_count, &environ_buf_size);
+    // The __wasi_environ_sizes_get function retrieves the number of environment variables (environ_count) 
+    // and the total size of their contents (environ_buf_size).
     
     if (environ_count == 0) {
         environ = empty_environ;
@@ -96,6 +101,10 @@ void __wasi_initialize_environ(void) {
     // pointers into those chars.
     // TODO: Remove the casts on `environ_ptrs` and `environ_buf` once the witx is updated with char8 support.
     __wasi_environ_get((unsigned char **)environ_ptrs, (unsigned char *)environ_buf);
+    // The implementation relies on the WASI (WebAssembly System Interface) API functions, 
+    // such as __wasi_environ_sizes_get and __wasi_environ_get, to fetch and initialize environment variables.
+    // This is necessary because WebAssembly doesn't inherently provide a standard way to access environment variables, 
+    // so WASI serves as an abstraction for this purpose.
 
     environ = environ_ptrs;
     return;
@@ -124,7 +133,14 @@ int main(int argc, char *argv[], char *envp[]);
 // See also the TODO comment on `__main_void` below.
 __attribute__((__weak__))
 int __main_argc_argv(int argc, char *argv[]) {
+  // This is an internal function designed to handle the program's main function when it is 
+  // expected to receive arguments (argc, argv).
+  // It serves as a wrapper around the user-defined main function.
+  
   return main(argc, argv, environ);
+  // This wrapper function allows the runtime system to call main with the correct arguments 
+  // in an environment where the main function's signature follows the standard int main(int argc, char *argv[], char *envp[]).
+  // If the user does not define their own main function, this weakly defined function can serve as a default implementation.
 }
 
 // If the user's `main` function expects arguments, the compiler will rename
--- a/src/glibc/lind_syscall/crt1/crt1.c
+++ b/src/glibc/lind_syscall/crt1/crt1.c
@@ -105,13 +105,6 @@ software:
     _Exit(EX_SOFTWARE);
 }
 
-int _start() {
-    __libc_setup_tls();
-    __wasi_init_tp();
-    __wasi_initialize_environ();
-    return __main_void();
-}
-
 void __wasm_call_dtors() {
     
 }
@@ -120,6 +113,9 @@ void __wasi_proc_exit(unsigned int exit_code) {
     
 }
 
+__attribute__((__weak__))
+int main(int argc, char *argv[], char *envp[]);
+
 // The user's `main` function, expecting arguments.
 //
 // Note that we make this a weak symbol so that it will have a
@@ -127,7 +123,9 @@ void __wasi_proc_exit(unsigned int exit_code) {
 // it need not be defined (e.g. in reactor-style apps with no main function).
 // See also the TODO comment on `__main_void` below.
 __attribute__((__weak__))
-int __main_argc_argv(int argc, char *argv[]);
+int __main_argc_argv(int argc, char *argv[]) {
+  return main(argc, argv, environ);
+}
 
 // If the user's `main` function expects arguments, the compiler will rename
 // it to `__main_argc_argv`, and this version will get linked in, which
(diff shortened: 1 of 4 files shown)


rennergade:
Looks good!

1. Can you add a few notes about how "environ" works. That's not necessarily clear to me and would be helpful later.
2. If you have a test case to add with this that would be useful.

Yaxuan-w (approved):
LGTM! My understanding is that this change enables the usage of environment variables, which are recognized by Clang during the compilation phase. Please let me know if my understanding is incorrect.

qianxichen233:
Have you tried to compile bash again, to verify if bash is able to run with main function with argc, argv and env?

qianxichen233:
There are three versions of main we declared here (main with no argument, main with argc and argv, and main with argc, argv and env). The way crt1 works in terms of redirecting user's main to the correct version of main implementation is mainly though weak functoin overriding. First, we have _start function always calling main_void function, while main_void is a weak function. And in main_void function, it actually sets up argc and argv and calls main_argc_argv, which is also a weak function. Then, for example, if user is trying main with no argument, that would override main_void in crt1 file because main_void here is a weak function, and _start will directly calls user's main. If user's main function is main with argc and argv, then main_void will be perserved, while main_argc_argv will be overriden. Then the logics would become something like _start->main_void->main_argc_argv(overriden by user's main). We further expanded the chain to add an main version with argc, argv and env at the end of main_argc_argv so this version of main is also supported right now.
However, there is still minor issues. So for the main version that uses argc and argv, we can just use `_main_argc_argv` as the name of the entry since clang will automatically rename `main` function with argc and argv to `_main_argc_argv` (https://github.com/llvm/llvm-project/blob/c7634c1b610e9abe13646c5936dd14402e0dc2f3/clang/lib/AST/Mangle.cpp#L180). But it looks like there is no support of renaming for main version with argc, argv and env. So in order to support this, we have to use the original name of the main (i.e. iterally `main`) for main version of argc, argv and env. This would cause clang to give a warming when compiling main version with no argument:
```
wasm-ld: warning: function signature mismatch: main
>>> defined as (i32, i32, i32) -> i32 in /home/lind-wasm/src/glibc/sysroot/lib/wasm32-wasi/crt1.o
>>> defined as (i32, i32) -> i32 in /tmp/tmp_void-3561c1.o
```
This warming does not seem to affect the usability and the this version of main can still run for some reason, and I checked its compiled wasm binary and found out the linking also looks fine. But it would probably be hard to completely get rid of this warming because clang does not have support for renaming this version of main.

yzhang71:
> Have you tried to compile bash again, to verify if bash is able to run with main function with argc, argv and env?

Good point! I have tried to compile bash again and it works fine.

rennergade:
> There are three versions of main we declared here (main with no argument, main with argc and argv, and main with argc, argv and env). The way crt1 works in terms of redirecting user's main to the correct version of main implementation is mainly though weak functoin overriding. First, we have _start function always calling main_void function, while main_void is a weak function. And in main_void function, it actually sets up argc and argv and calls main_argc_argv, which is also a weak function. Then, for example, if user is trying main with no argument, that would override main_void in crt1 file because main_void here is a weak function, and _start will directly calls user's main. If user's main function is main with argc and argv, then main_void will be perserved, while main_argc_argv will be overriden. Then the logics would become something like _start->main_void->main_argc_argv(overriden by user's main). We further expanded the chain to add an main version with argc, argv and env at the end of main_argc_argv so this version of main is also supported right now. However, there is still minor issues. So for the main version that uses argc and argv, we can just use `_main_argc_argv` as the name of the entry since clang will automatically rename `main` function with argc and argv to `_main_argc_argv` (https://github.com/llvm/llvm-project/blob/c7634c1b610e9abe13646c5936dd14402e0dc2f3/clang/lib/AST/Mangle.cpp#L180). But it looks like there is no support of renaming for main version with argc, argv and env. So in order to support this, we have to use the original name of the main (i.e. iterally `main`) for main version of argc, argv and env. This would cause clang to give a warming when compiling main version with no argument:
> 
> ```
> wasm-ld: warning: function signature mismatch: main
> >>> defined as (i32, i32, i32) -> i32 in /home/lind-wasm/src/glibc/sysroot/lib/wasm32-wasi/crt1.o
> >>> defined as (i32, i32) -> i32 in /tmp/tmp_void-3561c1.o
> ```
> 
> This warming does not seem to affect the usability and the this version of main can still run for some reason, and I checked its compiled wasm binary and found out the linking also looks fine. But it would probably be hard to completely get rid of this warming because clang does not have support for renaming this version of main.

We should definitely clean up this warning. Clang can compile libc normally right? How is it handled in the crt files for a normal compilation.

qianxichen233:
> > There are three versions of main we declared here (main with no argument, main with argc and argv, and main with argc, argv and env). The way crt1 works in terms of redirecting user's main to the correct version of main implementation is mainly though weak functoin overriding. First, we have _start function always calling main_void function, while main_void is a weak function. And in main_void function, it actually sets up argc and argv and calls main_argc_argv, which is also a weak function. Then, for example, if user is trying main with no argument, that would override main_void in crt1 file because main_void here is a weak function, and _start will directly calls user's main. If user's main function is main with argc and argv, then main_void will be perserved, while main_argc_argv will be overriden. Then the logics would become something like _start->main_void->main_argc_argv(overriden by user's main). We further expanded the chain to add an main version with argc, argv and env at the end of main_argc_argv so this version of main is also supported right now. However, there is still minor issues. So for the main version that uses argc and argv, we can just use `_main_argc_argv` as the name of the entry since clang will automatically rename `main` function with argc and argv to `_main_argc_argv` (https://github.com/llvm/llvm-project/blob/c7634c1b610e9abe13646c5936dd14402e0dc2f3/clang/lib/AST/Mangle.cpp#L180). But it looks like there is no support of renaming for main version with argc, argv and env. So in order to support this, we have to use the original name of the main (i.e. iterally `main`) for main version of argc, argv and env. This would cause clang to give a warming when compiling main version with no argument:
> > ```
> > wasm-ld: warning: function signature mismatch: main
> > >>> defined as (i32, i32, i32) -> i32 in /home/lind-wasm/src/glibc/sysroot/lib/wasm32-wasi/crt1.o
> > >>> defined as (i32, i32) -> i32 in /tmp/tmp_void-3561c1.o
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This warming does not seem to affect the usability and the this version of main can still run for some reason, and I checked its compiled wasm binary and found out the linking also looks fine. But it would probably be hard to completely get rid of this warming because clang does not have support for renaming this version of main.
> 
> We should definitely clean up this warning. Clang can compile libc normally right? How is it handled in the crt files for a normal compilation.

The issue is that Clang does not have the support to rename main(argc, argv, env) in wasm. For main(argc, argv) in wasm environment, clang will detect it and change the name of the main function to be `_main_argc_argv` to avoid duplicated name. I guess we should be able to modify the clang source code to let it also automatically rename main(argc, argv, env) to something like `_main_argc_argv_env` to get rid of the warming, and that should be only a few lines of addition. But modifying Clang would definitely add a huge overhead to our build process

rennergade:
Ah I see. Yeah we certainly don't want to do that.

Perhaps we're addressing this incorrectly. It seems like wasm/wasmtime has the ability to set environment variables. Can we set those variables before we call main in wasmtime, and then just call execv instead of execve?

JustinCappos:
Just catching up on this thread.   If this is just a warning but everything works, let's leave this for now and create an issue for it.   We can come back to it later.

qianxichen233:
> Ah I see. Yeah we certainly don't want to do that.
> 
> Perhaps we're addressing this incorrectly. It seems like wasm/wasmtime has the ability to set environment variables. Can we set those variables before we call main in wasmtime, and then just call execv instead of execve?

This is not an issue on environment variable creation, we already did this and it works well. It's just we need another version of main that explicitly pass env as one of its argument (i.e. main(argc, argv, env)). We always have an environ global that stores all the environment variables no matter which version of main is used.

yzhang71:
> Just catching up on this thread. If this is just a warning but everything works, let's leave this for now and create an issue for it. We can come back to it later.

Yes, currently it's just a warning, and everything else works fine. Fixing it is not difficult, as Qianxi mentioned above, but we would need a customized version of LLVM/Clang.

yzhang71:
> Looks good!
> 
> 1. Can you add a few notes about how "environ" works. That's not necessarily clear to me and would be helpful later.
> 2. If you have a test case to add with this that would be useful.

Good point! More comments are added, as well as the test-cases.

rennergade:
Great comments and tests. Think this is ready to merge.

Issue #32, opened 2024-11-14
Fix pthread


qianxichen233:
Currently pthread is using clone_syscall, which is implemented using Asyncify. There is an issue on current approach using Asyncify to implement pthread (i.e. unwind/rewind two wasm within same memory at the same time): The variable to control the unwind/rewind state is a global variable that is shared by parent and child threads, which means they will compete for the unwind/rewind state. I am actually pretty surprised that the pthread could work kind stably without triggering the issue previously, until I finally observed the issue.
To resolve the issue might be a little tricky, there are several ways I currently can think of to resolve the issue:
1. Try to syncronize the parent and child thread on asyncify state (i.e. use barrier to make parent and child always at the same asyncify step). This may add additional overhead on pthread creation and could still potentially have issues on other aspects.
2. Try to make the asyncify state global variable be non-shared for parent and child: create a non-shared memory region for threading that is used to store some per-thread global variables and make the asyncify state global variable sits in there. Since we now have vmmap, this approach shouldn't be hard to implement, but I am not sure if that aligns with our memory management architecture.
3. Fallback to wasi-thead implementation of pthread

rennergade:
Let's also fix epoch interruption to work with multi-processing here. Should be able to give us much more control for exiting.

rennergade:
Closing this issue and creating a separate issue for epoch_interruption.

PR #56, opened 2024-11-25
Porting VMMAP to RawPOSIX

ChinmayShringi:
Please note this is taken from RawPosix: (https://github.com/Lind-Project/RawPOSIX/pull/87)
## Description

Issue # 45

 

### Type of change

Porting pranav-bhatt and ruchjoshi-nyu implementation of VMMAP to RawPOSIX. Request review for porting location and if more comments are needed. 

I also added few comments to make things more clear. That would be great if ruchjoshi-nyu could review their correctness.

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been added to a pull request and/or merged in other modules (native-client, lind-glibc, lind-project)

PR #67, opened 2024-12-24
Fix pthread

qianxichen233:
This PR fixed issue https://github.com/Lind-Project/lind-wasm/issues/32, and did some clean up and re-structure to remove some warnings and make the code more readable.

The issue was due to incorrectly handle of unwind data storage. Previously all the unwind data was stored directly at the top of the memory (offset 0). It turns out this is the place where global variables store and storing unwind data here would overwrite the data here. This is fixed by adding an option to wasm-ld to let it output the stack_low global, which represents the top of the stack. This address can then be used to store the unwind data and should not have any conflict with other data.

There is a new issue I encountered while fixing pthread. I added a chained thread test to the test list, which will currently fail because it seems like when recursively creating threads, glibc will try to use a slightly different routine to handle, which calls mmap at some place. The mmap causes the recursive thread creation fail because vmmap PR(https://github.com/Lind-Project/lind-wasm/pull/56) hasn't been merged into main yet, so mmap is still a broken feature in this branch. Hopefully once we get mmap finalized and merged into main, this issue will go away.


src/wasmtime/src/commands/run.rs
```diff
@@ -207,10 +207,10 @@ impl RunCommand {
                 // exit the cage
                 lind_syscall_api(
                     1,
-                    30 as u32,
+                    30,
```
rennergade, line 210: if we're calling syscalls directly like this this should be at least a comment saying what syscall but more appropriately we should have some defines here

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -21,6 +21,8 @@ const ASYNCIFY_STOP_UNWIND: &str = "asyncify_stop_unwind";
 const ASYNCIFY_START_REWIND: &str = "asyncify_start_rewind";
 const ASYNCIFY_STOP_REWIND: &str = "asyncify_stop_rewind";
 
+const LIND_FS_ROOT: &str = "/home/lind-wasm/src/RawPOSIX/tmp";
```
rennergade, line 24: should root be tmp here?

qianxichen233: This is the root set in RawPOSIX
https://github.com/Lind-Project/lind-wasm/blob/57876e5c51e78c9285fa6842d5e1efe5f15c745f/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs#L31
I just used the same path here

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -231,97 +212,50 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
...
+        // -------------------------- <----- unwind_data_end (user's current stack pointer)
+        // |         .....          | ^
+        // |         .....          | |
+        // |       stack data       | | user's stack grow direction
+        // |         .....          | |
+        // |         .....          | |
+        // -------------------------- <----- stack high
+        unsafe {
+            // 16 because it is the size of two u64
+            *(unwind_data_start_sys as *mut u64) = unwind_data_start_usr + 16;
```
rennergade, line 248: again defines are better than magic numbers

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -331,14 +265,14 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
...
-            println!("running out of cageid!");
+            panic!("running out of cageid!");
         }
         let child_cageid = child_cageid.unwrap();
 
         // calling fork in rawposix to fork the cage
         lind_syscall_api(
             self.pid as u64,
-            68 as u32, // fork syscall
+            68, // fork syscall
```
rennergade, line 275: also better as a define

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -468,7 +399,7 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
                             // exit the cage with the exit code
                             lind_syscall_api(
                                 child_cageid,
-                                30 as u32,
+                                30,
```
rennergade, line 402: same

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -526,120 +457,49 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
...
-        else {
-            println!("asyncify_start_unwind export not found");
-            return Ok(-1);
+        let asyncify_start_unwind_func = caller.get_asyncify_start_unwind().unwrap();
+
+        // store the parameter at the top of the stack
+        // reference comments in fork_call
+        unsafe {
+            // 16 because it is the size of two u64
+            *(parent_unwind_data_start_sys as *mut u64) = parent_unwind_data_start_usr + 16;
```
rennergade, line 478: magic number

src/wasmtime/crates/lind-multi-process/src/lib.rs
Yaxuan-w: im curious that why we need using lind fs in wasmtime..?

qianxichen233: it is used in execve. The path passed in here should be within lind filesystem

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -231,97 +212,50 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
...
-            return Ok(-1);
+        let asyncify_start_unwind_func = caller.get_asyncify_start_unwind().unwrap();
+
+        // store the parameter at the top of the stack
+        // we need to tell two parameters to Asyncify:
+        // unwind_data_start: the start address to store the unwind data.
+        // unwind_data_end: the end address of the avaliable space Asyncify could work with
+        // These two parameters are usually stored on the top of the unwind data.
+        // Below is a graph describing the entire user's stack layout
+        // -------------------------- <----- _usr (stack low)
```
rennergade, line 231: i like this diagram a lot as we discussed, its still a bit confusing that unwind_data_start and end are pointers stored at that address, perhaps we can make that a bit more clear by not depicting them in large regions

rennergade:
This should be ready to merge with the minor change requests

rennergade (approved):
Looks good. Think this is ready to merge.

PR #69, opened 2024-12-24
add wasmtest.sh

robinyuan1002:
There are mainly three functions for this script(wasmtest.sh).
1.there opinion read from a file said which test to run
2.give specific test to run
3.run everything
For each function can compile and run together or run separately(depend on what user need).

### DIFF
--- a/wasmtest.sh
+++ b/wasmtest.sh
@@ -1,4 +1,17 @@
 #!/bin/bash
+
+# wasmtest.sh is a shell script that can compile and run wasm tests.
+# There are three funtions for wasmteest.sh(1.test single file 2.test all files 3.test all tests from the file user give)
+# To test single file use: ./wasmtest.sh single <your file name>.c 
+# To compile single file use: ./wasmtest.sh singlecompile <your file name>.c
+# To run single file use: ./wasmtest.sh singlecompile <your file name>.c
+# To test all files use: ./wasmtest.sh all
+# To compile all files use: ./wasmtest.sh allcompile
+# To run all files use: ./wasmtest.sh allrun
+# To test all tests from the file user give use: ./wasmtest.sh files
+# To compile all tests from the file user give use: ./wasmtest.sh filescompile
+# To run all tests from the file user give use: ./wasmtest.sh filesrun
+
 glibc_base="/home/lind-wasm/src/glibc"
 wasmtime_base="/home/lind-wasm/src/wasmtime"
 
@@ -12,10 +25,12 @@ run_cmd_precompile="$wasmtime_base/target/debug/wasmtime run --allow-precompiled
 
 test_file_base="/home/lind-wasm/tests/unit-tests"
 
+#color codes for terminal output
 RED='\033[31m'
 GREEN='\033[32m'
 RESET='\033[0m'
 
+#function to compile a single test file
 compile_single_test() {
     local test_name="$1"
 
@@ -46,6 +61,7 @@ compile_single_test() {
     fi
 }
 
+#function to run a single test file
 run_single_test() {
     local test_name="$1"
 
@@ -84,6 +100,7 @@ run_single_test() {
     echo
 }
 
+#function to compile all the .c test files in the unit-tests folder
 compile_all_tests() {
     echo -e "${GREEN}Compiling all test cases in $test_file_base...${RESET}"
 
@@ -99,6 +116,7 @@ compile_all_tests() {
     echo -e "${GREEN}All tests compiled.${RESET}"
 }
 
+#function to run all the .c test files in the unit-tests folder
 run_all_tests() {
     echo -e "${GREEN}Running all test cases in $test_file_base...${RESET}"
 
@@ -114,6 +132,7 @@ run_all_tests() {
     echo -e "${GREEN}All tests completed.${RESET}"
 }
 
+#function to compile all the files from the file user give
 compile_from_files() {
     local file_list="$1"
 
@@ -137,6 +156,7 @@ compile_from_files() {
     echo -e "${GREEN}All specified tests compiled.${RESET}"
 }
 
+#function to run all the files from the file user give
 run_from_files() {
     local file_list="$1"
 
@@ -234,4 +254,3 @@ case "$1" in
         run_all_tests
         ;;
 esac
-
--- a/wasmtest.sh
+++ b/wasmtest.sh
@@ -1,0 +1,237 @@
+#!/bin/bash
+glibc_base="/home/lind-wasm/src/glibc"
+wasmtime_base="/home/lind-wasm/src/wasmtime"
+
+CC="${CLANG:=/home/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/clang"
+
+export_cmd="export LD_LIBRARY_PATH=$wasmtime_base/crates/rustposix:\$LD_LIBRARY_PATH"
+
+compile_test_cmd_fork_test="$CC -pthread --target=wasm32-unknown-wasi --sysroot $glibc_base/sysroot -Wl,--import-memory,--export-memory,--max-memory=67108864,--export="__stack_pointer" [input] -g -O0 -o [output] && wasm-opt --asyncify --debuginfo [output] -o [output]"
+precompile_wasm="$wasmtime_base/target/debug/wasmtime compile [input] -o [output]"
+run_cmd_precompile="$wasmtime_base/target/debug/wasmtime run --allow-precompiled --wasi threads=y --wasi preview2=n [target]"
+
+test_file_base="/home/lind-wasm/tests/unit-tests"
+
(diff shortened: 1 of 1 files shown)


wasmtest.sh
```diff
@@ -0,0 +1,237 @@
...
+
+compile_test_cmd_fork_test="$CC -pthread --target=wasm32-unknown-wasi --sysroot $glibc_base/sysroot -Wl,--import-memory,--export-memory,--max-memory=67108864,--export="__stack_pointer" [input] -g -O0 -o [output] && wasm-opt --asyncify --debuginfo [output] -o [output]"
+precompile_wasm="$wasmtime_base/target/debug/wasmtime compile [input] -o [output]"
+run_cmd_precompile="$wasmtime_base/target/debug/wasmtime run --allow-precompiled --wasi threads=y --wasi preview2=n [target]"
+
+test_file_base="/home/lind-wasm/tests/unit-tests"
+
+RED='\033[31m'
+GREEN='\033[32m'
+RESET='\033[0m'
```
Yaxuan-w, line 17: That would be great to have some comments explaining those variables' usage

wasmtest.sh
```diff
@@ -0,0 +1,237 @@
...
+
+                run_single_test "$test_dir/$test_name"
+            done
+        fi
+    done < "$file_list"
+
+    echo -e "${GREEN}All tests from file list completed.${RESET}"
+}
+
+pmode=0
```
Yaxuan-w, line 166: Why we need this argument...?

wasmtest.sh
```diff
@@ -0,0 +1,237 @@
+#!/bin/bash
```
Yaxuan-w, line 1: That would be super helpful to add usage instruction as comments at the beginning of this file

Yaxuan-w (changes requested):
Great work! More comments requested

Yaxuan-w (approved):
LGTM

wasmtest.sh
```diff
@@ -0,0 +1,256 @@
...
+# To run all tests from the file user give use: ./wasmtest.sh filesrun
+
+glibc_base="/home/lind-wasm/src/glibc"
+wasmtime_base="/home/lind-wasm/src/wasmtime"
+
+CC="${CLANG:=/home/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/clang"
+
+export_cmd="export LD_LIBRARY_PATH=$wasmtime_base/crates/rustposix:\$LD_LIBRARY_PATH"
+
+compile_test_cmd_fork_test="$CC -pthread --target=wasm32-unknown-wasi --sysroot $glibc_base/sysroot -Wl,--import-memory,--export-memory,--max-memory=67108864,--export="__stack_pointer" [input] -g -O0 -o [output] && wasm-opt --asyncify --debuginfo [output] -o [output]"
```
qianxichen233, line 22: Change this to `compile_test_cmd_fork_test="$CC -pthread --target=wasm32-unknown-wasi --sysroot $glibc_base/sysroot -Wl,--import-memory,--export-memory,--max-memory=67108864,--export="__stack_pointer",--export=__stack_low [input] -g -O0 -o [output] && wasm-opt --asyncify --debuginfo [output] -o [output]"`
The `--export=__stack_low` added here is from current pthread fix PR(https://github.com/Lind-Project/lind-wasm/pull/67). Let's also add it for this file in this PR.

wasmtest.sh
```diff
@@ -0,0 +1,256 @@
...
+# To test all tests from the file user give use: ./wasmtest.sh files
+# To compile all tests from the file user give use: ./wasmtest.sh filescompile
+# To run all tests from the file user give use: ./wasmtest.sh filesrun
+
+glibc_base="/home/lind-wasm/src/glibc"
+wasmtime_base="/home/lind-wasm/src/wasmtime"
+
+CC="${CLANG:=/home/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/clang"
+
+export_cmd="export LD_LIBRARY_PATH=$wasmtime_base/crates/rustposix:\$LD_LIBRARY_PATH"
```
qianxichen233, line 20: This is no longer necessary. We can delete it

wasmtest.sh
```diff
@@ -0,0 +1,256 @@
...
+        echo -e "${RED}Error: Source file ${test_name} not found in $test_file_base.${RESET}"
+        exit 1
+    fi
+
+    output_dir=$(dirname "$source_file")
+    without_c="${test_name%??}"
+    output_file_wasm="${output_dir}/${twithout_c}.wasm"
+    output_file_cwasm="${output_dir}/${without_c}.cwasm"
+
+    eval $export_cmd
```
qianxichen233, line 49: no longer required

wasmtest.sh
```diff
@@ -0,0 +1,256 @@
...
+    pmode=1
+
+    # Remove the last argument
+    new_args=("${@:1:$#-1}")
+    # Overwrite the positional parameters with the new arguments
+    set -- "${new_args[@]}"
+fi
+
+case "$1" in
+    filescompile)
```
qianxichen233, line 197: I think it would be a good practice to avoid confusion between singular and plural forms in command-line argument. I believe if you use `filescompile` here, people will sometimes mistakenly type `filecompile` instead and get confused. I believe have both singular and plural form of the command supported is the most user-friendly approach. For example, change this to `filescompile|filecompile`. This applies all other commands below. Besides, have an abbreviated form of the command allowed is also something I prefer.

wasmtest.sh
```diff
@@ -0,0 +1,256 @@
...
+# To run single file use: ./wasmtest.sh singlecompile <your file name>.c
+# To test all files use: ./wasmtest.sh all
+# To compile all files use: ./wasmtest.sh allcompile
+# To run all files use: ./wasmtest.sh allrun
+# To test all tests from the file user give use: ./wasmtest.sh files
+# To compile all tests from the file user give use: ./wasmtest.sh filescompile
+# To run all tests from the file user give use: ./wasmtest.sh filesrun
+
+glibc_base="/home/lind-wasm/src/glibc"
+wasmtime_base="/home/lind-wasm/src/wasmtime"
```
qianxichen233, line 16: I'd prefer to have a LIND_WASM_BASE variable to add flexibility to the path. For example, the LIND_WASM_BASE variable has a default value of `/home/lind-wasm`, but it can be overwriten by environment variable

wasmtest.sh
```diff
@@ -0,0 +1,256 @@
...
+    else
+        final_cmd=$(echo "$run_cmd" | sed "s|\[target\]|$target_wasm|")
+        final_cmd="${final_cmd} $@"
+    fi
+
+    eval $export_cmd
+
+    echo -e "${GREEN}Running: $final_cmd${RESET}"
+
+    if timeout 10s bash -c "$final_cmd"; then
```
qianxichen233, line 91: It might be better to make `10s` be a configurable value. By default it is 10 second, but we can support one more optional command line option to customize this value

qianxichen233 (changes requested):
Another thing I prefer is to remove the auto-completion of extension names. This feature is something I added when I wrote lindtool, but later it turns out this is pretty annoying. For example, some automatically generated binary (e.g. bash) might probably does not have extension at all, this makes the script not even be able to run it because the script will always spontaneously add an extension name.

Yaxuan-w:
> Another thing I prefer is to remove the auto-completion of extension names. This feature is something I added when I wrote lindtool, but later it turns out this is pretty annoying. For example, some automatically generated binary (e.g. bash) might probably does not have extension at all, this makes the script not even be able to run it because the script will always spontaneously add an extension name.

I'd like to be know more here, this file is used only for general test suite right? Why we need to compile bash here

qianxichen233:
> > Another thing I prefer is to remove the auto-completion of extension names. This feature is something I added when I wrote lindtool, but later it turns out this is pretty annoying. For example, some automatically generated binary (e.g. bash) might probably does not have extension at all, this makes the script not even be able to run it because the script will always spontaneously add an extension name.
> 
> I'd like to be know more here, this file is used only for general test suite right? Why we need to compile bash here

I am actually not 100% sure about our test plan. But don't we need to run bash test or other application level tests?

rennergade:
> > > Another thing I prefer is to remove the auto-completion of extension names. This feature is something I added when I wrote lindtool, but later it turns out this is pretty annoying. For example, some automatically generated binary (e.g. bash) might probably does not have extension at all, this makes the script not even be able to run it because the script will always spontaneously add an extension name.
> > 
> > 
> > I'd like to be know more here, this file is used only for general test suite right? Why we need to compile bash here
> 
> I am actually not 100% sure about our test plan. But don't we need to run bash test or other application level tests?


We will but the first hurdle here is just getting a script to Yash so we can compile and run the unit tests automatically.

qianxichen233 (approved):

rennergade (approved):

PR #74, opened 2024-12-30
mmap/sbrk/brk implementation and integration into glibc/wasmtime

qianxichen233:
This PR finalized the implementation of mmap/munmap/sbrk/brk.

- On rawposix side, the changes are mainly
    1. Finishing the remaining program break implementation and file backed mmap feature
    2. Minor changes in vmmap to allow the pre-specified start and end address when find map space
    3. Minor adjustment on the vmmap initialization functions invoked from wasmtime.

- On wasmtime side, the changes are mainly about bypassing part of wasmtime's internal memory management logic, including:
    1. Disabled wasmtime's sys mmap function, as this is handled by rawposix right now.
    2. Have wasmtime always creating a length of 4294967296 bytes of accessible region, so that some of the wasm's instruction handled by wasmtime will not be rejected due to memory address.
    3. Restructured memory initialization code interacted with rawposix.

- On glibc side, the changes are mainly about
    1. Modified glibc's entry point of sbrk/brk syscall so that it directs to rawposix
    2. Revert pthread's memory allocation from malloc (old solution when mmap hasn't been implemented) to mmap.
    3. Modified `lind_syscall.c` to special handle mmap result. Since mmap is returning a 32-bit address, which might overflow `int` type and become negative, we need to distinguish the valid memory address from errno as errno are also returned as negative number from rawposix. This is achieved by checking the range of the negative value since mmap result is always multiple of pages (4096), while errno has a maximum value that is far less than 4096.

### DIFF
--- a/src/RawPOSIX/src/safeposix/syscalls/sys_calls.rs
+++ b/src/RawPOSIX/src/safeposix/syscalls/sys_calls.rs
@@ -179,7 +179,7 @@ impl Cage {
             sigset: newsigset,
             main_threadid: interface::RustAtomicU64::new(0),
             interval_timer: interface::IntervalTimer::new(child_cageid),
-            vmmap: interface::RustLock::new(Vmmap::new()), // Initialize empty virtual memory map for new process
+            vmmap: interface::RustLock::new(new_vmmap), // Initialize empty virtual memory map for new process
             zombies: interface::RustLock::new(vec![]),
             child_num: interface::RustAtomicU64::new(0),
         };
--- a/src/glibc/lind_syscall/lind_syscall.c
+++ b/src/glibc/lind_syscall/lind_syscall.c
@@ -32,15 +32,23 @@ int __imported_wasi_snapshot_preview1_lind_syscall(unsigned int callnumber, unsi
 //            handled here instead
 int lind_syscall (unsigned int callnumber, unsigned long long callname, unsigned long long arg1, unsigned long long arg2, unsigned long long arg3, unsigned long long arg4, unsigned long long arg5, unsigned long long arg6)
 {
-  int ret = __imported_wasi_snapshot_preview1_lind_syscall(callnumber, callname, arg1, arg2, arg3, arg4, arg5, arg6);
-  // handle the errno
-  if(ret < 0)
-  {
-    errno = -ret;
-  }
-  else
-  {
-    errno = 0;
-  }
-  return ret;
+    int ret = __imported_wasi_snapshot_preview1_lind_syscall(callnumber, callname, arg1, arg2, arg3, arg4, arg5, arg6);
+    // handle the errno
+    // in rawposix, we use -errno as the return value to indicate the error
+    // but this may cause some issues for mmap syscall, because mmap syscall
+    // is returning an 32-bit address, which may overflow the int type (i32)
+    // luckily we can handle this easily because the return value of mmap is always
+    // multiple of pages (typically 4096) even when overflow, therefore we can distinguish
+    // the errno and mmap result by simply checking if the return value is
+    // within the valid errno range
+    if(ret < 0 && ret > -256)
+    {
+        errno = -ret;
+        return -1;
+    }
+    else
+    {
+        errno = 0;
+    }
+    return ret;
 }
--- a/src/RawPOSIX/src/interface/mem.rs
+++ b/src/RawPOSIX/src/interface/mem.rs
@@ -12,6 +12,13 @@ use std::result::Result;
 // heap is placed at the very top of the memory
 pub const HEAP_ENTRY_INDEX: u32 = 0;
 
+/// Round up the address length to be multiple of pages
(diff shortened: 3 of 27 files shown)


qianxichen233:
There is a minor issue that I just noticed. Since mmap is now find memory space from highest address, so the first mmap will usually return an address that is close to 4294967296. In wasmtime, previously I was using `i32` to hold some of the address when doing the fork or pthread creation, now the mmap address will overflow i32 and cause some issues here. The fix should be pretty simple - just replace i32 with u32. I'll fix this later.

Besides, the test added by https://github.com/Lind-Project/lind-wasm/pull/67 (chain_thread.c), which was not able to run previously because it uses mmap somewhere, is verified to be able to run successfully after mmap has been integrated. Though I did some temporary tweak to make mmap is returning address from 2^31 so that it will not overflow i32. But this should at least tell us that things are pretty much there and we only have a minor issue (replace i32 to u32) to fix

src/RawPOSIX/src/safeposix/syscalls/sys_calls.rs
```diff
@@ -179,7 +179,7 @@ impl Cage {
             sigset: newsigset,
             main_threadid: interface::RustAtomicU64::new(0),
             interval_timer: interface::IntervalTimer::new(child_cageid),
-            vmmap: interface::RustLock::new(Vmmap::new()), // Initialize empty virtual memory map for new process
```
rennergade, line 182: are we sure that this deep copies on clone?

rennergade (commented):
Looks very good. I don't have any specific requests but would like to see a few more tests.

1. more complicated situations with mmap/munmap and fork
2. Tests that use malloc instead of sbrk

src/glibc/lind_syscall/lind_syscall.c
```diff
@@ -32,15 +32,23 @@ int __imported_wasi_snapshot_preview1_lind_syscall(unsigned int callnumber, unsi
...
-    errno = 0;
-  }
-  return ret;
+    int ret = __imported_wasi_snapshot_preview1_lind_syscall(callnumber, callname, arg1, arg2, arg3, arg4, arg5, arg6);
+    // handle the errno
+    // in rawposix, we use -errno as the return value to indicate the error
+    // but this may cause some issues for mmap syscall, because mmap syscall
+    // is returning an 32-bit address, which may overflow the int type (i32)
+    // luckily we can handle this easily because the return value of mmap is always
+    // multiple of pages (typically 4096) even when overflow, therefore we can distinguish
```
Yaxuan-w, line 41: I wonder if we should consider scenarios where the page size might change in the future (ie: performance optimizations / in specific environments that require a different page size / etc.). If that’s the case, would it make sense to dynamically retrieve the page size (ie: `sysconf(_SC_PAGESIZE)`) / just leave a TODO?

Yaxuan-w (commented):
Great work! This is a solid implementation! Overall logic makes sense to me.

src/glibc/lind_syscall/lind_syscall.c
JustinCappos: I would suggest we avoid doing this dynamically because 1) it'll be hard to test and 2) it'll slow things down a little.  I'd favor a TODO and assert approach here.

qianxichen233: I think page size does not really matter here unless it becomes smaller than 256, which will have conflict with errno and I didn't see any easy way to resolve it yet. If page size changed to other value but is still larger than 256, there isn't any changes needed to be done here since PAGE_SIZE is not directly used here

src/RawPOSIX/src/safeposix/syscalls/sys_calls.rs
qianxichen233: I actually cannot be 100% sure about this. `clone` method should be designed to deep-copying itself by convention from what I've read, but if the developer (in our case, developer of nodit library or its sub-dependent) is breaking the convention and doing some hacky things here, it would probably be hard for us to know. The documentation for the [NoditMap](https://docs.rs/nodit/latest/nodit/nodit/map/struct.NoditMap.html#method.clone) only says it returns a **copy** of the value without mentioning shallow or deep.

rennergade: This is something worth just adding a test for @ChinmayShringi

qianxichen233:
update:
1. fixed the u32 issue
2. added two tests:
    - mmap_complicated.c: more complicated use case of mmap that uses shared memory for ipc
    - malloc_large.c: try to malloc an extremely large memory region. This will fail previously but it is working now

There is already a simple malloc test under `memory_tests` so I am only adding one `malloc_large` test

yzhang71 (approved):
I have reviewed the implementations, and everything now looks good to me.

rennergade (approved):

Issue #58, opened 2024-11-26
Compile Python 2.7.2 with WASM


yzhang71:
We are currently attempting to compile Python version 2.7.2 with our Lind-WASM. We selected version 2.7.2 because it is the version used by Lind-NaCl, allowing for a fair comparison. I will also use this issue to track the compilation problems and document their solutions.

yzhang71:
In file included from ../Modules/python.c:3:
In file included from ../Include/Python.h:58:
../Include/pyport.h:243:13: error: "This platform's pyconfig.h needs to define PY_FORMAT_LONG_LONG"
           error "This platform's pyconfig.h needs to define PY_FORMAT_LONG_LONG"
            ^
1 error generated.
make: *** [Makefile:555: Modules/python.o] Error 1

rennergade:
Where did you pull the source code from?

yzhang71:
The error persists because we are performing a cross-compilation, and `PY_FORMAT_LONG_LONG` is not correctly defined for the platform. To resolve this issue in a robust way, we need to define `PY_FORMAT_LONG_LONG` directly in `pyconfig.h`.

#### Solution:
Add the following line to `/Python-2.7.2/build/pyconfig.h`:

```c
#define PY_FORMAT_LONG_LONG "ll"
```

yzhang71:
> Where did you pull the source code from?

I got the source code from here: https://www.python.org/downloads/release/python-272/

rennergade:
A lot of clues for things such as the one above should be in here: https://github.com/Lind-Project/lind_project/tree/main/tests/applications/python

Python was the hardest app to build because its build process uses itself to package everything, but because we couldn't switch between native and Lind in the build process we had to compile a bunch of things by scratch. It would be cool to figure it out seemlessly this time but that may not be possible.

Yaxuan-w:
This [issue](https://github.com/Lind-Project/lind_project/issues/65) would be a good reference from my view

yzhang71:
Parser/pgen ../Grammar/Grammar ../Include/graminit.h ../Python/graminit.c
Parser/pgen: 12: Syntax error: "(" unexpected
make: *** [Makefile:562: Parser/pgen.stamp] Error 2

yzhang71:
The error occurred because the `pgen` we generated was a WASM version. However, the default script attempts to run `pgen` directly, without using a WASM runtime. 

#### Solution
[...]

Issue #29, opened 2024-11-08
Integrate lind_syscall.c and crt1.c build into makefile


qianxichen233:
Currently in glibc, lind_syscall.c and crt1.c havn't been integrated into glibc's build system and have to be manually built. This should be fixed and these two files need to be within a makefile somewhere to be automatically built.

rennergade:
I'm not sure the source for this is in the monorepo. @yzhang71 can you check this and make sure its updated once you fix the crt? @yashaswi2000 can you investigate adding this to our build so we dont have to store object files.

rennergade:
@yzhang71 fixed the issue with main() signatures, but we didn't updated the object file. We either need to fix this for builds or update the cached object file for future builds.

Yaxuan-w:
That would be super helper if we can have a temporary instructions posted here, so that contributors won't be blocked before we finish building build sys 😄

rennergade:
Robin is putting up a PR with the makefiles to build the crt. @yashaswi2000 once those are merged can you take this from here and integrate it into the build process?

PR #76, opened 2025-01-06
readlink and readlinkat implementation on RawPOSIX / Wasmtime / glibc

Yaxuan-w:
This PR adds support for readlink and readlinkat in lind-wasm to address issues encountered during [Python compilation](https://github.com/Lind-Project/lind-wasm/issues/58).

### Changes:

**RawPOSIX:** 
- Added support for `readlink` and `readlinkat` along with necessary path resolution logic.

**glibc:** 
- Previously, glibc would call `readlinkat` from `readlink` if the latter was not fully implemented. Since both syscalls are now implemented separately, the conditional check has been removed.

**Tests:** 
- Added test cases for both `readlink` and `readlinka`t to ensure proper functionality.
- Added a script to set up the testing environment, making it easier to run and validate the changes.


src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
```diff
@@ -1,10 +1,24 @@
...
+/*
+* Edit Note:
+* In lind-wasm, we have separately implemented readlink and readlinkat, so we modified this part of the code to handle them individually.
+*/
+ssize_t
+__libc_readlinkat (int fd, const char *path, char *buf, size_t len)
 {
-  return 0;
+  int ret =  MAKE_SYSCALL(166, "syscall|readlinkat",(uint64_t) fd, (uint64_t) path, (uint64_t)(uintptr_t) buf, (uint64_t) len, NOTUSED, NOTUSED);
+  if (ret < 0) {
```
qianxichen233, line 18: this should be handled within MAKE_SYSCALL. I saw you modified lind_syscall.c, but if you are facing the issue that after modififation of lind_syscall.c, it still behaves unexpected, it is very likely because lind_syscall.c has not been compiled as lind_syscall.c hasn't been integrated into build system yet, and you have to manually run the full command to compile it.

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
```diff
@@ -1,10 +1,24 @@
 #include <unistd.h>
 #include <fcntl.h>
-#include <string.h>
+#include <errno.h>
+#include <sysdep-cancel.h>
```
qianxichen233, line 4: what is the purpose of this header file?

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
```diff
@@ -1,10 +1,24 @@
...
+#include <syscall-template.h>
 
-ssize_t __GI_readlinkat (int __fd, const char *__file_name, char *__buf, size_t __len)
```
qianxichen233, line 5: just wondering why you want to change the name of the function

src/glibc/sysdeps/unix/sysv/linux/readlink.c
```diff
@@ -18,18 +18,25 @@
...
 ssize_t
 __readlink (const char *path, char *buf, size_t len)
 {
-#ifdef __NR_readlink
-  return INLINE_SYSCALL_CALL (readlink, path, buf, len);
-#else
-  return INLINE_SYSCALL_CALL (readlinkat, AT_FDCWD, path, buf, len);
-#endif
+  int ret = MAKE_SYSCALL(165, "syscall|readlink", (uint64_t) path, (uint64_t)(uintptr_t) buf, (uint64_t) len, NOTUSED, NOTUSED, NOTUSED);
+  if (ret < 0) {
```
qianxichen233, line 36: same question

qianxichen233 (commented):
Have a few doubts but overall looks good to me. Also a reminder to @ChinmayShringi , once this PR is merged, we need to also add address checking for the two new syscalls we have here in current memory PR or a seperate PR, depends on whether this PR or the memory PR merges first

src/glibc/sysdeps/unix/syscalls.list
```diff
@@ -47,6 +47,7 @@ profil		-	profil		i:piii	__profil	profil
 ptrace		-	ptrace		i:iiii	ptrace
 read		-	read		Ci:ibU	__libc_read	__read read
 readlink	-	readlink	i:spU	__readlink	readlink
+readlinkat	-	readlinkat	i:spU	__libc_readlinkat	readlinkat
```
qianxichen233, line 50: This may does not matter but just wondering why you are adding this here. I think readlinkat is defined in another syscalls.list (`sysdeps/unix/sysv/linux/syscalls.list`) before and I am not very sure if this syscalls.list is really responsbie for readlinkat

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -370,6 +370,187 @@ impl Cage {
...
+    */
+    pub fn readlink_syscall(
+        &self, 
+        path: &str, 
+        buf: *mut u8, 
+        buflen: usize,
+    ) -> i32 {
+        // Convert the path from relative path (lind-wasm perspective) to real kernel path (host kernel
+        // perspective)
+        let relpath = normpath(convpath(path), self);
```
rennergade, line 397: this is sort of an aside, but I feel like we do this type of conversion a lot in fs_calls. Perhaps we can shorten this to some helper functions? Would be a good starter PR for someone new.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -370,6 +370,187 @@ impl Cage {
...
+        virtual_fd: i32, 
+        path: &str, 
+        buf: *mut u8, 
+        buflen: usize,
+    ) -> i32 {
+        let mut libcret;
+        let mut path = path.to_string();
+        let libc_buflen = buflen + LIND_ROOT.len();
+        let mut libc_buf = vec![0u8; libc_buflen];
+        if virtual_fd == libc::AT_FDCWD {
```
rennergade, line 484: we moved to using our own defined constants instead of libc, but similar to qianxi's comment were waiting on the memory PR to be merged. So this will have to be fixed depending what ordering we merge.

src/glibc/sysdeps/unix/syscalls.list
rennergade: Interesting. Do we have notes on "how to add a syscall"? Having these steps documented in the docs would be great.

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
Yaxuan-w: I wasn’t able to find any documentation specifically mentioning that lind_syscall.c requires separate compilation (including on the website and in the scripts). I did notice that lind_syscall.o is being generated, but I’m not entirely sure if this means lind_syscall.c is fully compiled or if there’s a special command needed to include this .o file in the final WASM binary. Could you provide more details about what you mean by the "full command" for compiling it?

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
Yaxuan-w: During a quick search, I noticed that many syscalls (ie: open), even those that are not directly blocking, include this header file in their implementations under the sysdeps directory. Based on my understanding, this header file is primarily used to handle thread cancellation cases. Considering potential edge cases and to maintain code consistency, I chose to include this header file in my implementation as well.

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
Yaxuan-w: I think glibc uses the `__GI_` prefix to distinguish whether a syscall is part of its public API. I noticed that in our previous reimplementations of syscalls, we used the `__libc_` prefix for this purpose. To maintain consistency in the lind-wasm codebase, I chose to adjust the naming in my modification accordingly.

src/glibc/sysdeps/unix/syscalls.list
Yaxuan-w: From my understanding, `sysdeps/unix/syscalls.list` defines general unix syscall support, while `sysdeps/unix/sysv/linux/syscalls.list` adds linux-specific extensions or overrides based on the former. In the previous glibc implementation, `readlinkat` wasn’t directly exposed as a public API -- it was invoked indirectly via `readlink` (as seen in the` __GI_ `naming beforehand).

For the lind-wasm implementation of `readlink` and `readlinkat`, I decided to expose both as public APIs. To support this, I modified `sysdeps/unix/syscalls.list`. Considering whether `sysdeps/unix/sysv/linux/syscalls.list` might overwrite these changes, my understanding is that the `weak_alias(__libc_readlinkat, readlinkat)` binds `readlinkat` to the `__libc_readlinkat` implementation, which should prevent it from being overwritten. As such, I think modifications to `sysdeps/unix/sysv/linux/syscalls.list` might not be strictly necessary. That said, I’m not entirely certain if modifying only `sysdeps/unix/sysv/linux/syscalls.list` could achieve the same effect. Any clarification on this would be greatly appreciated.

Additionally, I noticed we don’t currently have documentation outlining the code standards for lind-wasm. It might be helpful to discuss this in today’s meeting to ensure we’re aligned moving forward.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
Yaxuan-w: Agreed!! I'll open an issue to document this

src/glibc/sysdeps/unix/syscalls.list
Yaxuan-w: Also a quick note, I'm a bit confused by differences between `wasmtest.sh` and `lindtool.sh`... It seems some redundant functionality and there're some minor issue in `wasmtest.sh` (missing required abs path / incorrect command operation / etc.)..

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
qianxichen233: this is something runbin is working on. He will make all the files that requires manual builds automatically later

qianxichen233: So hopefully we will not need to care about this anymore. But for now you can use compile lind_syscall.c using `/home/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04/bin/clang --target=wasm32-unkown-wasi -v -Wno-int-conversion lind_syscall.c -c -std=gnu11 -fgnu89-inline -matomics -mbulk-memory -O2 -g -Wall -Wwrite-strings -Wundef -fmerge-all-constants -ftrapping-math -fno-stack-protector -fno-common -Wp,-U_FORTIFY_SOURCE -Wstrict-prototypes -Wold-style-definition -fmath-errno -fPIE -ftls-model=local-exec -I../include -I/home/lind-wasm/src/glibc/build/nptl -I/home/lind-wasm/src/glibc/build -I../sysdeps/lind -I../lind_syscall -I../sysdeps/unix/sysv/linux/i386/i686 -I../sysdeps/unix/sysv/linux/i386 -I../sysdeps/unix/sysv/linux/x86/include -I../sysdeps/unix/sysv/linux/x86 -I../sysdeps/x86/nptl -I../sysdeps/i386/nptl -I../sysdeps/unix/sysv/linux/include -I../sysdeps/unix/sysv/linux -I../sysdeps/nptl -I../sysdeps/pthread -I../sysdeps/gnu -I../sysdeps/unix/inet -I../sysdeps/unix/sysv -I../sysdeps/unix/i386 -I../sysdeps/unix -I../sysdeps/posix -I../sysdeps/i386/fpu -I../sysdeps/x86/fpu -I../sysdeps/i386 -I../sysdeps/x86/include -I../sysdeps/x86 -I../sysdeps/wordsize-32 -I../sysdeps/ieee754/float128 -I../sysdeps/ieee754/ldbl-96/include -I../sysdeps/ieee754/ldbl-96 -I../sysdeps/ieee754/dbl-64 -I../sysdeps/ieee754/flt-32 -I../sysdeps/ieee754 -I../sysdeps/generic -I.. -I../libio -I. -nostdinc -isystem /home/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04/lib/clang/16/include -isystem /usr/i686-linux-gnu/include -D_LIBC_REENTRANT -include /home/lind-wasm/src/glibc/build/libc-modules.h -DMODULE_NAME=libc -include ../include/libc-symbols.h -DPIC -DTOP_NAMESPACE=glibc -o ./lind_syscall.o`

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -370,6 +370,187 @@ impl Cage {
...
+            let relative_path = relpath.to_str().unwrap();
+            let full_path = format!("{}{}", LIND_ROOT, relative_path);
+            let c_path = CString::new(full_path).unwrap();
+            
+            libcret = unsafe {
+                libc::readlinkat(vfd.underfd as i32, c_path.as_ptr(), libc_buf.as_mut_ptr() as *mut c_char, libc_buflen)
+            };
+        }
+
+        if libcret < 0 {
```
m-hemmings, line 518: While this checks for a negative value, what would happen if libcret was the result of an empty link target? Would that return a 0 value?

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
Yaxuan-w: After doing this step, the return value still behaves unexpected.

1. If I understand everything correctly, this command misses the `lind-wasm` for clang dir, and needs to be run under `lind_syscall` dir. So the command line I used is:

`/home/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04/bin/clang --target=wasm32-unkown-wasi -v -Wno-int-conversion lind_syscall.c -c -std=gnu11 -fgnu89-inline -matomics -mbulk-memory -O2 -g -Wall -Wwrite-strings -Wundef -fmerge-all-constants -ftrapping-math -fno-stack-protector -fno-common -Wp,-U_FORTIFY_SOURCE -Wstrict-prototypes -Wold-style-definition -fmath-errno -fPIE -ftls-model=local-exec -I../include -I/home/lind-wasm/src/glibc/build/nptl -I/home/lind-wasm/src/glibc/build -I../sysdeps/lind -I../lind_syscall -I../sysdeps/unix/sysv/linux/i386/i686 -I../sysdeps/unix/sysv/linux/i386 -I../sysdeps/unix/sysv/linux/x86/include -I../sysdeps/unix/sysv/linux/x86 -I../sysdeps/x86/nptl -I../sysdeps/i386/nptl -I../sysdeps/unix/sysv/linux/include -I../sysdeps/unix/sysv/linux -I../sysdeps/nptl -I../sysdeps/pthread -I../sysdeps/gnu -I../sysdeps/unix/inet -I../sysdeps/unix/sysv -I../sysdeps/unix/i386 -I../sysdeps/unix -I../sysdeps/posix -I../sysdeps/i386/fpu -I../sysdeps/x86/fpu -I../sysdeps/i386 -I../sysdeps/x86/include -I../sysdeps/x86 -I../sysdeps/wordsize-32 -I../sysdeps/ieee754/float128 -I../sysdeps/ieee754/ldbl-96/include -I../sysdeps/ieee754/ldbl-96 -I../sysdeps/ieee754/dbl-64 -I../sysdeps/ieee754/flt-32 -I../sysdeps/ieee754 -I../sysdeps/generic -I.. -I../libio -I. -nostdinc -isystem /home/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04/lib/clang/16/include -isystem /usr/i686-linux-gnu/include -D_LIBC_REENTRANT -include /home/lind-wasm/src/glibc/build/libc-modules.h -DMODULE_NAME=libc -include ../include/libc-symbols.h -DPIC -DTOP_NAMESPACE=glibc -o ./lind_syscall.o`

2. I noticed this command will generate a `lind_syscall.o`, which I also generated by doing `./lind_tool.sh make_all`. I'm unsure what's their differences.

JustinCappos: Let's make sure this fails in some obvious way with a clear error message, if this step is skipped.

Let's also make running this command part of the build process which is automatically done.

qianxichen233: Yes, we have an issue for [this](https://github.com/Lind-Project/lind-wasm/issues/29) and Robin is working on this right now

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
Yaxuan-w: readlinkat returns the length of the symbolic link's target path. If the symbolic link's target path is empty (i.e. it points to a non-existent file or an empty path), readlinkat returns 0 without filling buf with any data.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -370,6 +370,187 @@ impl Cage {
...
+        path: &str, 
+        buf: *mut u8, 
+        buflen: usize,
+    ) -> i32 {
+        // Convert the path from relative path (lind-wasm perspective) to real kernel path (host kernel
+        // perspective)
+        let relpath = normpath(convpath(path), self);
+        let relative_path = relpath.to_str().unwrap();
+        let full_path = format!("{}{}", LIND_ROOT, relative_path);
+        let c_path = CString::new(full_path).unwrap();
```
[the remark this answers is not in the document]
Yaxuan-w: Thanks for pointing this out! I hadn't considered the possibility of the path containing non utf-8 characters during the design. You're right that if `full_path` contains non utf-8 characters, the line `let relative_path = relpath.to_str().unwrap();` would panic. To avoid this, we should implement better error handling (ie: using `return EINVAL` instead of `.unwrap()`). I will open an issue to document this and address it in the following refactor.

src/glibc/sysdeps/unix/syscalls.list
rennergade: Yes, definitely agree. This is one of the things I'm hoping @m-hemmings will guide us on in the short term while working with the devops team.

src/glibc/sysdeps/unix/sysv/linux/i386/i686/readlinkat.c
Yaxuan-w: thanks it works!

src/glibc/sysdeps/unix/syscalls.list
qianxichen233: So `lindtool.sh` is actually my personal script that's basically helping me to do things more efficient. So it basically contains all the commands that I might need to use during development. As more people get involved into lind-wasm, I shared this script to them and then this script somehow get uploaded to github.
`wasmtest.sh` is a new script wrote by Robin for testing purpose only (I think?). I think `wasmtest.sh` is directly modified upon `lindtool.sh` so there are a lot of similarity here.

src/glibc/sysdeps/unix/sysv/linux/readlink.c
```diff
@@ -18,18 +18,20 @@
...
 #include <fcntl.h>
-#include <sysdep.h>
+#include <sysdep-cancel.h>
+#include <syscall-template.h>
+#include <errno.h>
 
 /* Read the contents of the symbolic link PATH into no more than
    LEN bytes of BUF.  The contents are not null-terminated.
    Returns the number of characters read, or -1 for errors.  */
+/*
```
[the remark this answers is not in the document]
JustinCappos: If the system calls are separate, then these calls should be separate.   We're standardizing on the 64 bit Linux system call interface.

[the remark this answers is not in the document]
rennergade: That makes sense, looking at the system call table they have separate entries. But then I'm not exactly sure what this comment is explaining.

@Yaxuan-w were both readlink and readlinkat calling to readlink before this was modified for Lind?

rennergade (approved):

qianxichen233 (approved):

PR #83, opened 2025-01-12
Add conditional compilation to support multi-threaded cargo test

Yaxuan-w:
## Description

This PR addresses an issue encountered during the execution of the vmmap inner tests (reported by @ChinmayShringi). The test failures were caused by incorrect re-initialization of standard streams (stdin, stdout, stderr) in the cage setup during concurrent test execution in a multi-threaded cargo test environment, where the handling of standard file descriptors affected the test stability.

 

In fact, the test cases have all successfully executed (as evidenced by checking /tmp, where the corresponding test files are successfully created), so the issue is related to the inability to print out information. Since cargo test runs in a single process with multi-threading, theoretically, even after `kernel_close` closes stdin, stdout, and stderr, they should be re-initialized in the next cage setup (inside [`lindrustinit`](https://github.com/Lind-Project/lind-wasm/blob/main/src/RawPOSIX/src/safeposix/dispatcher.rs#L1119)). However, during `cargo test`, the failure might be caused by fd positions being occupied by other test threads (e.g., when a new test thread opens a test file), resulting in the cage initialization failing to reopen the host kernel's stdin, stdout, and stderr.

Considering scenarios such as running large applications like the LAMP stack, which may require handling background services or daemon processes, it could necessitate closing or redirecting standard streams to /dev/null. Therefore, from my perspective, the ability to close stdin, stdout, and stderr should not be removed. To address this issue, I have modified `kernel_close` to expand its logic, forcing it to disallow the closure of stdin, stdout, and stderr during testing by adding conditional compilation. If there is a more suitable solution for this case, I am open to discussing it.

Additionally, I reviewed several error messages from the inner test suite of file system-related tests. Some of these errors were caused by running tests under `sudo` mode, which bypasses normal permission limitations, or by incorrect test case settings.

### Type of change

## Checklist:

- [x] Any dependent changes have been added to a pull request and/or merged in other modules (native-client, lind-glibc, lind-project)


src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1746,19 +1746,26 @@ pub fn lindrustinit(verbosity: isize) {
     interface::cagetable_insert(0, utilcage);
     fdtables::init_empty_cage(0);
     // Set the first 3 fd to STDIN / STDOUT / STDERR
-    // STDIN
-    let dev_null = CString::new("/home/lind/lind_project/src/safeposix-rust/tmp/dev/null").unwrap();
+    // TODO:
+    // Replace the hardcoded values with variables (possibly by adding a LIND-specific constants file)
```
ChinmayShringi, line 1750: Is it okay if I add the constant in the constants folder?
`src/RawPOSIX/src/constants`

Yaxuan-w: I'm okay with this

rennergade:
Cool! Great that you figured this out.

Thinking this over my belief is that we should never actually close these streams at all, not just disallowing it for tests, and the closing for these per cage should just be virtual.

Otherwise, for example, if Cage 1 closes STDOUT, then any other cage wouldn't be able to print to the terminal anymore.

rennergade:
I think this is good to go. But would love to change the PR format so it's easier to track later on.

1. Can we rename this so its not test related, something like "Disable closing host streams" 
2. Can we merge this to main instead of the vmmap PR? We can just remerge main into vmmap once this is there.

Thanks!

Yaxuan-w:
This branch is merged from `vmmap-alice`, so I'll close this one and open another pr :D

Issue #11, opened 2024-08-08
Add Lind-WASM Test Cases with build system and test automation


yzhang71:
So for Lind, we have the test cases for each system call available here: https://github.com/Lind-Project/lind_project/tree/main/tests/test_cases. We want to handle the system calls in Lind-WASM as well. The task is to create Lind-WASM test cases for each system call, similar to the approach used in https://github.com/Lind-Project/lind_project/blob/main/tests/test_cases/suite.sh. If a system call is not supported, it needs to be added to the rustposix.

rennergade:
Now that we have the monorepo, @robinyuan1002 can we port these tests in the "tests" folder. Let's also set up a build script to build all of them and a run script to test all of them against native like we had in rustposix.

rennergade:
The next step here is to generate all the expected outputs.

@arnabdas1999 can you compile all of the c file tests for native linux using gcc in this folder: https://github.com/Lind-Project/lind-wasm/tree/main/tests/unit-tests and redirect the output to a *.expected file

ie
```
gcc hello.c -o hello
./hello > hello.expected
```

Once you have those can you PR them by adding them in a /expected subfolder for each category.

rennergade:
@ansalma6 is going to take over the above task. @yashaswi2000 can you help walk him through this?

rennergade:
We're getting closer to getting this up an integrated with CI

@qianxichen233 @Yaxuan-w here's the current report of whats succeeding and failing on main. 

[report.json](https://github.com/user-attachments/files/18758039/report.json)

@ansalma6 @m-hemmings are going to work on getting this integrated into CI and have the output be easier to read.

Yaxuan-w:
> current report of whats succeeding and failing on main.

I'm a bit confused by the report... I could successfully run `rename.c` this morning with main branch. The error msg on report is 

```txt
"/home/lind/lind-wasm/tests/unit-tests/file_tests/deterministic/rename.c": {
    "status": "Failure",
    "error_type": "Failure_GCC_running",
    "output": "ERROR WITH STAT CALL ON OLD NAME: No such file or directory\n"
},
```

which I think is related to test setup? For example: creating test files in `src/RawPOSIX/tmp` before testing

qianxichen233:
since vmmap has been merged, can we run the test again with the latest main branch? Vmmap is pretty big update so probably the test report could be much different

rennergade:
[...]

PR #90, opened 2025-01-23
add function to create expected outputs for deterministic tests

ansalma6:
PR for https://github.com/Lind-Project/lind-wasm/issues/11

Added a function in wasmtest.sh file for creating expected outputs. 
The function can be ran using "wasmtest.sh createexpected" or "wasmtest.sh ce"

All c files under each of the deterministic subfolder are taken and compiled using gcc. 
A new folder expected is created in each deterministic subfolder and the output is saved in a <test_case_filename>.output file
If there are any compile or runtime error, they are displayed in the console and alse added to the .output file.
The compiled files are deleted after running
Finally a count of total/success/failure is displayed in the console

### DIFF
--- a/wasmtest.sh
+++ b/wasmtest.sh
@@ -31,6 +31,9 @@ precompile_wasm="$wasmtime_base/target/debug/wasmtime compile [input] -o [output
 run_cmd_precompile="$wasmtime_base/target/debug/wasmtime run --allow-precompiled --wasi threads=y --wasi preview2=n [target]"
 
 test_file_base="$LIND_WASM_BASE/tests/unit-tests"
+deterministic_dir="deterministic"
+safe_test_list="testlist_safe.txt"
+unsafe_test_list="testlist_unsafe.txt"
 
 #color codes for terminal output
 RED='\033[31m'
@@ -202,12 +205,19 @@ run_from_files() {
 create_expected_outputs() {
     set +e +o pipefail
 
+    [ ! -f "$safe_test_list" ] && touch "$safe_test_list"
+    echo -n "" > "$safe_test_list"
+
+    [ ! -f "$unsafe_test_list" ] && touch "$unsafe_test_list"
+    echo -n "" > "$unsafe_test_list"
+
     total_files=0
     total_success=0
     total_failure=0
 
-    for deterministic_dir in $(find tests/unit-tests -type d -name deterministic); do
+    for deterministic_dir in $(find "$test_file_base" -type d -name "$deterministic_dir"); do
         mkdir -p "$deterministic_dir/expected"
+
         for c_file in "$deterministic_dir"/*.c; do
             total_files=$((total_files + 1))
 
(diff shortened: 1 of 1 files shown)


wasmtest.sh
```diff
@@ -198,6 +198,54 @@ run_from_files() {
...
 
+#function to create the expected outputs 
+create_expected_outputs() {
+    set +e +o pipefail
+
+    total_files=0
+    total_success=0
+    total_failure=0
+
+    for deterministic_dir in $(find tests/unit-tests -type d -name deterministic); do
```
m-hemmings, line 209: Instead of hard coding the name for the file to be found, could it be added as a parameter or configurable variable?
Something like:
`TARGET_DIR_NAME="deterministic"
for deterministic_dir in $(find tests/unit-tests -type d -name "$TARGET_DIR_NAME"); do`

wasmtest.sh
```diff
@@ -198,6 +198,54 @@ run_from_files() {
     echo -e "${GREEN}All tests from file list completed.${RESET}"
 }
 
+#function to create the expected outputs
```
m-hemmings, line 201: Could the comments at the head of the function be clearer as to the purpose and flow of the function? Perhaps something like:
```
# ----------------------------------------------------------------------
# Function: create_expected_outputs
#
# Purpose:
# Compiles `.c` files in "tests/unit-tests/deterministic" directories, 
# runs the binaries, and saves their output in "expected" subdirectories.
#
# Variables:
# - Input: None required; works on `.c` files in specific directories.
# - Output: Logs total processed, successes, and failures to stdout.
#
# Exceptions:
# - Compilation errors: Logged to stdout and corresponding `.output` files.
# - Runtime errors: Logged to stdout and corresponding `.output` files.
# - Skips directories or files if not found; no explicit errors thrown.
#
# Note:
# Temporarily disables `set -e` and `set -o pipefail` for error handling.
# ----------------------------------------------------------------------

```

Yaxuan-w (commented):
I was wondering if we might consider adding a cleanup as an optional feature as well. Since we will create a subfolder for each deterministic category, manually cleaning them up could become messy. 

Would it be possible to include something like `wasmtest.sh cleanup` or `wasmtest.sh cl`, to remove all the `/expected` folders and their contents?

ansalma6:
Changes from Review comments:
Added a new function to cleanup the generated expected outputs "wasmtest.sh cleanexpected" or "wasmtest.sh cle"
Added comment headers for functions
Outputs the paths of successful tests to testlist_safe.txt and failed tests to testlist_unsafe (one file per line)

Note:
Renamed the function to generate expected outputs to avoid potential confusion with cleanup arguments.
"wasmtest.sh generateexpected" or "wasmtest.sh ge"

wasmtest.sh
ansalma6: Added the header

wasmtest.sh
ansalma6: Changed to use variable

rennergade (approved):
This looks good, agree with the previous comments. @yashaswi2000 We should probably try to cache these expected outputs though since they shouldnt change.

Yaxuan-w (approved):
LGTM

yashaswi2000 (commented):
LGTM!

PR #92, opened 2025-01-25
implement clock_gettime

robinyuan1002:
Now Lind-wasm can supprot clock_gettime syscall.

### DIFF
--- a/src/RawPOSIX/src/safeposix/dispatcher.rs
+++ b/src/RawPOSIX/src/safeposix/dispatcher.rs
@@ -118,6 +118,7 @@ const WAIT_SYSCALL: i32 = 172;
 const WAITPID_SYSCALL: i32 = 173;
 
 const NANOSLEEP_TIME64_SYSCALL : i32 = 181;
+const CLOCK_GETTIME : i32 = 191;
 
 use std::ffi::CString;
 use std::ffi::CStr;
@@ -1032,6 +1033,14 @@ pub fn lind_syscall_api(
             
             interface::cagetable_getref(cageid)
                 .nanosleep_time64_syscall(clockid, flags, req, rem)
+        }
+
+        CLOCK_GETTIME => {
+            let clockid = arg1 as u32;
+            let tp = (start_address + arg2) as usize;
+            
+            interface::cagetable_getref(cageid)
+                .clock_gettime(clockid, tp)
         }
 
         WAIT_SYSCALL => {
--- a/src/RawPOSIX/src/safeposix/dispatcher.rs
+++ b/src/RawPOSIX/src/safeposix/dispatcher.rs
@@ -118,7 +118,7 @@ const WAIT_SYSCALL: i32 = 172;
 const WAITPID_SYSCALL: i32 = 173;
 
 const NANOSLEEP_TIME64_SYSCALL : i32 = 181;
-const CLOCK_GETTIME : i32 = 191;
+const CLOCK_GETTIME_SYSCALL : i32 = 191;
 
 use std::ffi::CString;
 use std::ffi::CStr;
@@ -1035,7 +1035,7 @@ pub fn lind_syscall_api(
                 .nanosleep_time64_syscall(clockid, flags, req, rem)
         }
 
-        CLOCK_GETTIME => {
+        CLOCK_GETTIME_SYSCALL => {
             let clockid = arg1 as u32;
             let tp = (start_address + arg2) as usize;
             
--- a/src/RawPOSIX/src/safeposix/dispatcher.rs
+++ b/src/RawPOSIX/src/safeposix/dispatcher.rs
@@ -1040,7 +1040,7 @@ pub fn lind_syscall_api(
             let tp = (start_address + arg2) as usize;
             
             interface::cagetable_getref(cageid)
-                .clock_gettime(clockid, tp)
+                .clock_gettime_syscall(clockid, tp)
         }
 
         WAIT_SYSCALL => {
--- a/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
+++ b/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
@@ -1901,6 +1901,15 @@ impl Cage {
         }
         ret
     }
+
+    pub fn clock_gettime(&self, clockid: u32, tp: usize) -> i32 {
+        let ret = unsafe { syscall(SYS_clock_gettime, clockid, tp) as i32 };
+        if ret < 0 {
+            let errno = get_errno();
+            return handle_errno(errno, "clock_gettime");
+        }
+        ret
+    }
 }
 
 /// Lind-WASM is running as same Linux-Process from host kernel perspective, so standard fds shouldn't 
--- a/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
+++ b/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
@@ -1902,7 +1902,7 @@ impl Cage {
         ret
     }
 
-    pub fn clock_gettime(&self, clockid: u32, tp: usize) -> i32 {
+    pub fn clock_gettime_syscall(&self, clockid: u32, tp: usize) -> i32 {
         let ret = unsafe { syscall(SYS_clock_gettime, clockid, tp) as i32 };
         if ret < 0 {
             let errno = get_errno();
--- a/src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
+++ b/src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
@@ -26,6 +26,10 @@
 #include <syscall-template.h>
 
 /* Get current value of CLOCK and store it in TP.  */
+
+// Now after clang compile it will call gettime64 first(but in real case we have only 32 bit)
+// In this case, we need “#if __TIMESIZE != 64” and this is always true. And always call __clock_gettime at last.
+
 int
 __clock_gettime64 (clockid_t clock_id, struct __timespec64 *tp)
 {
--- a/src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
+++ b/src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
@@ -23,66 +23,13 @@
 #include "kernel-posix-cpu-timers.h"
 #include <sysdep-vdso.h>
 #include <shlib-compat.h>
+#include <syscall-template.h>
 
 /* Get current value of CLOCK and store it in TP.  */
 int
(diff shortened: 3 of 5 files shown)


src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
```diff
@@ -23,66 +23,13 @@
...
-  r = INTERNAL_SYSCALL_CALL (clock_gettime, clock_id, &tp32);
-  if (r == 0)
-    {
-      *tp = valid_timespec_to_timespec64 (tp32);
-      return 0;
-    }
-#endif
-
-  return INLINE_SYSCALL_ERROR_RETURN_VALUE (-r);
+  return MAKE_SYSCALL(191, "syscall|clock_gettime", (uint64_t) clock_id, (uint64_t) tp, NOTUSED, NOTUSED, NOTUSED, NOTUSED);
```
Yaxuan-w, line 32: More comments are welcomed

Yaxuan-w (commented):
I noticed that more than one PR is showing 'merge blocked' because it requires every commit to be signed. Should we talk with everyone about [setting this up](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and resubmitting new PRs to ensure compliance with this requirement?

JustinCappos:
> I noticed that more than one PR is showing 'merge blocked' because it requires every commit to be signed. Should we talk with everyone about [setting this up](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and resubmitting new PRs to ensure compliance with this requirement?

Yes, please reach out and get this set up.   

For right now, people can ask me and I'll manually bypass the check for PRs so that we don't need to amend / re-sign commits, etc.

Yaxuan-w (approved):
> @Yaxuan-w is this ready to merge?

@JustinCappos I think the PR is good to merge. 

My understanding is that in NaCl, the logic for `clock_gettime` is handled within the sandbox. In Wasmtime, since rawposix directly invokes the kernel, it fetches the time directly from the kernel. And I have verified the test case logic and found no issues.

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -118,6 +118,7 @@ const WAIT_SYSCALL: i32 = 172;
 const WAITPID_SYSCALL: i32 = 173;
 
 const NANOSLEEP_TIME64_SYSCALL : i32 = 181;
+const CLOCK_GETTIME : i32 = 191;
```
JustinCappos, line 121: Is this a syscall?   Should it have this listed in the name?

src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
```diff
@@ -91,26 +42,11 @@ libc_hidden_def (__clock_gettime64)
...
-        {
-          __set_errno (EOVERFLOW);
-          return -1;
-        }
-
-      *tp = valid_timespec64_to_timespec (tp64);
-    }
-
-  return ret;
+  return MAKE_SYSCALL(191, "syscall|clock_gettime", (uint64_t) clock_id, (uint64_t) tp, NOTUSED, NOTUSED, NOTUSED, NOTUSED);
```
JustinCappos, line 45: Why is this 191 instead of the #define value?

JustinCappos (changes requested):

src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
qianxichen233: yeah, I believe this is a legacy issue when the glibc codebase for lind-wasm was initially set up. It might be worth creating a separate PR to address this, such as migrating `lind_platform.h` from lind-nacl and use its defined values here.
Besides, I remember you also mentioned in the previous meeting that we should align with the exact syscall numbers used by Linux? If that's the case, this change would also be good to be included together.

qianxichen233 (changes requested):
Have you created any testcase for this syscall? Ideally we want all the new syscalls that get added have at least one test case. And for this syscall (clock_gettime), I think it might worth to have one test case testing for `clock_gettime` syscall specifically, and another test case that uses a more high level glibc function that internally used `clock_gettime` such as `time` function.

src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
Yaxuan-w: Agree with this point. During the integration of 3i, I noticed that wasm-glibc codebase contains some unused files, such as lind_platform.h. Additionally, there are several instances of missing unified system call constants, which can be seen in both glibc and wasmtime. I also came across some mistakenly commented-out code that is still useful.

I’m considering conducting a thorough review when the overall codebase is relatively stable (toward the end of the ongoing rawposix refactoring and the implementation of wasmtime thread/signal). This review would focus on:

- Identifying files that were previously developed but are now obsolete compared to standard glibc.
- Unifying syscall numbers and aligning them with the Linux standard.
- Correcting mistakenly commented-out code.
- Addressing any newly discovered issues during integration.

Some issues, such as mistakenly commented-out code, may be easier to identify during the integration process than through manual inspection.

Does this approach seem reasonable?

JustinCappos: Yes.   I would even perhaps starting on it soon instead of waiting.   But do favor parts of the code which should not change.

robinyuan1002:
Yeah, I have create a really simple test case for clock_gettime syscall.

src/RawPOSIX/src/safeposix/dispatcher.rs
robinyuan1002: Yes this is a syscall, I will add _SYSCALL to it.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1901,6 +1901,15 @@ impl Cage {
         }
         ret
     }
+
+    pub fn clock_gettime(&self, clockid: u32, tp: usize) -> i32 {
```
Yaxuan-w, line 1905: add `syscall` at the end of function name

tests/unit-tests/file_tests/non-deterministic/clock_gettime_highlevel.c
qianxichen233, line 1: change this test (or add another test if you prefer this way) to measure the performance of a piece of code using `time` related function. This is the original motivation of why we are adding `clock_gettime` syscall so we need a test for this

qianxichen233 (changes requested):

tests/unit-tests/file_tests/non-deterministic/clock_gettime_highlevel.c
robinyuan1002: @qianxichen233  I think the test case for unit-tests is to test the function for a syscall. To test performance of a piece of code I think we have two ways to deal with it.  First, we can write a script file to test the performance of a c file(but I don't think we want this). Second, we need to add specific codes to the test case(but this depends on benchmark). If we choose the second one, should it be in unit-tests?

qianxichen233: refer to [this](https://stackoverflow.com/questions/5248915/execution-time-of-c-program) for what I am talking about

src/glibc/sysdeps/unix/sysv/linux/clock_gettime.c
```diff
@@ -23,66 +23,17 @@
...
-  r = INTERNAL_SYSCALL_CALL (clock_gettime, clock_id, &tp32);
-  if (r == 0)
-    {
-      *tp = valid_timespec_to_timespec64 (tp32);
-      return 0;
-    }
-#endif
-
-  return INLINE_SYSCALL_ERROR_RETURN_VALUE (-r);
+  return MAKE_SYSCALL(191, "syscall|clock_gettime", (uint64_t) clock_id, (uint64_t) tp, NOTUSED, NOTUSED, NOTUSED, NOTUSED);
```
yzhang71, line 36: Here, we should use 'CLOCK_GETTIME_SYSCALL' as well.

robinyuan1002: I think all the previous format don't have _syscall. Should we add _syscall for this one?

tests/unit-tests/file_tests/non-deterministic/clock_gettime_highlevel.c
```diff
@@ -0,0 +1,32 @@
+#include <stdio.h>
+#include <unistd.h>
+#include <time.h>
+
+int main() {
+    struct timespec begin, end, diff;
+
+    // Get the start time
+    clock_gettime(CLOCK_REALTIME, &begin);
```
qianxichen233, line 9: For high level functions, I mean do not directly use `clock_gettime` syscall, but use more standard c library time functions like `clock`. We need to make sure glibc is compatible with our implementation of `clock_gettime` and common glibc time functions are working correctly with it

qianxichen233 (changes requested):

tests/unit-tests/file_tests/non-deterministic/clock_gettime_highlevel.c
```diff
@@ -1,43 +1,32 @@
 #include <stdio.h>
 #include <unistd.h>
 #include <time.h>
-#include <sys/time.h>
 
 int main() {
```
Yaxuan-w, line 5: Brief explanation about what this test is testing is needed.

tests/unit-tests/file_tests/non-deterministic/clock_gettime_highlevel.c
robinyuan1002: Ok, but I think clock can not count the time for sleep?

tests/unit-tests/file_tests/non-deterministic/clock_gettime_highlevel.c
```diff
@@ -0,0 +1,32 @@
+#include <stdio.h>
+#include <time.h>
+
+/*
+    This is higher level test for clock_gettime, which we don't directly call clock_gettime function but still call this syscall eventually
```
rennergade, line 5: This comment could be better, like explaining that clock() calls clock gettime, and briefly say what the test does.

rennergade (changes requested):
Had a small request about making the test comment better but this should be good to go.

rennergade:
Are we waiting until all the libc syscall numbers are fixed to merge this?

robinyuan1002:
I think it will be easier if we first merge all the syscall PR before we modify. I believe this PR is ready to merge.

rennergade (approved):

PR #95, opened 2025-01-29
Shell script for Lind-Wasm

Utkarsh272:


lind_shell.sh
```diff
@@ -0,0 +1,58 @@
...
+                CURRENT_PATH="$args"
+            else
+                echo "cd: $args: No such file or directory"
+            fi
+        fi
+        continue
+    fi
+
+    # Prepend the default path for file-related commands, if applicable
+    if [[ "$command" =~ ^(ls|cat|touch|rm|mv|cp|mkdir)$ ]]; then
```
JustinCappos, line 50: Wow, this is very command specific.   what if I do rmdir, etc.?   Why exactly these commands?   How would a user know which commands get this path added vs not?

lind_shell.sh
```diff
@@ -0,0 +1,58 @@
+
+# Set the default directory where all file-related operations will occur
```
JustinCappos, line 2: Why?

lind_shell.sh
```diff
@@ -0,0 +1,58 @@
...
+DEFAULT_PATH="/home/lind-wasm/src/RawPOSIX/tmp"
```
qianxichen233, line 3: we should allow the lind fs root to be configurable. One way to do this is by using environment variable. we can try to read from environment for something like `LIND_FS_ROOT`. If it does not exist, then use a fallback path here.

lind_shell.sh
```diff
@@ -0,0 +1,58 @@
...
+            CURRENT_PATH="$DEFAULT_PATH"
+        else
+            # Append to the current path if not an absolute path
+            if [[ ! "$args" =~ ^/ ]]; then
+                args="$CURRENT_PATH/$args"
+            fi
+
+            # Update CURRENT_PATH if the directory exists
+            if [[ -d "$args" ]]; then
+                CURRENT_PATH="$args"
```
qianxichen233, line 41: you are directly appending the args to path. What if user is passing something like `cd ../`? Then the `CURRENT_PATH` would be messed up. It may still point to the correct directory, but the path string would be a mess

qianxichen233, line 41: There is no bound check for directory updating. If I am currently in lind filesystem root, and then I tried `cd ../`, then I am able to escape out of lind filesystem.

lind_shell.sh
```diff
@@ -0,0 +1,58 @@
...
+
+    # Prepend the default path for file-related commands, if applicable
+    if [[ "$command" =~ ^(ls|cat|touch|rm|mv|cp|mkdir)$ ]]; then
+        if [[ ! "$args" =~ ^/ ]]; then
+            args="$CURRENT_PATH/$args"
+        fi
+    fi
+
+    # Execute the command
+    $command $args 2>/dev/null || echo "Error: Command not found or failed"
```
qianxichen233, line 57: there is also no bound check for file operations. If user is doing something like `mkdir ../test`, the folder will be created outside of the lind filesystem

qianxichen233 (changes requested):
Besides these, two other features that we might want:
1. `pwd` is a pretty useful and commonly used command and we might want this command to work **correctly** within lind filesystem
2. we also need to have a cross-filesystem `cp` command that can copy files from host filesystem to lind filesystem. e.g. `lind-cp /home/test.sh /test.sh` can copy the `/home/test.sh` from host to `$LIND_FS_ROOT/test.sh`

lind_shell.sh
```diff
@@ -0,0 +1,58 @@
...
+# Notify the user about the default path
+echo "All file-related changes will occur in: $DEFAULT_PATH"
+
+# Main shell loop
+while true; do
+    # Print the shell prompt with the default path
+    echo -n "lind-wasm > "
+
+    # Read user input
+    read -r input
```
Yaxuan-w, line 16: We should add a `-h/--help` option so users can easily see what commands are supported and how to use them. 

The help message should include:
1. A brief description of what this is at the beginning 
2. A list of supported command options, explaining what each command does, how to use it (e.g., `cp <src> <dest>`), and a short description for clarity.

Yaxuan-w (changes requested):
Great work on this! There are a few details that need further refinement.

In addition to those, I think we need:

1. A high-level PR description explaining what this PR changes and its purpose.
2. A comment section at the beginning of the script that provides a clear explanation of how to use it.

lind_shell.sh
Utkarsh272: I have added shell description making it more readable and easy to understand

lind_shell.sh
Utkarsh272: It resolves the path on its own, it does not just append ../ to the current path, you can use pwd to get the Current Path, which reflects it

Utkarsh272: I have added a validate_path method, this will restrict users from exiting Default Path.

lind_shell.sh
Utkarsh272: I have added a validate_path method, this will restrict users from exiting Default Path for all the commands. Something like mkdir ../../test will also fail

lind_shell.sh
Utkarsh272: I have added the --help/-help command.

User can use it in two ways:
1. lind-wasm > --help : this will give a description of what this shell does, what all commands are supported.
2. lind-wasm > cd --help : this will give command description and how to use it.

rennergade:
@AayushA10 can you take over this PR? Utkarsh is no longer working with us for the time being.

Issue #42, opened 2024-11-19
Refactoring `sockaddr` Handling


Yaxuan-w:
## Description:

This issue tracks the refactoring of sockaddr handling and addresses the problems encountered with `accept_syscall`.

## Background:

Currently, `sockaddr` is initialized using the default Unix `GenSockaddr` borrowed from RustPOSIX, which has the largest memory space. This may cause issues with `getpeername`, as IPv4/IPv6 shouldn't have a path field.

It's been observed that byte-level operations should be preferred over struct-type checks.

In Linux, there are various `sockaddr` structures, such as `sockaddr_in` for IPv4 and` sockaddr_in6` for IPv6. However, tracing through the PostgreSQL source code shows that PostgreSQL uses sockaddr_storage, which is cast to `sockaddr` when necessary. The key difference between `sockaddr` and `sockaddr_storage` is size, with `sockaddr_storage` being large enough to accommodate all `sockaddr` types.

PostgreSQL's `accept` syscall passes a `sockaddr->family=0` with `sock_len=128`, making it impossible to determine the family based on size alone.

## Problem:

When using `accept`, `recvfrom`, `getsockname`, or `getpeername`, issues arise because the wrong `sockaddr` family is being inferred. The system receives a `NULL` `sockaddr,` and the `copy_out` function from RustPOSIX doesn't perform any operations.

## Current Patch

I made a temporary modification to `accept_syscall` to make RawPOSIX work for now: 
- Initialize a default `GenSockaddr` struct based on the `sockaddr` family received at the dispatcher stage.
- Handle the UNIX path conversion within the syscall itself.
- Additionally, in the `copy_out` function, the number of bytes to copy will be determined by comparing `initaddrlen` with the actual length of the structure (taking the minimum value), ensuring that no more than the reserved space is copied.

## Proposed Further Solution:

To resolve this, I suggest:

- Refining the `GenSockAddr` data structure. Directly allocate a buffer of size 128 bytes for syscalls like `accept`, `recvfrom`, `getsockname`, and `getpeername`.
- Implement a new function: Pass pointers to avoid `NULL` values being sent to syscalls.

Yaxuan-w:
I’d like to remind that after modifying the structure, we also need to update the relevant type conversion logic accordingly

Yaxuan-w:
Talking with @robinyuan1002 a bit about actual implementation here. 

Now we proposed two solutions:

## 1. Directly translate using `libc::sockaddr_un`

[...]

Issue #20, opened 2024-09-25
Fix getifaddrs syscall


qianxichen233:
The getifaddrs syscall is not working in lind-wasm since the memory allocation to ifaddrs struct seems to be wrong. The getifaddrs syscall takes a double pointers (struct ifaddrs **ifap), and it is underlying syscall that is responsible for allocating the memory for the ifap struct. In rawposix, getifaddrs_syscall treats  the ifap as an one-dimensional pointer (*mut u8), and fill the pointer with the ifaddrs from libc. The correct steps should be something like:
```
*buf = malloc_wasm_memory(...)
fill(*buf, &ifap_from_libc)
... (malloc_wasm_memory for next ifap struct)
```
The main difficuly with this approach is that, it has to malloc the wasm memory inside rawposix, but the wasm memory is managed by glibc, so it would be kind tricky to let rawposix malloc the wasm memory.

Another approach would be using the ```getifaddrs_internal``` function provided by glibc. This function does not actually use getifaddrs syscall directly, but instead used some other network syscalls to simulate the getifaddrs. Currently this approach would fail since it used bind syscall on socket type that is not directly supported by rawposix (AF_NETLINK socket). However, this is an issue in rawposix (https://github.com/Lind-Project/RawPOSIX/issues/3) and will be likely fixed soon.

qianxichen233:
Looks like `getifaddrs` is not a syscall but a glibc function. But we somehow has a `getifaddrs_syscall` in rawposix. This should be deleted I guess.

Turns out `getifaddrs` in glibc is internally using some socket types that we currently do not support, so I guess this issue is strongly associated with issue #42. Getting #42 fixed will probably also fix this issue

robinyuan1002:
Yeah, seem like that getifaddrs if a function and it may call socket, ioctl syscall. Should I Delete this syscall first?

rennergade:
Ah that makes sense. We handled getiffaddrs like this because we wanted to circumvent the weirder syscalls for popular paths.

Can we get an inventory of what syscalls getifaddrs uses and what kind of socket types? @robinyuan1002 

Also might make sense for you to work on #42 next so we can get to this.

rennergade:
> Yeah, seem like that getifaddrs if a function and it may call socket, ioctl syscall. Should I Delete this syscall first?

If you want to put in a PR just removing getifaddrs as is from RawPOSIX for now that would be fine as well.

robinyuan1002:
[...]

PR #101, opened 2025-02-05
delete getifaddrs

robinyuan1002:
When I tried to fix the [issue 20](https://github.com/Lind-Project/lind-wasm/issues/20), I found out that getifaddrs is not a syscall but a glibc function instead. According to our disscussion, we decided to delete getifaddrs from Rawposix and handle rest of mechanism through fixing sockaddr data structure, which has been mentioned in issue #42


qianxichen233 (changes requested):
I believe there is also `getifaddrs` test case in rawposix that should be deleted together. Specifically, https://github.com/Lind-Project/lind-wasm/blob/540dec5c6d9f969a567b3e7adb2d6f1a6455794c/src/RawPOSIX/src/tests/networking_tests.rs#L4272

JustinCappos:
> I believe there is also `getifaddrs` test case in rawposix that should be deleted together. Specifically,
> 
> https://github.com/Lind-Project/lind-wasm/blob/540dec5c6d9f969a567b3e7adb2d6f1a6455794c/src/RawPOSIX/src/tests/networking_tests.rs#L4272

Should this eventually get converted to an integration test where we actually run an app which calls this glibc function?

Yaxuan-w:
> Should this eventually get converted to an integration test where we actually run an app which calls this glibc function?

We've had that already: https://github.com/Lind-Project/lind-wasm/blob/main/tests/unit-tests/config_tests/non-deterministic/getifaddrs.c

robinyuan1002:
@rennergade this PR is ready to merge.

rennergade:
I believe we should also get rid of the `gen_netdevs` files and the `getifaddrs_from_file` function in comm.rs even though its currently commented out.

rennergade (changes requested):
I believe we should also get rid of the gen_netdevs files and the getifaddrs_from_file function in comm.rs even though its currently commented out.

Yaxuan-w:
FYI: gen_netdevs files files should locate on 

```
src/RawPOSIX/gen_netdevs.c
src/RawPOSIX/gen_netdevs.sh
```

robinyuan1002:
> I believe we should also get rid of the gen_netdevs files and the getifaddrs_from_file function in comm.rs even though its currently commented out.

It seems like there are no getifaddrs_from_file function in comm.rs.

rennergade (approved):

rennergade:
We can merge this once @qianxichen233 approves his change requests

qianxichen233 (approved):

PR #103, opened 2025-02-07
Ansal test report script

ansalma6:
PR for the python script that generates the test report.
We take all the files with .c extension. Test cases are considered deterministic or non deterministic by checking the parent folder name.
For determinstic tests, we compile and run using gcc, stores the output and then compiles using wasm and stores the output after running. The outputs are compared and is success if they match.
For non deterministic tests, we compile and using wasm only

All wasm compilations and runs are done using the lindtool.sh bash script

Finally a JSON file is generated with determinstic and non-deterministic as childs and each having the counts/lists of success/failure/timeout/segfaults and a list of all files and what their result was.


rennergade:
Off the bat the PR should be named something more clear, and it probably shouldnt have your name in it. (The branch name should be better as well but thats fine for now).

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
...
+                add_test_result(result, str(source_file), "Success", None, output)
+            elif retcode == 134 or retcode == 139:
+                add_test_result(result, str(source_file), "Failure", "Segmentation_Fault", output)
+            else:
+                add_test_result(result, str(source_file), "Failure", "Unknown_Failure", output)
+    finally:
+        if wasm_file and wasm_file.exists():
+            wasm_file.unlink()
+        if cwasm_file and cwasm_file.exists():
+            cwasm_file.unlink()
```
Yaxuan-w, line 238: Related to above

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
...
+#
+# Purpose:
+#   Compiles and runs a single test, 
+#   First compiles and runs using gcc, then wasm
+#   Finally compares the wasm output to the native GCC output to ensure they match.
+#   Logs results (success/failure/timeouts/seg faults)
+#
+# Variables:
+# - Input:
+#   source_file : The .c file path.
```
Yaxuan-w, line 251: Maybe using sth like `source test file` instead of `.c file` would be clearer

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
...
+import json
+import os
+import subprocess
+from pathlib import Path
+
+DEFAULT_TIMEOUT = 5
+
+JSON_OUTPUT = "results.json"
+SKIP_FOLDERS = [] # Add folders to be skipped, the test cases inside these will not run
+RUN_FOLDERS = [] # Add folders to be run, only test cases in these folders will run
```
Yaxuan-w, line 11: I'm a bit confused by how we determined which folders to run? Are those determined through user input?

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
...
+#   Updates 'result' dictionary.
+#
+# Exceptions:
+#   Recorded into 'result'.
+#
+# Note:
+#   Cleans up generated files(wasm/cwasm)
+#   Segmentation Fault is identified by return code of 134/139
+# ----------------------------------------------------------------------
+def test_single_file_non_deterministic(source_file, result, timeout_sec=DEFAULT_TIMEOUT):
```
Yaxuan-w, line 213: We should also run non-deterministic tests on native execution, right?

Non-deterministic means that the exact number / order of outputs might differ between Lind and native execution. However, we still need to verify whether the rest of the output is correct. e.g.: For cases where the output order varies, we can store the output in a list of strings and compare them accordingly.

If we only check for error messages, many failures might go undetected. For example, some syscalls could exhibit incorrect behavior without necessarily producing an error.

This approach will require additional effort since many tests will need unique python scripts. That is fine if we are only performing high-level detection here, but we should explicitly indicate in the command line at the beginning of this function what aspects are being tested and what aspects are not (maybe listed as `TODO`).

Yaxuan-w (changes requested):
This script is very detailed and includes the necessary comments, which is great! The structure is clear, and the logic is well-organized.

I have a few questions and some requests for changes.

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
+#!/usr/bin/env python3
+import json
```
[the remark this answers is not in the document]
ansalma6: Added Instructions

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
...
+# Function: run_compiled
+#
+# Purpose:
+#   Executes the compiled wasm/cwasm file using an external bash script 
+#   and returns the result code and filtered output.
+#
+# Variables:
+# - Input:
+#    wasm_file (Path): path to the .wasm 
+#    cwasm_file (Path): path to the .cwasm
```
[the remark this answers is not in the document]
ansalma6: The python script used run the commands for compiling and running directly. It was changed to call the ./lindtool.sh script to run those commands, the wasm and cwasm were verified to check if the compilation got over, since it is changed, the variables can be removed. 
Removed in the latest commit

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
...
+#
+# Exceptions:
+#   Catches TimeoutExpired and other Exceptions.
+#
+# Note:
+#   Dependancy on the script "./lindtool.sh run"
+#   Since the script outputs the command being run, we ignore 
+#   the first line in stdout by the script which is the command itself
+# ----------------------------------------------------------------------
+def run_compiled(wasm_file, cwasm_file, timeout_sec=DEFAULT_TIMEOUT):
```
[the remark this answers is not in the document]
ansalma6: Replied above

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
...
+# ----------------------------------------------------------------------
+# Function: compile_c_to_wasm
+#
+# Purpose:
+#   Given a path to a .c file, calls the lindtool script to compile it into 
+#   .wasm and .cwasm.
+#
+# Variables:
+# - Input: source_file - path to the .c file.
+# - Output: (wasm_file, cwasm_file, error_message).
```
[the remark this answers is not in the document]
ansalma6: Changed the cleanup to use the basefilename + extension for cleaning

wasmtestreport.py
ansalma6: Updated the approach for cleaning the files

wasmtestreport.py
ansalma6: Changed

wasmtestreport.py
ansalma6: Added instructions on how to run in the beginning

wasmtestreport.py
```diff
@@ -0,0 +1,422 @@
+#!/usr/bin/env python3
+import json
+import os
+import subprocess
+from pathlib import Path
+
+DEFAULT_TIMEOUT = 5
```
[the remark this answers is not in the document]
ansalma6: Added comments

wasmtestreport.py
ansalma6: There is a pending work on modifying the test cases to have better asserts to add details on how the test cases failed. Once that is done, we will have a better idea on how it got failed and it will be present in the report as well. That modification of test cases is a TODO and will be in another PR

ansalma6: Added a TODO

rennergade:
@m-hemmings and @yashaswi2000 can you review this today?

wasmtestreport.py
```diff
@@ -0,0 +1,482 @@
...
+# Exceptions:
+#   Recorded into 'result'.
+#
+# Note:
+#   Cleans up generated files(wasm/cwasm)
+#   Segmentation Fault is identified by return code of 134/139
+# ----------------------------------------------------------------------
+# TODO: Currently only running the test case in non deterministic cases, success means the compiled test case ran, need to add more specific tests
+# 
+def test_single_file_non_deterministic(source_file, result, timeout_sec=DEFAULT_TIMEOUT):
```
yashaswi2000, line 220: lets do a check if the test compiles and runs for now, once the testcase is stable we compare the result with native and wasm. and we probably need custom logic in determining PASSED or FAILED.

yashaswi2000 (approved):
LGTM

Yaxuan-w (approved):
LGTM! Thanks!

rennergade (approved):
Looks great, thank you!

PR #104, opened 2025-02-07
Implement dup3 syscall

robinyuan1002:
I have implement dup3 syscall, there are three functionality for dup3. First, dup3 have the functionality of dup2. Second, dup3 support O_CLOEXEC which is close newfd when it exec. Third, dup3 return error EINVAL if oldfd equal to newfd. Since there are issue with exec syscall right now I can only test the first and third functionality. The second functionality should work by logic but I need to wait for exec syscall.


src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -780,6 +780,44 @@ impl Cage {
         
     }
 
+    pub fn dup3_syscall(&self, old_virtualfd: i32, new_virtualfd: i32, flags: i32) -> i32 {
```
Yaxuan-w, line 783: Add comments about what functionality we have implemented and how

Yaxuan-w (changes requested):
Need to add test case

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
robinyuan1002: I still want wait until we fix the exec syscall. Then I will write a integration test case with this PR.

src/glibc/sysdeps/unix/sysv/linux/dup3.c
```diff
@@ -19,10 +19,11 @@
...
+#include <syscall-template.h>
 
 int
 __dup3 (int fd, int fd2, int flags)
 {
-  return 0;
+  return MAKE_SYSCALL(27, "syscall|dup3", (uint64_t) fd, (uint64_t) fd2, (uint64_t) flags, NOTUSED, NOTUSED, NOTUSED);
 }
-// weak_alias(__GI___dup3, __dup3)
+weak_alias (__dup3, dup3)
```
Yaxuan-w, line 29: add new line

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -780,6 +780,44 @@ impl Cage {
...
+                let _ret_kernelfd = unsafe{ libc::dup2(old_vfd.underfd as i32, new_kernelfd) };
+
+                if flags == O_CLOEXEC{   
+                    let _ = fdtables::get_specific_virtual_fd(self.cageid, new_virtualfd as u64, old_vfd.fdkind, new_kernelfd as u64, true, old_vfd.perfdinfo).unwrap();
+                    return new_virtualfd; 
+                }
+                else {
+                    let _ = fdtables::get_specific_virtual_fd(self.cageid, new_virtualfd as u64, old_vfd.fdkind, new_kernelfd as u64, false, old_vfd.perfdinfo).unwrap();
+                    return new_virtualfd;
+                }
```
JustinCappos, line 812: Why do you use this logic instead of making O_CLOEXEC the 4th parameter?

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
JustinCappos, line 783: Somewhere here, I'd recommend having a link to the manpage for the call.   I'd recommend this as a common pattern for all of the system calls, so this isn't meant to be specific...

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
robinyuan1002: I believe making O_CLOEXEC can simplify code better.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -802,14 +802,9 @@ impl Cage {
...
 
-                if flags == O_CLOEXEC{   
-                    let _ = fdtables::get_specific_virtual_fd(self.cageid, new_virtualfd as u64, old_vfd.fdkind, new_kernelfd as u64, true, old_vfd.perfdinfo).unwrap();
-                    return new_virtualfd; 
-                }
-                else {
-                    let _ = fdtables::get_specific_virtual_fd(self.cageid, new_virtualfd as u64, old_vfd.fdkind, new_kernelfd as u64, false, old_vfd.perfdinfo).unwrap();
-                    return new_virtualfd;
-                }
+                let _ = fdtables::get_specific_virtual_fd(self.cageid, new_virtualfd as u64, old_vfd.fdkind, new_kernelfd as u64, flags != 0, old_vfd.perfdinfo).unwrap();
```
JustinCappos, line 805: You'll need to comment this.   Also, you are changing the meaning here, right?   What if the flag is not O_CLOEXEC?   This code will do something different now, right?

robinyuan1002: I will comment this. The flags can only be O_CLOEXEC or 0 else it will return EINVAL(if statement in the front).

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -780,6 +780,46 @@ impl Cage {
...
+        }
+
+        match fdtables::translate_virtual_fd(self.cageid, old_virtualfd as u64) {
+            Ok(old_vfd) => {
+                let new_kernelfd = unsafe {
+                    libc::dup(old_vfd.underfd as i32)
+                };
+                let _ret_kernelfd = unsafe{ libc::dup2(old_vfd.underfd as i32, new_kernelfd) };
+
+                //right now flags can only be O_CLOEXEC or 0, by definition of O_CLOEXEC should_cleexec will be true(the value of O_CLOEXEC is a positive value)
```
rennergade, line 811: this comment probably makes sense to be with the CLOEXEC check at the beginning of the function

Yaxuan-w (commented):
We'd better to include test files at this point even things like `exec` would block testing.. This could help us test when other things has been fixed

Yaxuan-w:
I've fixed dup2 on #117. I think the implementation of dup3 could be something like:

```pseducode
check if oldfd == newfd

call dup2(oldfd, newfd)

call fdtables::set_cloexec(self.cageid, newfd, flag)
```

Yaxuan-w (approved):
LGTM now. Would be great to have a review from @rennergade

PR #107, opened 2025-02-10
Fix LIND_ROOT

Yaxuan-w:
Patch to correct `LIND_ROOT` variable

### DIFF
--- a/src/RawPOSIX/src/safeposix/dispatcher.rs
+++ b/src/RawPOSIX/src/safeposix/dispatcher.rs
@@ -1185,7 +1185,7 @@ pub fn lindrustinit(verbosity: isize) {
     // Set the first 3 fd to STDIN / STDOUT / STDERR
     // TODO:
     // Replace the hardcoded values with variables (possibly by adding a LIND-specific constants file)
-    let dev_null = CString::new("/home/lind-wasm/src/RawPOSIX/tmp/dev/null").unwrap();
+    let dev_null = CString::new("/home/lind/lind-wasm/src/RawPOSIX/tmp/dev/null").unwrap();
 
     // Make sure that the standard file descriptor (stdin, stdout, stderr) is always valid, even if they 
     // are closed before.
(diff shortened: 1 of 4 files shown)


qianxichen233 (changes requested):
there is also one in wasmtime, could you change that as well
https://github.com/Lind-Project/lind-wasm/blob/44504428c208eeb82c4f48a4b0e9ac8f6b74fb0a/src/wasmtime/crates/lind-multi-process/src/lib.rs#L25

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1185,7 +1185,7 @@ pub fn lindrustinit(verbosity: isize) {
     // Set the first 3 fd to STDIN / STDOUT / STDERR
     // TODO:
     // Replace the hardcoded values with variables (possibly by adding a LIND-specific constants file)
-    let dev_null = CString::new("/home/lind-wasm/src/RawPOSIX/tmp/dev/null").unwrap();
+    let dev_null = CString::new("/home/lind/lind-wasm/src/RawPOSIX/tmp/dev/null").unwrap();
```
JustinCappos, line 1188: This is part of the dispatcher so is going away, right?   Why is the prefix listed twice?

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -22,7 +22,7 @@ const ASYNCIFY_STOP_UNWIND: &str = "asyncify_stop_unwind";
 const ASYNCIFY_START_REWIND: &str = "asyncify_start_rewind";
 const ASYNCIFY_STOP_REWIND: &str = "asyncify_stop_rewind";
 
-const LIND_FS_ROOT: &str = "/home/lind-wasm/src/RawPOSIX/tmp";
+const LIND_FS_ROOT: &str = "/home/lind/lind-wasm/src/RawPOSIX/tmp";
```
JustinCappos, line 25: This should be a configuration setting instead of hard coded.

rennergade:
Is this PR still active? I think there was some discussion offline about it becoming part of the 3i changes.

Issue #17, opened 2024-09-13
Signals


rennergade:
Adding this issue as a tracker for implementing signals, which we'll pursue in phase 3.

rennergade:
This thread seems to be very helpful in describing the current state: https://github.com/WebAssembly/WASI/issues/166

rennergade:
Adding a few links I found while looking ahead a bit on this:

Compiling for signals - https://withbighair.com/webassembly/2024/08/26/WebAssembly-and-signals.html

StoreExt - https://docs.rs/wasmtime/latest/wasmtime/unix/trait.StoreExt.html#tymethod.set_signal_handler

Possible PoC - https://cocalc.com/github/bytecodealliance/wasmtime/blob/main/tests/custom_signal_handler.rs

rennergade:
Apparently exception handling is something being worked on which could be useful: https://github.com/WebAssembly/exception-handling

Also may be helpful for the problems with longjmp that @qianxichen233 reported.

rennergade:
ByteCodeAlliance recently merged this PR to wasmtime: https://github.com/bytecodealliance/wasmtime/pull/9614

Seems to have a potential to answer some of our problems here. @qianxichen233 can you try to get a handle on whats happening here and we'll discuss next week.

rennergade:
There seems to be a working version of signals using wasmer/wasix: https://github.com/wasix-org/wasix.org/blob/69a419132c797fcd6e02768355395e9215bdaf32/pages/docs/language-guide/c/tutorials/signal.mdx#L5

This seems like its probably our best roadmap further if we can dig into whats happening here behind the scenes.

rennergade:
Per @qianxichen233 and my chat with Jonathan it seems like we should figure out exactly what weird cases postgres actually introduces with signals. I'm going to go back and run through the native code to make sure I'm correct that in my assumption that postgres both calls fork() and longjmp from handlers.

qianxichen233:
If longjmp() is guaranteed to be called after fork(), then using the detached thread to run the signal handler might work, since longjmp does not require to suspend the main process as the state is already captured via setjmp previously.

[...]

PR #109, opened 2025-02-12
Signal Implementation

qianxichen233:
This PR adds basic signal implementation to lind-wasm (issue #17). For signal implementation designs, refer to [this doc](https://docs.google.com/document/d/1AYB3bYZJtnqQerE3iwez8VPDumUEDmQ17V5nAlrkLwY/edit?usp=sharing) for more detail

This PR is coupled with PR #110. PR #110 is migrating Binaryen codebase together with the binary rewriting (signal's epoch injection), while this PR are mainly for the changes made in glibc/wasmtime/rawposix.


rennergade:
Can you add a bit more detailed summary of the changes to the PR description, maybe broken apart between libc/wasmtime/rawposix?

This is a pretty big PR so maybe we should divide and conquer with the review here. My thoughts are to have @ChinmayShringi  do an initial review on the RawPOSIX parts and then @Yaxuan-w do a secondary review there. As for libc/wasmtime @robinyuan1002 can do an initial review and @yzhang71 do a secondary review. Then @m-hemmings and I can go over everything, and finally Justin if we think thats warranted.

Fine with other suggestions but I think that may work well.

Yaxuan-w:
> My thoughts are to have @ChinmayShringi do an initial review on the RawPOSIX parts and then @Yaxuan-w do a secondary review there

This works for me!

src/RawPOSIX/src/constants/sys_constants.rs
ChinmayShringi, line 1: Should signals have its own seperate constants file?

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,234 @@
...
+
+const EPOCH_NORMAL: u64 = 0;
+const EPOCH_SIGNAL: u64 = 1;
+const EPOCH_KILLED: u64 = 2;
+
+// switch the epoch of the main thread of the cage to "signal" state
+pub fn signal_epoch_trigger(cageid: u64) {
+    let cage = cagetable_getref(cageid);
+    let main_threadid = cage.main_threadid.load(RustAtomicOrdering::Relaxed) as i32;
+    let epoch_handler = cage.epoch_handler.get(&main_threadid).unwrap();
```
ChinmayShringi, line 11: I think this may panic if main_threadid is not found/invalid/undefined, maybe something like:
`let epoch_handler = cage.epoch_handler.get(&main_threadid).expect("Main thread epoch handler not found");
`

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,234 @@
...
+    let main_threadid = cage.main_threadid.load(RustAtomicOrdering::Relaxed) as i32;
+    // we iterate through the epoch handler of each thread in the cage
+    for entry in cage.epoch_handler.iter() {
+        if entry.key() == &main_threadid {
+            // main thread should be the one invoking this method
+            // so main thread could kill itself and we do not need to notify it again
+            continue;
+        }
+        let epoch_handler = entry.value();
+        let guard = epoch_handler.write();
```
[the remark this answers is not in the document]
qianxichen233: what issue are you refering to? I think the write lock is released at the end of each iteration of loop, so there shouldn't be any deadlock issue

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,234 @@
...
+        let epoch = *guard;
+        unsafe {
```
yzhang71, line 33: The unsafe block should be accompanied by a comment explaining why the operation is sound. Such as:
`// SAFETY: We ensure that the pointer stored in epoch is valid, and that no other thread will concurrently modify the data.`

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1031,12 +1015,80 @@ pub fn lind_syscall_api(
...
             let sig = arg2 as i32;
             interface::cagetable_getref(cageid)
                 .kill_syscall(cage_id, sig)
-        } 
+        }
+
+        SIGPROCMASK_SYSCALL => {
+            let cage = interface::cagetable_getref(cageid);
+            let how = arg1 as i32;
+            let mut set = {
```
yzhang71, line 1062: I think we need some comments here, and the variable names are not informative; for example, 'set' is too confusing.

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1031,12 +1015,80 @@ pub fn lind_syscall_api(
...
+                        },
+                        Err(errno) => return syscall_error(errno, "sigaction", "invalid address"),
+                    }
+                }
+            };
+            let mut old_set = {
+                if arg3 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg3, std::mem::size_of::<interface::SigsetType>(), PROT_WRITE) {
```
ChinmayShringi, line 1079: I think this function was removed due to too much overhead `check_and_convert_addr_ext` check impl for `translate_vmmap_addr` it adds the start address

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1206,6 +1258,41 @@ pub fn lind_syscall_api(
...
 
+        SETITIMER_SYSCALL => {
+            let cage = interface::cagetable_getref(cageid);
+            let which = arg1 as i32;
+
+            let itimeval = {
+                if arg2 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg2, std::mem::size_of::<interface::ITimerVal>(), PROT_READ) {
```
ChinmayShringi, line 1269: I think this function was removed due to too much overhead `check_and_convert_addr_ext` check impl for `translate_vmmap_addr` it adds the start address

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,234 @@
+use crate::{constants::{SA_NODEFER, SA_RESETHAND, SIG_DFL}, interface::{cagetable_getref, cagetable_getref_opt, RustAtomicOrdering}};
+
+const EPOCH_NORMAL: u64 = 0;
+const EPOCH_SIGNAL: u64 = 1;
+const EPOCH_KILLED: u64 = 2;
```
ChinmayShringi, line 5: Should these be under the constants folder?

ChinmayShringi:
@Yaxuan-w I have reviewed the RawPosix part of this PR

src/RawPOSIX/src/interface/signal.rs
qianxichen233: These are not system constants, it's my own defined constants for epoch implementation that will only be used within this file (at least for now). So I think it makes sense to me to put the constants here without making them public.

JustinCappos: What are these?   Is any value just as meaningful as any other?   If so, I'd suggest considering using some weird value like 0x82fa329c... so that if you end up seeing them appear in an unexpected place you can find that value in the code and know where it came from.

qianxichen233: These are the epoch state constants that the injected epoch check is checking for. The injected check logic in wasm binary is `if epoch > 0: call callback()`, so only `EPOCH_NORMAL` must be `0` and other two states can indeed be some other arbitrary values greater than 0

src/RawPOSIX/src/safeposix/cage.rs
```diff
@@ -63,10 +63,15 @@ pub struct Cage {
...
-    // block / unblock / replace the signal mask for a thread.
-    pub sigset: interface::RustHashMap<u64, interface::RustAtomicU64>,
+    // sigset is an atomic signal sets representing the signals 
+    // currently blocked for the cage. Interacts with sigprocmask_syscall() to 
+    // block / unblock / replace the signal mask for a the cage.
+    pub sigset: interface::RustAtomicU64,
+    // pending_signals are signals that are pending to be handled
+    pub pending_signals: interface::RustLock<Vec<i32>>,
+    // epoch_handler is a hash map where key is the thread id of the cage, and the value is the epoch handler
+    // address
```
Yaxuan-w, line 73: Add a bit more detailed explanation about the epoch_handler address is used on injected check in wasm binary

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1031,12 +1015,80 @@ pub fn lind_syscall_api(
...
         }
 
+        SIGACTION_SYSCALL => {
+            let cage = interface::cagetable_getref(cageid);
+            let sig = arg1 as i32;
+            let mut sigaction = {
+                if arg2 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg2, std::mem::size_of::<interface::SigactionStruct>(), PROT_WRITE) {
```
Yaxuan-w, line 1025: should remove

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1031,12 +1015,80 @@ pub fn lind_syscall_api(
...
+                        },
+                        Err(errno) => return syscall_error(errno, "sigaction", "invalid address"),
+                    }
+                }
+            };
+            let mut old_sigaction = {
+                if arg3 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg3, std::mem::size_of::<interface::SigactionStruct>(), PROT_WRITE) {
```
Yaxuan-w, line 1038: should remove

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1031,12 +1015,80 @@ pub fn lind_syscall_api(
...
+        }
+
+        SIGPROCMASK_SYSCALL => {
+            let cage = interface::cagetable_getref(cageid);
+            let how = arg1 as i32;
+            let mut set = {
+                if arg2 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg2, std::mem::size_of::<interface::SigsetType>(), PROT_WRITE) {
```
Yaxuan-w, line 1066: should remove

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1206,6 +1258,41 @@ pub fn lind_syscall_api(
...
+                        Err(errno) => return syscall_error(errno, "setitimer", "invalid itimeval struct address"),
+                    }
+                }
+            };
+
+            let old_itimeval = {
+                if arg3 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg3, std::mem::size_of::<interface::ITimerVal>(), PROT_WRITE) {
```
Yaxuan-w, line 1283: should remove

src/glibc/lind_syscall/crt1/crt1.c
```diff
@@ -186,6 +186,19 @@ int __main_void(void) {
...
 
+// define an unused function pointer
+// the reason for this is that function pointer are translated to function index in WebAssembly
+// which starts from 1. But in native environment, function pointer are real address so they are
+// assumed that it cannot be 1. Hence, some constant value like SIG_IGN will have conflict with function
+// indexes in WebAssembly.
+// The workaround we are using here is to define an unused function pointer to occupy the function index of 1
+// so that normal function pointer could have their index starting from 2.
+int __unused_function_pointer() {
+    return 42;
```
Yaxuan-w, line 197: Does the number 42 have a special meaning?

tests/unit-tests/signal_tests/deterministic/alarm.c
```diff
@@ -0,0 +1,31 @@
...
+    sa.sa_flags = 0;
+    sigemptyset(&sa.sa_mask);
+    sigaction(SIGALRM, &sa, NULL);
+
+    printf("Setting an alarm to trigger in 1 seconds...\n");
+    alarm(1);  // First alarm triggers after 2 seconds
+
+    while (1) {
+        // Wait for signals
+    }
```
Yaxuan-w, line 28: This test won't end automatically. Maybe we could add a counter to stop?

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1206,6 +1258,41 @@ pub fn lind_syscall_api(
...
+            let which = arg1 as i32;
+
+            let itimeval = {
+                if arg2 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg2, std::mem::size_of::<interface::ITimerVal>(), PROT_READ) {
+                        Ok(addr) => match interface::get_constitimerval(addr) {
+                            Ok(itimeval) => itimeval,
+                            Err(_) => return syscall_error(Errno::EFAULT, "socketpair", "failed to get socket pair array"),
```
Yaxuan-w, line 1272: should be `setitimer`

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1206,6 +1258,41 @@ pub fn lind_syscall_api(
...
+            };
+
+            let old_itimeval = {
+                if arg3 == 0 {
+                    None
+                } else {
+                    match check_and_convert_addr_ext(&cage, arg3, std::mem::size_of::<interface::ITimerVal>(), PROT_WRITE) {
+                        Ok(addr) => match interface::get_itimerval(addr) {
+                            Ok(itimeval) => itimeval,
+                            Err(_) => return syscall_error(Errno::EFAULT, "socketpair", "failed to get socket pair array"),
```
Yaxuan-w, line 1286: should be setitimer

src/wasmtime/crates/wasmtime/src/runtime/store.rs
```diff
@@ -1372,6 +1389,34 @@ impl<'a, T> StoreContextMut<'a, T> {
         }
     }
 
+    // append the signal callstack information
+    pub fn append_signal_asyncify_data(&mut self, signal_handler: i32, signo: i32) {
+        self.0.signal_asyncify_data.push(SignalAsyncifyData { signal_handler, signo });
+    }
+
+    // pop the signal callstack information
+    pub fn pop_signal_asyncify_data(&mut self, signal_handler: i32, signo: i32) {
```
Yaxuan-w, line 1398: I'm curious why we pass `signal_handler` and `signo` here?

Yaxuan-w (changes requested):
Awesome!!

src/wasmtime/crates/wasmtime/src/runtime/store.rs
qianxichen233: This is used for supporting Asyncify on signal handler. It is essentially not part of logic used for signal handling

src/glibc/lind_syscall/crt1/crt1.c
qianxichen233: no. it's just a random number inside this unused function

src/RawPOSIX/src/interface/signal.rs
qianxichen233: `expect` is essentially a panic but with an error message. Ideally we should never encounter panics here, but replace with `expect` for more verbose debug information is always good

src/RawPOSIX/src/constants/sys_constants.rs
qianxichen233: I think sys_constants.rs is used for storing constants for system related syscalls and signal is within the range. Actually currently 90% of this file are all about signal related constants, if we are moving out signal constants, this file would be almost empty

rennergade: Agreed fine to keep this here.

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,243 @@
...
+    // However, Potential BUG (TODO): We still need to verify the lifetime of the pointer. This pointer
+    // is created by wasmtime and will be destroyed at some point when the wasm instance is destroyed
+    // we still need to figure out when is the destroy happening and make sure it is destroyed after the
+    // information in rawposix is updated
+    unsafe {
+        *epoch = EPOCH_SIGNAL;
+    }
+}
+
+// swtich the epoch of all threads of the cage to "killed" state
```
rennergade, line 24: *switch

src/RawPOSIX/src/interface/signal.rs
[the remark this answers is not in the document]
rennergade: is it possible for multiple threads to access this function simultaneously?

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,243 @@
...
+// send specified signal to the cage, return value indicates whether the cage exists
+pub fn lind_send_signal(cageid: u64, signo: i32) -> bool {
+    if let Some(cage) = cagetable_getref_opt(cageid) {
+        let mut pending_signals = cage.pending_signals.write();
+        // TODO: currently we are queuing the same signals instead of merging the same signal
+        // this is different from linux which always merge the same signal if they havn't been handled yet
+        // we queue the signals for now because our epoch based signal implementation could have much longer
+        // gap for signal checkings than linux. We need to finally decide whether do the queuing or merging
+        // in the future, probably based on some experimental data
+        pending_signals.push(signo);
```
rennergade, line 117: so if a signal is blocked it still gets pushed to this list but the epoch isn't triggered?

what happens if another signal triggers the epoch but this blocked signal is in the queue?

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,243 @@
...
+                
+                // restorer is called when the signal handler finishes. It should restore the signal mask
+                let restorer = Box::new(move |cageid| {
+                    let cage = cagetable_getref(cageid);
+                    cage.sigset.store(sigset, RustAtomicOrdering::Relaxed);
+                });
+                Some((signo, signal_handler, restorer))
+            }
+            None => {
+                // retrieve the signal handler
```
rennergade, line 178: im confused at whats happening here, its getting a handler in this case even though no handler is set?

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,243 @@
...
+            }
+        }
+    } else {
+        // if there is no pending unblocked signal, we return None
+        None
+    }
+}
+
+// check if there is any pending unblocked signals
+pub fn lind_check_no_pending_signal(cageid: u64) -> bool {
```
rennergade, line 194: confusing function name, comment needs a bit more of whats happening here and retval

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,243 @@
...
+
+// initialize the signal for a new thread
+pub fn lind_signal_init(cageid: u64, epoch_handler: *mut u64, threadid: i32, is_mainthread: bool) {
+    let cage = cagetable_getref(cageid);
+
+    // if this is specified as the main thread, then replace the main_threadid field in cage
+    if is_mainthread {
+        cage.main_threadid.store(threadid as u64, RustAtomicOrdering::SeqCst);
+    }
+    let epoch_handler = super::RustLock::new(epoch_handler);
```
rennergade, line 217: are we still using the RustLock type elsewhere?

src/glibc/lind_syscall/crt1/crt1.c
rennergade: I appreciate the Hitchhiker's reference but agree that there should be a small comment that its not a significant value.

src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -131,7 +131,12 @@ pub fn add_to_linker<T: LindHost<T, U> + Clone + Send + 'static + std::marker::S
             let host = caller.data().clone();
             let ctx = get_cx(&host);
 
-            ctx.lind_syscall(call_number, call_name, &mut caller, arg1, arg2, arg3, arg4, arg5, arg6)
+            let retval = ctx.lind_syscall(call_number, call_name, &mut caller, arg1, arg2, arg3, arg4, arg5, arg6);
+            
+            // TODO: add a signal check here as Linux also has a signal check when transition from kernel to userspace
```
rennergade, line 136: is this the return springboard from wasmtime to user space? I think we definitely need to check here.

tests/unit-tests/signal_tests/deterministic/alarm.c
```diff
@@ -0,0 +1,34 @@
+// test for alarm function
```
rennergade, line 1: We should probably have a signal test for each signal type. A bit tedious but will be useful once the CI kicks in.

rennergade: We can create a separate issue for this if you want though, don't need to handle it here explicitly.

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -327,6 +333,16 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
...
                     }).unwrap();
+                
+                // retrieve the epoch global
+                let lind_epoch = instance
+                    .get_export(&mut store, "epoch")
+                    .and_then(|export| export.into_global())
+                    .expect("Failed to find shared_global");
+                // retrieve the handler (underlying pointer) for the epoch global
+                let pointer = lind_epoch.get_handler(&mut store);
+                // initialize the signal for the main thread of forked cage
```
rennergade, line 344: comment what "true is indicating"

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -383,27 +400,30 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
...
-                                *val as u64,
-                                0,
-                                0,
-                                0,
-                                0,
-                                0,
-                            );
-                            // let _ = on_child_exit(*val);
+                            // exit the thread
+                            if rawposix::interface::lind_thread_exit(child_cageid, 1) {
```
rennergade, line 404: what does 1 signify?

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -556,6 +578,16 @@ impl<T: Clone + Send + 'static + std::marker::Sync, U: Clone + Send + 'static +
                     .unwrap();
                 let _ = stack_pointer_setter.call(&mut store, (stack_addr - offset) as i32);
 
+                // retrieve the epoch global
```
rennergade, line 581: basically the same code as above? probably better to functionize this?

src/wasmtime/crates/lind-multi-process/src/signal.rs
```diff
@@ -0,0 +1,165 @@
...
+
+// default signal handler actions
+enum SignalDefaultHandler {
+    Terminate,  // terminate the process
+    Ignore,     // ignore the signal
+    Stop,       // stop the current process
+    Continue,   // resume the stopped process
+}
+
+// handle all the epoch callback
```
rennergade, line 14: I think a long important function like this needs a more detailed comment at the top explaining what its going to do

src/wasmtime/crates/lind-multi-process/src/signal.rs
```diff
@@ -0,0 +1,165 @@
...
+    // because this is the trap type used by wasmtime's built-in epoch
+    // and epoch is the only possible source of this type of trap
+    let err = Trap::Interrupt;
+    unsafe {
+        raise_trap(err.into());
+    }
+    unreachable!();
+}
+
+// maps each signal to its default handler
```
rennergade, line 129: probably link to the posix signal defaults here and explain if anything is different than that spec

src/wasmtime/src/commands/run.rs
```diff
@@ -211,21 +215,26 @@ impl RunCommand {
...
-                    0,
-                    code as u64,
-                    0,
-                    0,
-                    0,
-                    0,
-                    0,
-                );
+                // exit the thread
+                if rawposix::interface::lind_thread_exit(1, 1) {
```
rennergade, line 219: need to now significance of 1,1
avoid magic numbers

src/wasmtime/src/commands/run.rs
```diff
@@ -581,6 +590,18 @@ impl RunCommand {
...
+                let lind_epoch = instance
+                    .get_export(&mut *store, "epoch")
+                    .and_then(|export| export.into_global())
+                    .expect("Failed to find epoch global export!");
+
+                // retrieve the handler (underlying pointer) for the epoch global
+                let pointer = lind_epoch.get_handler(&mut *store);
+
+                // initialize the signal for the main thread of the cage
+                rawposix::interface::lind_signal_init(pid, pointer as *mut u64, 1, true);
```
rennergade, line 603: again avoid magic numbers

rennergade (changes requested):
Had some comments about clarification of functionality and requests for some more thorough comments some places.

Awesome work here.

src/RawPOSIX/src/interface/signal.rs
qianxichen233: the blocked signals will be skipped when retrieve the signal in the pending list, see `lind_send_signal`

src/RawPOSIX/src/interface/signal.rs
qianxichen233: if there is no signal handler, `signal_get_handler` will return SIG_DFL

src/wasmtime/crates/lind-common/src/lib.rs
qianxichen233: This is when a syscall just completed and ready to return to user space. We could add the check here to more closely mimic linux later, but I think this shouldn't be required for signal to work, like if we do not check here, after it went back to user space, it will be triggered by epoch as well very soon

src/wasmtime/crates/lind-multi-process/src/lib.rs
qianxichen233: thread id, which always start from 1

src/RawPOSIX/src/interface/signal.rs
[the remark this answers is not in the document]
qianxichen233: Currently only one thread of the same cage should be able to access this function since this function is only invoked by signal termination, which is always done by the main thread. But later when we fixed the exec/exit issue, this will then probably be accessed by multiple threads of the same cage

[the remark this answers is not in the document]
rennergade: Lets note that it needs to be called by only the main thread to avoid deadlock. When were figuring out exec/exit termination we can probably figure out a way for only the main thread to call it as well.

src/RawPOSIX/src/interface/signal.rs
rennergade: okay this needs to be explained in the comments here since its not apparent

src/wasmtime/crates/lind-common/src/lib.rs
rennergade: if theres multiple syscalls in a row without reaching a loop or function header this would probably be inefficient.

But I think moreso when we interrupt blocking syscalls they'll need to return EINTR which infers the signal has been handled, so this could lead to incorrect behavior

src/wasmtime/crates/lind-multi-process/src/lib.rs
rennergade: need to either comment this or probably better have a THREADID_START macro or something like that

src/RawPOSIX/src/interface/signal.rs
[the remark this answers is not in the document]
qianxichen233: yeah, but I just thought this a little bit, I think even if this function is accessed by multiple threads of same cage, it shouldn't cause any deadlock, since every lock on each thread's epoch is accessed by a very short amount of time (like just taking the lock, set the value, then the lock is released), I didn't see any potential situation a deadlock could happen

rennergade (dismissed):
This looks good from my end now.

rennergade:
@Yaxuan-w @yzhang71 @ChinmayShringi if you all can give this a second look and approve if you think your changes have been addressed/no additional issues that would be great!

src/RawPOSIX/src/constants/sys_constants.rs
```diff
@@ -78,10 +80,28 @@ pub const SIGIO: i32 = 29;          // I/O now possible
...
+pub const SA_NOCLDWAIT: u32 = 0x00000002;       // Don't create zombie on child death
+pub const SA_SIGINFO: u32 = 0x00000004;         // Signal handler with SA_SIGINFO args
+pub const SA_UNSUPPORTED: u32 = 0x00000400;     // Unsupported
+pub const SA_EXPOSE_TAGBITS: u32 = 0x00000800;  // exposes an architecture-defined set of tag bits in siginfo.si_addr
+pub const SA_ONSTACK: u32 = 0x08000000;         // Take signal on signal stack
+pub const SA_RESTART: u32 = 0x10000000;         // Restart syscall on signal return
+pub const SA_NODEFER: u32 = 0x40000000;         // Don't automatically block the signal when its handler is being executed
+pub const SA_RESETHAND: u32 = 0x80000000;       // Reset to SIG_DFL on entry to handler
+
+// Special Signal Handlers
```
JustinCappos, line 101: This is also from src/glibc/target/include/asm-generic/signal-defs.h, right?

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+// check the specified thread with specified cage is in "killed" state
+pub fn thread_check_killed(cageid: u64, thread_id: u64) -> bool {
+    let cage = cagetable_getref(cageid);
+    let epoch_handler = cage.epoch_handler.get(&(thread_id as i32)).unwrap();
+    let guard = epoch_handler.write();
+    let epoch = *guard;
+    // SAFETY: see comment at `signal_epoch_trigger`
```
JustinCappos, line 51: I like this style.   Good job commenting the code to explain what is happening / what may not work.

src/RawPOSIX/src/interface/timer.rs
```diff
@@ -111,7 +114,7 @@ impl IntervalTimer {
                         .saturating_sub(guard.start_instant.elapsed());
 
                     if remaining_seconds == RustDuration::ZERO {
-                        lind_kill_from_id(guard.cageid, 14);
+                        lind_send_signal(guard.cageid, SIGALRM);
```
JustinCappos, line 117: Perhaps a short comment explaining why you're doing this (i.e. why you're using a signal at all to set up a timer) would be helpful.   I think this is largely for compatibility / interruptibility of timers, etc. but just writing 2-3 sentences somewhere here may give context.

src/glibc/lind_syscall/crt1/crt1.c
JustinCappos, line 197: What is this?   Why this magic number?

src/wasmtime/crates/wasmtime/src/runtime/vm.rs
```diff
@@ -30,7 +30,7 @@ mod send_sync_ptr;
 mod store_box;
 mod sys;
 mod table;
-mod traphandlers;
+pub mod traphandlers;
```
JustinCappos, line 33: Is this change intentional?

JustinCappos (approved):
This looks good!   I have a few minor questions / comments, but it's ready to merge, I think.

src/RawPOSIX/src/constants/sys_constants.rs
qianxichen233: yes

src/wasmtime/crates/wasmtime/src/runtime/vm.rs
qianxichen233: yes, we need to expose this function for our signal crates' use (to perform a thread suicide)

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -0,0 +1,245 @@
...
+
+// manually check if the epoch is not in "normal" state
+// useful if we want to do our own epoch check in host
+pub fn signal_check_trigger(cageid: u64) -> bool {
+    let cage = cagetable_getref(cageid);
+    let main_threadid = cage.main_threadid.load(RustAtomicOrdering::Relaxed) as i32;
+
+    let epoch_handler = cage.epoch_handler.get(&main_threadid).unwrap();
+    let guard = epoch_handler.write();
+    let epoch = *guard;
```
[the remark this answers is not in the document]
qianxichen233: it will not be dropped if I am just dereferencing it. Dereference is just getting the underlying data from pointer, there is no even a transfer of ownership happening here. So guard is still valid after the dereference, so it is not dropped

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -904,12 +888,82 @@ pub fn lind_syscall_api(
...
+                        Ok(addr) => match interface::get_sigactionstruct(addr) {
+                            Ok(val) => val,
+                            Err(errno) => return syscall_error(Errno::EFAULT, "sigaction", "invalid sigaction struct format"),
+                        },
+                        Err(errno) => return syscall_error(errno, "sigaction", "invalid address"),
+                    }
+                }
+            };
+
+            interface::cagetable_getref(cageid)
```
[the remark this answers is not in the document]
qianxichen233: That's a good catch. Fixed

src/RawPOSIX/src/interface/signal.rs
```diff
@@ -289,3 +289,18 @@ pub fn lind_thread_exit(cageid: u64, thread_id: u64) -> bool {
...
+// trigger the epoch if pending signal list is not empty
+// This function is invoked only by a newly exec-ed cage 
+// immediately after it completes its initialization. 
+// Its purpose is to handle the scenario where Linux resets 
+// the signal mask but preserves pending signals after exec. 
+// As a result, the new process may receive signals that were 
+// pending in the previous process right after it starts.
+pub fn signal_may_trigger(cageid: u64) {
+    let cage = cagetable_getref(cageid);
+    let pending_siangls = cage.pending_signals.read();
```
rennergade, line 302: this looks like a typo

rennergade:
@qianxichen233 I think you need to pull in the latest format merge. Can you also double check that you've run cargo fmt on these changes as well?

rennergade (approved):
Approved once the formatting is clear.

rennergade (approved):
This is ready to merge once @Yaxuan-w approves.

Yaxuan-w (approved):

PR #110, opened 2025-02-12
upload modified Binaryen binary

qianxichen233:
Migrated Binaryen codebase into lind-wasm for signal epoch injection usage.

most of the code are directly from Binaryen, the only modification I made are `lindtool.sh` and `src/passes/EpochInjection.cpp`


rennergade:
Is there any way to do this without storing the entire binaryen codebase. Id love to avoid that.

JustinCappos:
Agreed.   We'd like to be up-to-date with changes / fixes to the project.

qianxichen233:
> Is there any way to do this without storing the entire binaryen codebase. Id love to avoid that.

Two possible ways I can think of are
1. Considering that we should rarely need to modify binaryen, we could just upload the pre-built `wasm-opt` binary. This might lose some compatibility if lind is compiled and used in different environment other than our current docker image.
2. Only upload the `EpochInjection.cpp` file, together with a shell script that automatically fetches Binaryen codebase, and move the `EpochInjection.cpp` into it with some other minor modification to the source code to make Binaryen accepts `EpochInjection.cpp` and then do the compilation

JustinCappos:
Do we need to change it or just use it?   If we use it, can we make this a
submodule or similar?  (Maybe ask Partick or Aditya in the lab about how
they think you should handle this?)

rennergade:
> > Is there any way to do this without storing the entire binaryen codebase. Id love to avoid that.
> 
> Two possible ways I can think of are
> 
>     1. Considering that we should rarely need to modify binaryen, we could just upload the pre-built `wasm-opt` binary. This might lose some compatibility if lind is compiled and used in different environment other than our current docker image.
> 
>     2. Only upload the `EpochInjection.cpp` file, together with a shell script that automatically fetches Binaryen codebase, and move the `EpochInjection.cpp` into it with some other minor modification to the source code to make Binaryen accepts `EpochInjection.cpp` and then do the compilation

I think option 1 should suffice for now as long as we mark this as a future issue.

qianxichen233:
Updated, only binaries are uploaded now

lindtool.sh
```diff
@@ -1,20 +1,24 @@
 #!/bin/bash
+LIND_WASM_DEFAULT_BASE=/home/lind-wasm
```
Yaxuan-w, line 2: should be `/home/lind/lind-wasm`

lindtool.sh
```diff
@@ -1,20 +1,24 @@
...
-rustposix_base="$PWD/src/safeposix-rust"
-rawposix_base="$PWD/src/RawPOSIX"
+glibc_base="$LIND_WASM_BASE/src/glibc"
+wasmtime_base="$LIND_WASM_BASE/src/wasmtime"
+rustposix_base="$LIND_WASM_BASE/src/safeposix-rust"
+rawposix_base="$LIND_WASM_BASE/src/RawPOSIX"
+
+wasm_opt_base="$LIND_WASM_BASE/tools/binaryen/bin/wasm-opt"
 
 CC="${CLANG:=/home/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/clang"
```
Yaxuan-w, line 12: same here

lindtool.sh
```diff
@@ -1,20 +1,24 @@
 #!/bin/bash
+LIND_WASM_DEFAULT_BASE=/home/lind-wasm
+LIND_WASM_BASE="${LIND_WASM_BASE:-/home/lind-wasm}"
```
Yaxuan-w, line 3: same here

lindtool.sh
```diff
@@ -103,7 +109,7 @@ compile_src() {
 }
 
 pmode=0
-if [ "${!#}" = "-p" ]; then
+if [ "${!#}" = "-p" ]; then    git push --set-upstream origin migrate-binaryen-bin
```
Yaxuan-w, line 112: I'm wondering why we need push to remote?

Yaxuan-w (changes requested):

rennergade:
> Updated, only binaries are uploaded now

Could you provide some context as to what all these binaries are? I was under the impression that it was only one or two

lindtool.sh
qianxichen233: it's something I accidentally copied, just deleted

qianxichen233:
> > Updated, only binaries are uploaded now
> 
> Could you provide some context as to what all these binaries are? I was under the impression that it was only one or two

Yeah, most of them are just directly copied from binaryen's bin folder. I just deleted all of them since they are not used by us. Only `wasm-opt` is left

rennergade:
@qianxichen233 what is libgtest?

qianxichen233:
`wasm-opt` needs its `lib` folder to be able to run so I just copied the entire lib folder here, I am not sure which files will possibly be needed by `wasm-opt`

rennergade:
> `wasm-opt` needs its `lib` folder to be able to run so I just copied the entire lib folder here, I am not sure which files will possibly be needed by `wasm-opt`

Sorry its a bit annoying but its probably worth taking a few minutes to know whats required here instead of us merging unknown binaries.

rennergade (approved):
This looks good now.

Tagging @m-hemmings since this will interface with buids. Not sure if he'll have to change anything.

rennergade:
Also ready.

Yaxuan-w (approved):

PR #111, opened 2025-02-17
Add mprotect syscall

ChinmayShringi:
## Description

Adds test cases for the mprotect syscall implementation to verify memory protection functionality.

 

### Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- `cargo build`
- `cargo test ut_lind_fs_mprotect_unmapped_addr`

Added the following test cases in `src/RawPOSIX/src/tests/fs_tests.rs`:

- `ut_lind_fs_mprotect_unmapped_addr`: Tests error handling for unmapped addresses
  - Attempts to protect unmapped memory region
  - Verifies ENOMEM error is returned

## Checklist:

- [x] Test cases follow existing patterns in the codebase
- [x] Tests verify both successful operations and error conditions
- [x] All tests clean up resources properly


rennergade:
A few things off the back:

Title should be adding mprotect since youre adding the syscall, not just test cases

We need to at least update the vmmap region after a successful mprotect call, though I'm realizing there are two things we probably want to discuss as a group

A) Do we want to disallow mprotects to non-allocated regions? POSIX technically allows this but is labeled undefined behavior, allowing this could potentially be screwy for us and I don't believe allowing it would remove any compatibility

B) Do we want to disallow using PROT_EXEC?

rennergade:
> A few things off the back:
> 
> Title should be adding mprotect since youre adding the syscall, not just test cases
> 
> We need to at least update the vmmap region after a successful mprotect call, though I'm realizing there are two things we probably want to discuss as a group
> 
> A) Do we want to disallow mprotects to non-allocated regions? POSIX technically allows this but is labeled undefined behavior, allowing this could potentially be screwy for us and I don't believe allowing it would remove any compatibility
> 
> B) Do we want to disallow using PROT_EXEC?

@ChinmayShringi after discussing this with Qianxi I was mistaken, mprotects to unallocated regions should return ENOMEM. As for PROT_EXEC this isn't even possible with WASM so we'll disallow the host from doing it.

So you need to add a function (in mem.rs) before the syscall that checks if the region is in the vmmap, and if not return ENOMEM, also if PROT_EXEC is being attempted return EINVAL like in the mmap_handler function.

After the syscall, if the host mprotect was successful you have to update the protections on the vmmap regions that were changed. I believe if only part of a mapping is changed you'll have to split that mapping into two.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1009,6 +1009,22 @@ impl Cage {
         ret
     }
 
+    //------------------------------------MPROTECT SYSCALL------------------------------------
+    /*
+    *   mprotect() changes protection for memory pages
```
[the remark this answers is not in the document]
ChinmayShringi: Added

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -274,6 +275,15 @@ pub fn lind_syscall_api(
             interface::mmap_handler(cageid, addr, len, prot, flags, fd, off) as i32
         }
 
+        MPROTECT_SYSCALL => {
+            let addr = arg1 as *mut u8;
```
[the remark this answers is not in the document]
ChinmayShringi: Done

Yaxuan-w:
> So you need to add a function (in mem.rs) before the syscall that checks if the region is in the vmmap, and if not return ENOMEM, also if PROT_EXEC is being attempted return EINVAL like in the mmap_handler function.
> 
> After the syscall, if the host mprotect was successful you have to update the protections on the vmmap regions that were changed. I believe if only part of a mapping is changed you'll have to split that mapping into two.

Will those come with this PR or on another PR?

rennergade:
> > So you need to add a function (in mem.rs) before the syscall that checks if the region is in the vmmap, and if not return ENOMEM, also if PROT_EXEC is being attempted return EINVAL like in the mmap_handler function.
> > After the syscall, if the host mprotect was successful you have to update the protections on the vmmap regions that were changed. I believe if only part of a mapping is changed you'll have to split that mapping into two.
> 
> Will those come with this PR or on another PR?

needs to be with this one

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -262,6 +262,73 @@ pub fn mmap_handler(cageid: u64, addr: *mut u8, len: usize, mut prot: i32, mut f
...
+        libc::mprotect(sysaddr as *mut libc::c_void, rounded_length as usize, prot)
+    };
+
+    if result < 0 {
+        return -1;
+    }
+
+    // Update vmmap entries with new protection
+    let mut vmmap = cage.vmmap.write();
+    // vmmap.update_protections(start_page, npages, prot);
```
rennergade, line 327: Still need to update protections appropriately here

src/RawPOSIX/src/tests/fs_tests.rs
```diff
@@ -724,6 +724,20 @@ pub mod fs_tests {
         lindrustfinalize();
     }
 
+    #[test]
+    pub fn ut_lind_fs_mprotect_unmapped_addr() {
```
rennergade, line 728: good first test, lets add a few more, especially one where it splits an existing region

rennergade (changes requested):
Changes look good so far, need to implement protection updates and add some more tests

src/RawPOSIX/src/interface/mem.rs
ChinmayShringi: Added the `update_protections`

src/RawPOSIX/src/safeposix/vmmap.rs
```diff
@@ -958,4 +971,54 @@ impl VmmapOps for Vmmap {
...
+
+            // Update protection for the overlapping part
+            entry.prot = new_prot;
+        }
+
+        // Insert the split regions
+        for (interval, mut entry) in to_insert {
+            entry.page_num = interval.start();
+            entry.npages = interval.end() - interval.start();
+            let _ = self.entries.insert_overwrite(interval, entry);
```
qianxichen233, line 1021: you are making the splited region has the same entry information including page_num and npages. Should these information be updated as well after you split them?

qianxichen233 (changes requested):

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -262,6 +262,73 @@ pub fn mmap_handler(cageid: u64, addr: *mut u8, len: usize, mut prot: i32, mut f
...
+/// * `prot` - New protection flags (e.g., `PROT_READ`, `PROT_WRITE`)
+///
+/// # Returns
+/// * `i32` - Returns 0 on success, -1 on failure
+pub fn mprotect_handler(cageid: u64, addr: *mut u8, len: usize, prot: i32) -> i32 {
+    let cage = cagetable_getref(cageid);
+
+    // PROT_EXEC is not allowed in WASM
+    if prot & PROT_EXEC > 0 {
+        return syscall_error(Errno::EINVAL, "mprotect", "PROT_EXEC is not allowed");
```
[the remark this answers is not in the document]
JustinCappos: Agree with @qianxichen233 .   I would favor an error message and/or a panic in this case over returning EINVAL.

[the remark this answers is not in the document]
rennergade: I'm in favor of adding some sort of logging. I'm not sure having an easy way for a user to trigger a panic makes sense to me.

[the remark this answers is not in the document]
JustinCappos: The thinking is that it's easier for us when debugging now.   We want to crash early when trying to run code that we don't support.   We'd turn this off later, when opening up to real user code.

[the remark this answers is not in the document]
rennergade: Fine with inserting a panic for now as long as there are comments that are easy to search for when we go to remove this.

As far as logging, if you look at the syscall_error function it actually has functionality to print out logging if verbose mode is enabled. I'm not sure if we actually set up for verbose mode now with wasm but getting that working would be trivial for betting logging.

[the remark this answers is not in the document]
ChinmayShringi: Added logging and added panic

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -262,6 +262,73 @@ pub fn mmap_handler(cageid: u64, addr: *mut u8, len: usize, mut prot: i32, mut f
...
+    }
+
+    // Get system address for the actual mprotect call
+    let sysaddr = vmmap.user_to_sys(addr as u32);
+    
+    drop(vmmap);
+
+    // Perform the actual mprotect syscall
+    let result = unsafe {
+        libc::mprotect(sysaddr as *mut libc::c_void, rounded_length as usize, prot)
```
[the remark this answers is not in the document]
ChinmayShringi: Added

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -274,6 +275,15 @@ pub fn lind_syscall_api(
             interface::mmap_handler(cageid, addr, len, prot, flags, fd, off) as i32
         }
 
+        MPROTECT_SYSCALL => {
+            let cage = interface::cagetable_getref(cageid);
+            let addr = translate_vmmap_addr(&cage, arg1).unwrap() as *mut u8;
```
Yaxuan-w, line 280: As my understanding, all operations in mprotect_handler are based on user space address and we also performs translation in mem.rs line 316. Is there any reason we need to translate here?

rennergade:
This seems ready once @Yaxuan-w 's dispatcher comment is addressed and @qianxichen233 approves the protection changes

src/RawPOSIX/src/safeposix/dispatcher.rs
ChinmayShringi: Removed translate

rennergade (dismissed):
I think this is good now. Approved.

@Yaxuan-w can you see if this fixed your change requests?

Yaxuan-w (changes requested):
Great work!

I met the compilation error with this branch:

```
error[E0308]: mismatched types
   --> crates/rawposix/src/safeposix/dispatcher.rs:284:49
    |
284 |             interface::mprotect_handler(cageid, addr, len, prot)
    |             ---------------------------         ^^^^ expected `*mut u8`, found `u64`
    |             |
    |             arguments to this function are incorrect
    |
    = note: expected raw pointer `*mut u8`
                      found type `u64`
```

My steps were checkout to this branch and `./lindtool.sh cpwasm`. Should I do any other steps?

Yaxuan-w:
@qianxichen233  and I tried to test mprotect test case in our unit test suite, and met compilation error:

```
wasm-ld: error: /tmp/mprotect-66d274.o: undefined symbol: mprotect
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```

We think it would be better to also fix [glibc mprotect](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/sysdeps/unix/sysv/linux/mprotect.c) to support full usage in this pr

rennergade:
@qianxichen233 and @Yaxuan-w can you two review this?

src/glibc/sysdeps/unix/sysv/linux/mprotect.c
```diff
@@ -1,11 +1,19 @@
 #include <unistd.h>
 #include <sysdep-cancel.h>
+#include <stdint.h>
+#include <fcntl.h>
+#include <syscall-template.h>
 
 int
-__GI___mprotect (int fd, const void *buf, size_t nbytes)
```
rennergade, line 5: I'm confused as to why this function originally had this function signature for mprotect? is this even the mprotect syscall?

@qianxichen233 not sure if you can add any insight here

qianxichen233: I just looked into the history in original glibc repo, this entire file seems to be added by Dennis

src/glibc/sysdeps/unix/sysv/linux/mprotect.c
```diff
@@ -1,11 +1,19 @@
...
 {
+  return MAKE_SYSCALL(177, "syscall|mprotect", (uint64_t)(uintptr_t) addr, (uint64_t) len, (uint64_t) prot, NOTUSED, NOTUSED, NOTUSED);
+}
 
-  return 0;
+int
+mprotect (void *addr, size_t len, int prot)
+{
+  return MAKE_SYSCALL(177, "syscall|mprotect", (uint64_t)(uintptr_t) addr, (uint64_t) len, (uint64_t) prot, NOTUSED, NOTUSED, NOTUSED);
 }
```
qianxichen233, line 17: We do not want to have two duplicated syscall entry. Let's only keep the above one and create an alias to `mprotect`

ChinmayShringi: Added alias

Yaxuan-w (approved):

src/RawPOSIX/src/tests/fs_tests.rs
```diff
@@ -724,6 +724,112 @@ pub mod fs_tests {
...
+        assert!(readonlydata >= 0, "mmap should succeed");
+
+        // Change the protection to allow writing
+        // This simulates the C program's mprotect() call to make the page writable
+        let result = cage.mprotect_syscall(
+            readonlydata as *mut u8,
+            PAGESIZE as usize,
+            PROT_READ | PROT_WRITE    // Add write permission
+        );
+        println!("mprotect result: {}", result);
```
qianxichen233, line 755: are we supposed to have these `println` in test?

ChinmayShringi: removed

qianxichen233 (approved):

qianxichen233 (dismissed):
LGTM as long as you resolved the conflict

Yaxuan-w (dismissed):
Reminder to remove those two lines ([line1](https://github.com/Lind-Project/lind-wasm/blob/1747e9f4f05c8e062e7652a8a8201737522821ea/src/sysdefs/src/constants/mod.rs#L5) ; [line2](https://github.com/Lind-Project/lind-wasm/blob/1747e9f4f05c8e062e7652a8a8201737522821ea/src/sysdefs/src/constants/mod.rs#L11)). I fixed them in #117 but it hasn't been merged yet, so that would be great to fix here too. Everything else lgtm

Yaxuan-w (approved):

rennergade (approved):

PR #115, opened 2025-02-21
Add Required Test files at Lind Wasm root

ansalma6:
PR to add the required testfiles to the wasm root at src/RawPOSIX/tmp. The testfiles are taken from lind_project repo without any changes.


Yaxuan-w (commented):
Thanks for adding these! I recalled we have [a preparation script for readlink / readlinkat](https://github.com/Lind-Project/lind-wasm/blob/main/pre-test.sh). That would be helpful if we could add those files too

Yaxuan-w:
> The lind fs root in the script could also be configurable by reading the env variable

I recalled we've discussed this in [a previous pr](https://github.com/Lind-Project/lind-wasm/pull/95). Maybe would be a good idea to integrate them together? (ie: similar to old lindfs shell)

qianxichen233:
> > The lind fs root in the script could also be configurable by reading the env variable
> 
> I recalled we've discussed this in [a previous pr](https://github.com/Lind-Project/lind-wasm/pull/95). Maybe would be a good idea to integrate them together? (ie: similar to old lindfs shell)

Configurable could possible be another topic, but I think we should only load these files into lind fs root when we want to run the tests. These files are only used by test files, so it is not necessary to always have them in lind root. The best way to load these files could be making loading these files be part of the test script, i.e. add some logic in the auto test script that automatically load these files into lind fs root before running the test, and also automatically remove these files after the tests are done

ansalma6:
Updated the PR. The testfiles are added at tests/testfiles repo. We copy the files to Lind Filesystem Root when tests are ran.

LIND_FS_ROOT is taken by os.environ.get("LIND_FS_ROOT", "/home/lind/lind-wasm/src/RawPOSIX/tmp") We Check for the env variable LIND_FS_ROOT, if not found, default to /home/lind/lind-wasm/src/RawPOSIX/tmp. 

There are new arguments --pre-test-only(which only copies the testfiles to lind fs root) and --clean-testfiles(which only removes the testfiles from lind fs root)

rennergade (approved):
This works for me. @Yaxuan-w and @qianxichen233 will use this more so deferring to them for final reviews.

qianxichen233 (approved):

Yaxuan-w (approved):

PR #116, opened 2025-02-22
Implement SYSV shared memory

ChinmayShringi:
Implements shared memory syscalls (shmat, shmdt) in RawPOSIX to match NaCl behavior:
- Added shmat_handler for memory mapping with proper page alignment and hole finding
- Added shmdt_handler for unmapping shared memory segments
- Updated dispatcher for syscall to use handlers

### Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- `cargo build`
- `cargo test`
- Added test cases:
-- `ut_lind_fs_shmat_invalid_args`
-- `ut_lind_fs_shmat_multiple_attaches`
-- `ut_lind_fs_shmat_read_only`

## Checklist:

- [x] I have commented my code, particularly in memory management areas
- [x] I have added tests that prove shared memory operations work correctly
- [x] Implementation matches NaCl behavior for shared memory operations
- [x] Proper error handling for invalid addresses and permissions
- [x] Memory safety checks for page alignment and address space limits


rennergade:
You seem to have added the shmat_handler but didn't implement using it in the syscall?

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -774,12 +775,52 @@ pub fn lind_syscall_api(
 
         SHMAT_SYSCALL => {
             let shmid = arg1 as i32;
-            let cage = interface::cagetable_getref(cageid);
-            // Convert virtual address to physical address
-            let shmaddr = translate_vmmap_addr(&cage, arg2).unwrap() as *mut u8;
+            let shmaddr = arg2 as *mut u8; // No need to translate - shmat_handler will handle address translation
```
rennergade, line 778: Its strange to me that you seem to implement shmat directly in the dispatcher while shmdt from the inner syscall. We should be consistent in coding style here and I'd defer to matching what we did with the mmap syscalls for now.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1468,8 +1473,22 @@ impl Cage {
                 interface::RustHashEntry::Occupied(mut occupied) => {
                     let segment = occupied.get_mut();
 
-                    segment.unmap_shm(shmaddr, self.cageid);
-
+                    // Use shmdt_handler to unmap and remove from vmmap
```
rennergade, line 1476: the comments in shmat are more clear than the ones here, if you could add a bit of context that would be good for shmdt

qianxichen233 (changes requested):
We need to update this to correct its return value for lind-wasm as well:
https://github.com/Lind-Project/lind-wasm/blob/ee056f9a9288240781edfbd935fa7d231890682b/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs#L1460

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -775,11 +776,52 @@ pub fn lind_syscall_api(
         SHMAT_SYSCALL => {
             let shmid = arg1 as i32;
             let cage = interface::cagetable_getref(cageid);
-            // Convert virtual address to physical address
             let shmaddr = translate_vmmap_addr(&cage, arg2).unwrap() as *mut u8;
             let shmflg = arg3 as i32;
-            // Perform shmat operation through cage implementation
-            cage.shmat_syscall(shmid, shmaddr, shmflg)
+            
+            // Get reference to global shared memory metadata
```
rennergade, line 782: I believe that the code before the handler call as well as the code after can be moved into the shmat_handler or into the syscall. We should probably do that to match how the dispatcher works elsewhere (ie not doing critical functionality in the dispatcher)

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1379,14 +1377,31 @@ impl Cage {
...
+     */
     pub fn shmget_syscall(&self, key: i32, size: usize, shmflg: i32) -> i32 {
         if key == IPC_PRIVATE {
             return syscall_error(Errno::ENOENT, "shmget", "IPC_PRIVATE not implemented");
         }
         let shmid: i32;
         let metadata = &SHM_METADATA;
-
+    
+        // Round up size to nearest page size using round_up_page function
```
rennergade, line 1402: this comment probably should say why we do this, which I believe is because mappings need to be aligned to a page boundary (double check this)

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -775,11 +776,52 @@ pub fn lind_syscall_api(
...
+                    
+                    // Set mapping flags - shared memory must use MAP_SHARED
+                    let flags = MAP_SHARED;
+                    
+                    // Call shmat_handler which:
+                    // 1. Handles address translation from user to system space
+                    // 2. Finds appropriate memory region if addr is NULL
+                    // 3. Validates alignment if addr is specified
+                    // 4. Performs the actual mapping
+                    let result = interface::shmat_handler(
```
Yaxuan-w, line 803: we are calling to handler function in `shmat` from dispatcher, but calling to syscall implementation in `shmdt` and calling handler function inside from there. Would be great if we can align usage of handler functions to make future codebase maintenance easier

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -122,6 +122,117 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+
+    // Use mmap with PROT_NONE to effectively unmap the region
+    let result = cage.mmap_syscall(
+        sysaddr as *mut u8,
+        round_up_page(length as u64) as usize,
+        PROT_NONE,
+        (MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED) as i32,
+        -1,
+        0
+    );
```
Yaxuan-w, line 220: I'm a bit confused. Is there specific reason not to perform unmap here? My understanding is that when we use shmat, we call mmap to allocate a region, and we also remove these region from vmmap. If this is the case, would it be more appropriate to use munmap here to prevent memory leaks?

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1481,7 +1543,7 @@ impl Cage {
                         metadata.shmkeyidtable.remove(&key);
                     }
 
-                    return shmid; //NaCl relies on this non-posix behavior of returning the shmid on success
+                    return shmid as i32;
```
qianxichen233, line 1546: We should not return shmid for lind-wasm. From man page: `On success, shmdt() returns 0; on error -1 is returned, and errno is set to indicate the error.`

src/RawPOSIX/src/interface/mem.rs
rennergade: This is how we handle munmap as well, which we inherited from how NaCl does it. You do make a good point that we should explain ourselves here. I think the reasoning is we still want to maintain an unusable mapping here instead of removing it completely and creating a hole, because in the latter case that region could be theoretically reused by mmap when searching for available memory.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
ChinmayShringi: Updated comment

qianxichen233:
What test cases have you tested for SYSV shared memory?

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1481,7 +1564,7 @@ impl Cage {
                         metadata.shmkeyidtable.remove(&key);
                     }
 
-                    return shmid; //NaCl relies on this non-posix behavior of returning the shmid on success
+                    return 1;
```
rennergade, line 1567: Why are we returning 1 here?

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1468,8 +1509,29 @@ impl Cage {
...
+                    
+                    // Update per-cage attachment tracking
+                    // Each cage maintains a count of how many times it has attached this segment
+                    if let Some(mut count) = segment.attached_cages.get_mut(&self.cageid) {
+                        *count -= 1;
+                        if *count == 0 {
+                            // Remove cage entry when no more attachments exist
+                            segment.attached_cages.remove(&self.cageid);
+                        }
+                    }
```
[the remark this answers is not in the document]
ChinmayShringi: used the function here

src/RawPOSIX/src/safeposix/dispatcher.rs
ChinmayShringi: Updated to call syscall

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
ChinmayShringi: Sorry updated to 0 as per man page

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -122,6 +122,117 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+        drop(vmmap);
+    }
+
+    // Use mmap_handler to perform the actual mapping with MAP_FIXED
+    let result = cage.mmap_syscall(
+        useraddr as *mut u8,
+        round_up_page(length as u64) as usize,
+        prot,
+        flags | MAP_FIXED as i32,
+        -1,
```
qianxichen233, line 174: I believe there should be a file backing; otherwise, the memory won't be shared. A simple scenario would be:
```
id = shmget(1234, ...)
fork()
addr = shmat(id, ...)
```
In this case, the memory region returned by shmat should be synchronized between the parent and child. However, the current implementation creates separate shared memory regions within each cage instead.

This is why I’m asking about test cases. Ideally, each new feature PR should specify which test cases it is used for. If it uses an existing test, the PR description should include the test name. If it adds a new test, it should be included as part of the PR.

qianxichen233 (changes requested):

src/RawPOSIX/src/interface/mem.rs
ChinmayShringi: Yes I have added fs_tests for now and I am running and updating code I will be pushing soon

ChinmayShringi: I have added the fs_test they are working I tried this case with a c file and it failed trying to fix it.

rennergade: Okay I was confused as how you implemented this.

We want to use the vmmap and the handlers to find the hole but we should be patching into the functions in shm.rs 

Let's meet and discuss how this should actually look.

ChinmayShringi:
Working on changes after discussion

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -145,6 +149,53 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+/// the shared memory detachment operation. It performs address validation, converts user 
+/// addresses to system addresses, and updates the virtual memory mappings accordingly.
+///
+/// # Arguments
+/// * `cageid` - Identifier of the cage that calls the `shmdt`
+/// * `addr` - Starting address of the shared memory region to detach
+///
+/// # Returns
+/// * `i32` - 0 for success and negative errno for failure
+pub fn shmdt_handler(cageid: u64, addr: *mut u8) -> i32 {
```
rennergade, line 164: nitpick comment but it probably makes sense to have the shmat handler appear before the shmdt handler in this file logisitically

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -305,6 +356,105 @@ pub fn mmap_handler(
...
+        return syscall_error(Errno::EINVAL, "shmat", "address is not aligned") as u32;
+    }
+
+    // Round up the length to a multiple of the page size.
+    let rounded_length = round_up_page(len as u64);
+
+    // Determine the user address.
+    // If addr==0, then we need to allocate space from the vmmap.
+    let mut useraddr = addr as u32;
+    if useraddr == 0 {
```
rennergade, line 411: I believe this isn't completely the correct behavior. If useraddr is 0 then you want to find map space as you do here, but if it is provided you want to call find map space with hint

ChinmayShringi: Updated to call with a hint

rennergade:
@ChinmayShringi this seems to not be passing the CI, have you built and tested yourself?

ChinmayShringi:
Yes, I have run the branch using `cargo test` and `cargo build`. They have suceeded

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -140,11 +144,161 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+    // Check that the provided address is page aligned.
+    let rounded_addr = round_up_page(addr as u64);
+    if rounded_addr != addr as u64 {
+        return syscall_error(Errno::EINVAL, "shmat", "address is not aligned") as u32;
+    }
+
+    // Round up the length to a multiple of the page size.
+    let rounded_length = round_up_page(len as u64);
+
+    // Determine the user address.
```
rennergade, line 200: I think this block between 200-216 is pretty dense. I think we need some more detailed comments and some line separation.

src/RawPOSIX/src/safeposix/shm.rs
```diff
@@ -139,9 +143,18 @@ impl ShmMetadata {
...
+        self.shmtable.get(&shmid).map(|segment| segment.get_shm_length())
+    }
+    
 
     pub fn new_keyid(&self) -> i32 {
         self.nextid
             .fetch_add(1, interface::RustAtomicOrdering::Relaxed)
     }
 }
+
```
rennergade, line 156: I'm not sure if we need this much abstraction just to get the length. Not sure if theres a better refactor here. Curious to know @qianxichen233 thoughts

src/RawPOSIX/src/tests/fs_tests.rs
```diff
@@ -3032,6 +3032,98 @@ pub mod fs_tests {
         lindrustfinalize();
     }
 
+    #[test]
+    pub fn ut_lind_fs_shm_pipe_communication() {
```
rennergade, line 3036: im very confused what this is? this just seems like a normal pipe test. What does this have to do with shared memory?

src/RawPOSIX/src/tests/fs_tests.rs
```diff
@@ -3032,6 +3032,98 @@ pub mod fs_tests {
...
+        assert_eq!(cbuf2str(&read_buf[..bytes_read as usize]), "Hello from child!");
+
+        child.join().unwrap();
+        assert_eq!(cage.close_syscall(pipe_fds.readfd), 0);
+        assert_eq!(cage.exit_syscall(libc::EXIT_SUCCESS), libc::EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
+    #[test]
+    pub fn ut_lind_fs_shm_socket_communication() {
```
rennergade, line 3076: same with this? what does this have to do with shared memory?

rennergade (changes requested):
I initially wrote some of this so would love to have @qianxichen233 review to get some of his thoughts. I'm not sure if there's a better way to refactor this in a sane way.

It does look thorough now in terms of the implementation though which is awesome.

As far as the RawPOSIX tests I'm really not sure what's going on there. We definitely need some relevant and fleshed out tests here.

src/RawPOSIX/src/interface/mem.rs
ChinmayShringi: Updated comments

src/RawPOSIX/src/tests/fs_tests.rs
ChinmayShringi: Removed test case for pipe

src/RawPOSIX/src/tests/fs_tests.rs
ChinmayShringi: Removed test case for socket

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -140,11 +144,161 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+            panic!("shmat did not attach at the expected address");
+        }
+        let mut vmmap = cage.vmmap.write();
+        let backing = MemoryBackingType::SharedMemory(shmid as u64);
+        // Use the effective protection (prot) for both the current and maximum protection.
+        let maxprot = prot;
+        // Add a new vmmap entry for the shared memory segment.
+        // Since shared memory is not file-backed, there are no extra mapping flags
+        // or file offset parameters to consider; thus, we pass 0 for both.
+        let _ = vmmap.add_entry_with_overwrite(
```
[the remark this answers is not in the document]
ChinmayShringi: Added in both handlers

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -140,11 +144,176 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+
+    let shmid = cage.shmdt_syscall(sysaddr as *mut u8);
+    if shmid < 0 {
+        return shmid;
+    }
+
+    // Get a reference to the inner ShmMetadata
+    let length = match get_shm_length(shmid) {
+        Some(l) => l,
+        None => return syscall_error(Errno::EINVAL, "shmat", "invalid shmid") as i32,
```
rennergade, line 305: error should be "shmdt"

ChinmayShringi: fixed

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -140,11 +144,161 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+    };
```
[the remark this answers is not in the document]
ChinmayShringi: Updated error

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1452,6 +1458,8 @@ impl Cage {
             match metadata.shmtable.entry(shmid) {
                 interface::RustHashEntry::Occupied(mut occupied) => {
                     let segment = occupied.get_mut();
+                    // Get the length before unmapping
```
rennergade, line 1461: lets add a bit more to this comment as to why were getting/returning the length

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -140,11 +144,174 @@ pub fn munmap_handler(cageid: u64, addr: *mut u8, len: usize) -> i32 {
...
+
+    // Convert the user address into a system address using the vmmap.
+    let vmmap = cage.vmmap.read();
+    let sysaddr = vmmap.user_to_sys(rounded_addr as u32);
+    drop(vmmap);
+
+    // Call shmdt_syscall which returns:
+    // - positive value: the length of the detached segment on success
+    // - negative value: the errno code on failure
+    let length = cage.shmdt_syscall(sysaddr as *mut u8);
```
rennergade, line 300: also add why we need the length to the comment

rennergade (changes requested):
I think this is good to go once the two comments I added are expanded on.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
ChinmayShringi: Updated comment added more details

src/RawPOSIX/src/interface/mem.rs
ChinmayShringi: reverted the comment  and reoved the value specification:
`Call shmdt_syscall which returns length of the detached segment`

rennergade (approved):
Approved! @qianxichen233 can you look at this now as well?

qianxichen233 (approved):

Yaxuan-w (approved):

Issue #112, opened 2025-02-19
Build Postgres with lind-wasm


qianxichen233:
Progress tracking issue for building postgres with lind-wasm

qianxichen233:
I looked into the build process of postgres to figure out how exactly does postgres is using dynamic loading. It turns out that postgres is trying to build many of its library as a shared object file. They used some options on linker that is not recognized by `wasm-ld` like `-soname`, `--version-script`, `rpath`, etc, causing `wasm-ld` to fail

qianxichen233:
I currently did some simple hacking around on the build process to manually adjust the compile options for each file and I got all the postgres binary successfully built, with all the dynamically linked function as imported function. It would be perferable to find a way to modify the makefile to automatically change the options for linker (remove linker options not supported by wasm-ld, and add the required options for wasm'd dynamic loading)

rennergade:
If those options are stored in a variable like LDFLAGS you should be able to override them with a script wrapping configure/make like we were doing while testing lind-nacl

qianxichen233:
yeah, I am currently thinking that why can't we just make the postgres statistically linked? From my understanding currently, postgres is essentially just compile a list of object into a shared library and then try to compile/run the main binary with these shared library? It seems not hard to me to modify the build process so that all the object file that are supposed in the shared library are bundled as a static archive, and then change the linker option to make main binary load the static archive rather than dynamic library

rennergade:
I didn't delve into this much further than reading these responses in the postgres forum: https://www.postgresql.org/message-id/10384.1639260044%40sss.pgh.pa.us

I know things like libpq are necessary for building the psycopg2 python module.

qianxichen233:
I just tried to statically compiled postgres by manually gather all the objects for dynamic library into a static archive and I am able to build postgres again. And I am able to run the binary as well:
```
wasmtime run --wasi threads=y --wasi preview2=n --allow-precompiled psql --version
```
and I am getting:
```
psql (PostgreSQL) 13.0
```

`libpq` is also something I encountered during the build of postgres, it's initially intended to be built as a shared library (i.e. libpq.so), but I modified its build command and I generated something like `libpq.a` for static linking usage

qianxichen233:
[...]

Issue #62, opened 2024-12-04
Lind Wasm Performance Benchmarks


rennergade:
Issue to track our desired benchmarks for Lind-Wasm.

Benchmark List:

- [ ] Cage startup time benchmark

JustinCappos:
I've updated this.   Please let me know if this isn't clear enough

rennergade:
> I've updated this. Please let me know if this isn't clear enough

This is great for now. Thank you for adding these!

rennergade:
Hopefully we can start building this in the coming weeks.

@m-hemmings Can you look into this to figure out what may be a good infrastructure to use to run benchmarks? We've used [criterion](https://docs.rs/criterion/latest/criterion/struct.Criterion.html) for benchmarking things inside of Rust, but here we need something that can benchmark a full program possibly?

rennergade:
As I posted channel, lmbench may be helpful for some of these.

We can compile individually using our fork here: https://github.com/Lind-Project/lmbench

I know this has benchmarks for context switching, as well as pipes/domain sockets/sockets.

rennergade:
@m-hemmings and @yashaswi2000 are looking into a way to run the above benchmarks as an easy to run tool. The problem is that the above tests are pretty variable in how they'd be setup and their output.

ie.

- Pipe/domain socket/context switching tests are all part of LMBench, which we can compile and run on Lind/Native, comparing their outputs
- Tests benchmarking fork()/exec()/signals/call overhead will probably have to be custom written C files where we add timers that the program will output
- Macrobenchmarks for different applications would most likely be using something like /bin/time unless theres a provided benchmarking output

Each of these would take different steps to compile and different outputs would have to be collated in some way, which means that using some standard benchmarking tool like criterion is probably not going to work.

We can probably build something from scratch that works for different cases, but wanted to make sure that makes sense.

@JustinCappos or @rickmcgeer  do you have any thoughts here?

JustinCappos:
Let's use lmbench when and where we can.

The second line are probably trivial enough we can just code these.   (Or
for the call overhead, using the benchmark I found online which seems more
general is likely fine.)

Macrobenchmarks would ideally use the application's benchmark suite.  If
not, the test suite may work.

Yaxuan-w:
[...]

PR #117, opened 2025-02-22
Fix dup/dup2 + Fix fcntl + New fdtables API

Yaxuan-w:
This PR addresses issues related to `dup`/`dup2` and `fcntl`. Related to bug reported in #112 

### dup/dup2:

Since RawPOSIX introduces the concept of virtual file descriptors through fdtables, duplicating a file descriptor using dup/dup2 in RawPOSIX does not require calling the underlying kernel syscall. Instead, it only needs to ensure that the two virtual fds point to the same underlying kernel fd.

One consideration is that the O_CLOEXEC flag is not automatically inherited when duplicating a file descriptor. Therefore, we need to explicitly set it to false for the duplicated fd.

### fcntl:

Due to the design of fdtables, we need to handle F_DUPFD, F_DUPFD_CLOEXEC, F_GETFD, and F_SETFD separately. The operations F_DUPFD and F_DUPFD_CLOEXEC cannot directly use RawPOSIX dup/dup2 calls because fcntl allocates the lowest available fd starting from the given argument.

To accommodate this behavior, I added a new fdtables API specifically for handling these cases.

### fdtables:

As mentioned above, I introduced a new API in all four algorithms in fdtables to support fcntl operations. I considered to change arguments list of original API, but this will cause modification spread to almost entire RawPOSXI codebase, since get_unused_virtual_fd is used in almost every fs syscalls. We could mark this as a TODO if we want future optimization 

## Test

Passed unit test: 
- tests/unit-tests/file_tests/deterministic/dupwrite.c
- tests/unit-tests/file_tests/deterministic/dup2.c
- tests/unit-tests/file_tests/non-deterministic/dup.c

Added new test file: 
- tests/unit-tests/file_tests/deterministic/fcntl.c


rennergade:
The changes look good and nice comment updates.

What I think would really be helpful here is a longerform comment explaining why we call to the kernel for dup here even though were virtually handling file descriptors. There definitely was a decision/reason for that made at some point but I forget all the details so was struggling to reason through this, seems like an issue that deserves some detail for people looking at this in the future.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -811,31 +811,67 @@ impl Cage {
...
-    *   Then return virtual fd
-    */
-    pub fn dup_syscall(&self, virtual_fd: i32, _start_desc: Option<i32>) -> i32 {
+    /// Reference: https://man7.org/linux/man-pages/man2/dup.2.html
+    /// 
+    /// Duplicates a virtual file descriptor by translating it to a kernel FD though fdtables then 
+    /// calling kernel `dup()`, and mapping the new kernel fd back to a new virtual fd. 
+    /// 
+    /// In Linux, `dup` / `dup2` needs to duplicate the actual kernel file descriptor, ensuring that 
+    /// the new fd references the same underlying kernel object (e.g., a file, socket, or pipe).
```
rennergade, line 820: Are there some specific examples? I think like a file fd's position value is one

rennergade: Hmm, that actually confuses me a bit more. If the dup'ed fds refer to the same descriptor in the kernel as far as offsets etc I'm not sure why we need to replicate them via the kernel dup.

Sorry to be nitpicky on this I just feel like this will be something we'll need to be able to explain clearly for papers etc

Yaxuan-w: If we only handle the dup/dup2 in fdtables, then we also need to handle all the fd related information inside fdtables no matter this fd is duplicated or not. Duplicated fd will share all fd informations, which means any modification to one fd would influence another. For example, we have a fd1 and then duplicate a fd2=fd1 and we do lseek(10) on fd2. Then if user calls fd1 again, the ptr should point to 10. If we only handle the dup/dup2 inside fdtables, we will need to first check if this fd is duplicated or not --> if yes, all changes need to notify underlying kernel + document inside fdtables

JustinCappos: @rennergade is correct,  dup and dup2 are handled within fdtables entirely.  You don't need to make a system call

src/fdtables/docs/get_unused_virtual_fd_from_arg.md
```diff
@@ -0,0 +1,24 @@
+Get a virtualfd mapping to put an item into the fdtable.
+
+This is used to request an unused fd from specific starting position. This is similar to 
+`get_unused_virtual_fd_from_arg` except this function starts from a specific starting position
```
rennergade, line 4: you say its similar to this but isnt this the functions name? I'm confused about the wording here

src/fdtables/src/dashmaparrayglobal.rs
```diff
@@ -141,6 +141,47 @@ pub fn get_unused_virtual_fd(
     Err(threei::Errno::EMFILE as u64)
 }
 
+/// This is used to request an unused fd from specific starting position. This is 
+/// similar to `get_unused_virtual_fd` except this function starts from a specific 
+/// starting position mentioned by `arg` arguments. This will be used for `fcntl`.
+#[doc = include_str!("../docs/get_unused_virtual_fd_from_arg.md")]
+pub fn get_unused_virtual_fd_from_arg(
```
rennergade, line 148: I'm not competely sure about the naming conventions were going with in fdtables, but my gut feeling is that "arg" could be clearer here in what this function does.

Replacing "arg" with something like "startfd" I think would be much easier to quickly understand.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -775,48 +775,72 @@ impl Cage {
...
         }
+        // Get underlying kernel fd
         let wrappedvfd = fdtables::translate_virtual_fd(self.cageid, virtual_fd as u64);
         if wrappedvfd.is_err() {
-            return syscall_error(Errno::EBADF, "dup", "Bad File Descriptor");
+            return syscall_error(Errno::EBADFD, "dup", "Bad File Descriptor");
         }
         let vfd = wrappedvfd.unwrap();
-        let ret_kernelfd = unsafe { libc::dup(vfd.underfd as i32) };
+
```
rennergade, line 800: I think just a one line comment about what were doing here would help out anyone looking at the code.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -775,48 +775,72 @@ impl Cage {
...
 
+        // If the file descriptor newfd was previously open, it is closed before being reused; the
+        // close is performed silently (i.e., any errors during the close are not reported by dup2()).
+        // This step is handled inside `fdtables`
         match fdtables::translate_virtual_fd(self.cageid, old_virtualfd as u64) {
             Ok(old_vfd) => {
-                let new_kernelfd = unsafe { libc::dup(old_vfd.underfd as i32) };
-                // Map new kernel fd with provided kernel fd
-                let _ret_kernelfd = unsafe { libc::dup2(old_vfd.underfd as i32, new_kernelfd) };
+                // The two file descriptors do not share file descriptor flags (the
```
rennergade, line 831: good comment!

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -775,48 +775,72 @@ impl Cage {
...
+                // close-on-exec flag).  The close-on-exec flag (FD_CLOEXEC; see fcntl_syscall())
+                // for the duplicate descriptor is off
                 let _ = fdtables::get_specific_virtual_fd(
```
rennergade, line 834: I think an additional comment here similar to in dup about how the duplication works would be good too.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -845,60 +869,134 @@ impl Cage {
...
-       F_GET_SEALS
-              A bit mask identifying the seals that have been set for
-              the inode referred to by fd.
-
-       All other commands
-              Zero.
-
-       On error, -1 is returned
-    */
+    /// Reference: https://man7.org/linux/man-pages/man2/fcntl.2.html
```
rennergade, line 872: Another really great and thorough comment!

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -845,60 +869,134 @@ impl Cage {
...
+                let cloexec_flag: bool = arg != 0;
+                match fdtables::set_cloexec(self.cageid, virtual_fd as u64, cloexec_flag) {
+                    Ok(_) => return 0,
+                    Err(_e) => return syscall_error(Errno::EBADFD, "fcntl", "Bad File Descriptor"),
                 }
+            }
+            // Return (as the function result) the process ID or process
+            // group ID currently receiving SIGIO and SIGURG signals for
+            // events on file descriptor fd.
+            (F_GETOWN, ..) => 1000,
```
rennergade, line 989: what is 1000? should avoid magic numbers. is this the GID?

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1566,3 +1664,20 @@ pub fn kernel_close(fdentry: fdtables::FDTableEntry, _count: u64) {
         panic!("kernel_close failed with errno: {:?}", errno);
     }
 }
+
+/// This function will be different in new code base (when splitting out type conversion function)
+/// since the conversion from u64 -> i32 in negative number will be different
+///
+/// ## Return
+///
```
rennergade, line 1672: can this have a better comment explaining its use?

rennergade (changes requested):
Looks great. I asked for a few additional comments but think this is in good shape.

src/fdtables/docs/get_unused_virtual_fd_from_arg.md
```diff
@@ -0,0 +1,24 @@
+Get a virtualfd mapping to put an item into the fdtable.
+
+This is used to request an unused fd from specific starting position. This is similar to 
+`get_unused_virtual_fd_from_arg` except this function starts from a specific starting position 
+mentioned by `arg` arguments. This will be used for `fcntl()`.
```
JustinCappos, line 5: What part of this call?   I don't really understand what the different behavior you need is.

Yaxuan-w: [Current `get_unused_fd()` searches from 0](https://github.com/Lind-Project/lind-wasm/blob/main/src/fdtables/src/dashmaparrayglobal.rs#L130), but [`F_DUPFD` and `F_DUPFD_CLOEXEC` of `fcntl()`](https://man7.org/linux/man-pages/man2/fcntl.2.html) are supposed to return the lowest-numbered available fd greater than or equal to the given argument. So I updated the new API to start the search from `arg` to match `fcntl()` behavior.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
Yaxuan-w: Updated with constants from sysdefs

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
Yaxuan-w: Added comments

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
Yaxuan-w: Added comments

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
Yaxuan-w: Added comments

src/fdtables/src/dashmaparrayglobal.rs
Yaxuan-w: Renamed to `get_unused_virtual_fd_from_startfd`

src/fdtables/docs/get_unused_virtual_fd_from_arg.md
Yaxuan-w: Thanks for pointing out! I've updated this doc descriptions. Hope this will explain my thoughts more precisely

rennergade (approved):
This looks good now from my end.

JustinCappos:
The way I implemented fd tracking didn't take fcntl's behavior into account.   I'm fine with what you've done.   

One question is what real world code uses fcntl like this and would break if we did dup2 / grabbed an available fd (but not the lowest) instead?   We're likely to be much, much, much slower than native here.

Yaxuan-w:
> One question is what real world code uses fcntl like this and would break if we did dup2 / grabbed an available fd (but not the lowest) instead? We're likely to be much, much, much slower than native here.

Theoretically, the new API shouldn't introduce additional performance overhead compared to dup2 or grabbing an available fd. The only difference is just offset by the different starting point in the FD search. The rest of the logic remains essentially the same.

Motivated by this, I started doing some benchmarks (#62) to better understand the performance in lind-wasm to help optimize implementation

rennergade:
This looks good to me now but we should be sure as to why this is failing the CI (I know its a WIP for now but still is useful). When it's failing early like this it usually means RawPOSIX isn't building correctly so lets confirm that.

Yaxuan-w:
> When it's failing early like this it usually means RawPOSIX isn't building correctly so lets confirm that.

I can run unit-test on server. I'm going to check about CI.

Yaxuan-w:
> but we should be sure as to why this is failing the CI (I know its a WIP for now but still is useful). When it's failing early like this it usually means RawPOSIX isn't building correctly so let's confirm that.

I talked with @m-hemmings about the failure. This PR actually passed all checks,
image 
but it took over an hour, so it timed out and reported an error. 
image (1)

This should stop happening once we merge #154.

rennergade:
Ok great I appreciate you looking into this.


> @rennergade, I talked with @m-hemmings about the failure. This PR actually passed all checks, ![image 
)
> 
> This should stop happening once we merge #154.

rennergade (approved):
Approved. Should to be good to merge with approval from @JustinCappos

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -845,60 +872,134 @@ impl Cage {
...
-                    return syscall_error(Errno::EBADF, "fcntl", "Bad File Descriptor");
+            // As for `F_DUPFD`, but additionally set the close-on-exec flag
+            // for the duplicate file descriptor.
+            (F_DUPFD_CLOEXEC, arg) => {
+                // Get fdtable entry
+                let vfd = match _fcntl_helper(self.cageid, virtual_fd) {
+                    Ok(entry) => entry,
+                    Err(e) => return syscall_error(e, "fcntl", "Bad File Descriptor"),
+                };
+                // Get lowest-numbered available file descriptor greater than or equal to `arg`
```
JustinCappos, line 946: Explain why (it is because this matches the POSIX system call behavior)

JustinCappos (approved):
Looks good.   One very minor change requested to a comment.   Please fix and then squash and merge.

PR #126, opened 2025-03-07
Format wasmtime and rawposix

Yaxuan-w:
Since I ran `cargo fmt` in the new 3i codebase for formatting adjustments, I also applied `cargo fmt` to the old codebase to eliminate differences caused by formatting inconsistencies. This ensures a clearer and more efficient code review during the upcoming integration. This PR does not modify any actual code, so it can be merged directly. Additionally, the currently waiting PRs (#116 #111 #110 #109) should run `cargo fmt` before merging.


JustinCappos (approved):
thanks for applying these formatting fixes.   We hopefully will get our CI/CD system to check for this in the future.

rennergade (approved):
Looks great. Thanks!

Issue #129, opened 2025-03-12
Fix printf in multi-threading


qianxichen233:
Currently doing `printf` in multi-threaded program will have a small chance to stuck. From I've observed, it will stuck only when two `printf` in different threads gets called at exactly the same time. From syscall logs, we can also observe that the printf gets stucked inside `futex` syscall but never wake up for some reason.

JustinCappos:
Maybe check to see if you can trigger it with only futex?

rennergade:
@qianxichen233 did the TLS PR resolve this? If so lets close it.

qianxichen233:
Not yet. This is still a TODO item that we'll need to address later, but it's probably only a mid-level priority.

Issue #125, opened 2025-03-06
Fix TLS in lind-wasm


qianxichen233:
Open this issue to actually fix TLS feature. Previously we enabled threading to work, but TLS is actually not fixed together.

Some information I've figured out:
1. Previously when I was fixing thread after mmap was integrated, I encountered a crash in `_dl_allocate_tls` under `nptl/allocatestack.c`. At that time, I just commented out this line of code to bypass the issue. But now when I looked into this function, I then believe this is something supposed to fail because this function is responsible for initializating TLS data, it works by parsing the ELF header to retrieve the TLS related information, which is clearly not possible in wasm. So we will probably need to delete these glibc's TLS allocator logic, and pass down the wasmtime to do the actual TLS allocation
2. In wasm, TLS seems to work in a pretty simple way: it has a `__tls_base` global which points to the base address of the TLS, and whenever an access to TLS data occured, it will just read this global and add the offset. So It looks like an easy way to do TLS support is just mmap another TLS region for each thread, and then set the `__tls_base` to the address and then it should work. (Currently since we are not handling TLS at all, all the new threads will have `__tls_base` equal to 0)

rennergade:
Closed with TLS PR

PR #130, opened 2025-03-12
Fix TLS / Fix futex syscall / refactor glibc MAKE_SYSCALL

qianxichen233:
This PR fixed issue [125](https://github.com/Lind-Project/lind-wasm/issues/125) to allow each thread has its own TLS region. The thread stack region is reorganized into following sections:
```
----------- <--- stack top (address returned by `mmap`)
|         |
|         |
|         |
| pthread |
|  stack  |
|         |
|         |
|         |
----------- <--- real user stack bottom
| tls_base|
----------- <--- thread's clone_arg structs
|  clone  |
|   args  |
----------- <--- per-thread TLS data
|   TLS   |
|   data  |
----------- <--- per-thread pthread struct
| pthread |
|  struct |
-----------
```

Besides fixing TLS, `futex_syscall` is both rawposix and glibc is also fixed. On rawposix side, some argument of `futex_syscall` is conditionally address translated based on `futex_op`.  On glibc side, `MAKE_SYSCALL` macro is refactored to make it behavior more closely to glibc's `INTERNAL_SYSCALL` macro. Some glibc function would expect `MAKE_SYSCALL` not handling errno and directly return the raw syscall value (i.e. `-errno`). To be compatible with it, an alternative macro `MAKE_RAW_SYSCALL` is added to return raw syscall value without handling `errno`. Both `MAKE_SYSCALL` and `MAKE_RAW_SYSCALL` macro still calls the same glibc `lind_syscall` function but with different options to specify whether or not to handle `errno`.

A new testcase for TLS has been added. Though it has a small chance to stuck due to a new [issue](https://github.com/Lind-Project/lind-wasm/issues/129).


rennergade:
This all makes sense but it seems weird to me that this is one PR and not three separate PRs.

@JustinCappos and @Yaxuan-w may have more comments on the MAKE_SYSCALL changes since that closely intertwines with 3i I believe.

We most likely can fast-track the TLS and futex changes.

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -975,8 +975,11 @@ pub fn lind_syscall_api(
             // Convert remaining arguments
             let futex_op = arg2 as u32;
             let val = arg3 as u32;
-            let timeout = arg4 as u32;
-            let uaddr2 = arg5 as u32;
+            let timeout = match futex_op as i32 {
+                libc::FUTEX_WAIT => translate_vmmap_addr(&cage, arg4).unwrap() as usize,
```
rennergade, line 979: is this not defined in constants?

src/glibc/nptl/pthread_create.c
```diff
@@ -340,16 +300,51 @@ static int create_thread (struct pthread *pd, const struct pthread_attr *attr,
 
   TLS_DEFINE_INIT_TP (tp, pd);
 
-  unsigned char *stack = 0;
-
-  struct clone_args *args = (void *)pd->stackblock + pd->stackblock_size - sizeof(struct clone_args) - TLS_TCB_SIZE;
+  /*
```
rennergade, line 303: I like the diagram here.

This section of code probably needs a paragraph comment explaining whats going on as well since its pretty dense.

qianxichen233:
> This all makes sense but it seems weird to me that this is one PR and not three separate PRs.
> 
> @JustinCappos and @Yaxuan-w may have more comments on the MAKE_SYSCALL changes since that closely intertwines with 3i I believe.
> 
> We most likely can fast-track the TLS and futex changes.

These three fixes are here together inside this PR since they are all issues related to threading, I just fixed them together when I was fixing TLS related issues.

Regarding to the MAKE_SYSCALL stuff. I think it shouldn't be related to 3i? The MAKE_SYSCALL I am mentioning here is just a macro warpper around glibc's `lind_syscall` function, which is the actual glibc function that does the syscall (and the real `make_syscall` function in 3i if I understand correctly). MAKE_RAW_SYSCALL is just another macro wrapper around `lind_syscall`.

rennergade:
> > This all makes sense but it seems weird to me that this is one PR and not three separate PRs.
> > @JustinCappos and @Yaxuan-w may have more comments on the MAKE_SYSCALL changes since that closely intertwines with 3i I believe.
> > We most likely can fast-track the TLS and futex changes.
> 
> These three fixes are here together inside this PR since they are all issues related to threading, I just fixed them together when I was fixing TLS related issues.
> 
> Regarding to the MAKE_SYSCALL stuff. I think it shouldn't be related to 3i? The MAKE_SYSCALL I am mentioning here is just a macro warpper around glibc's `lind_syscall` function, which is the actual glibc function that does the syscall (and the real `make_syscall` function in 3i if I understand correctly). MAKE_RAW_SYSCALL is just another macro wrapper around `lind_syscall`.

Cool. Yeah lets get some comments from them. This is still a relatively small PR so should be fine.

Yaxuan-w:
> Regarding to the MAKE_SYSCALL stuff. I think it shouldn't be related to 3i? The MAKE_SYSCALL I am mentioning here is just a macro warpper around glibc's lind_syscall function, which is the actual glibc function that does the syscall (and the real make_syscall function in 3i if I understand correctly). MAKE_RAW_SYSCALL is just another macro wrapper around lind_syscall.

The `MAKE_SYSCALL` macro in glibc (which essentially wraps around `lind_syscall`) serves as a trampoline function from glibc --> wasmtime (`lind_syscall` itself contains the entry function of wasmtime). 3i's `make_syscall` handles the transition from wasmtime --> 3i. These two are distinct and operate independently, so these changes should be fine.

src/glibc/nptl/allocatestack.c
```diff
@@ -430,18 +431,20 @@ allocate_stack (const struct pthread_attr *attr, struct pthread **pdp,
 	  /* Don't allow setxid until cloned.  */
 	  pd->setxid_futex = -1;
 
-    // BUG: TLS related stuff is not working in lind-wasm currently
-	  /* Allocate the DTV for this thread.  */
-	  // if (_dl_allocate_tls (TLS_TPADJ (pd)) == NULL)
-	  //   {
-	  //     /* Something went wrong.  */
-	  //     assert (errno == ENOMEM);
+    // lind-wasm: glibc's original TLS initialization depends on ELF header
```
rennergade, line 434: I'm not sure "clearly not possible" is the right comment, this is more of a TODO possibly since this is what we'll be looking at with dylinking

src/glibc/lind_syscall/lind_syscall.h
```diff
@@ -1 +1 @@
-int lind_syscall (unsigned int callnumber, unsigned long long callname, unsigned long long arg1, unsigned long long arg2, unsigned long long arg3, unsigned long long arg4, unsigned long long arg5, unsigned long long arg6);
\ No newline at end of file
+int lind_syscall (unsigned int callnumber, unsigned long long callname, unsigned long long arg1, unsigned long long arg2, unsigned long long arg3, unsigned long long arg4, unsigned long long arg5, unsigned long long arg6, int raw);
```
rennergade, line 1: newline

rennergade:
I think per our discussions wednesday we need to factor our the undefined code here, but otherwise this should be good to go.

qianxichen233:
Encountered a weird issue reported by @Yaxuan-w , fixed together in this PR.
Currently, on the main branch, thread creation fails because mprotect is called with PROT_EXEC for the thread stack. This behavior originates from nptl/allocatestack.c, where glibc checks a flag called _dl_stack_flags. If this flag contains PF_X, glibc attempts to assign PROT_EXEC permissions to the stack region.

The odd part is that this issue did not exist at least a month ago. Threading was working correctly, and _dl_stack_flags did not include PF_X, so glibc wasn’t applying PROT_EXEC to the stack. However, at some point in the past month, something merged into main changed this behavior, causing glibc to attempt executable stack mappings. This is strange because none of the recent PRs appear related to this area.

The fix was pretty simple though, I just removed the PF_X bit from the default stack flags so that stacks are no longer mapped with PROT_EXEC by default. That said, I am not sure if it might be worth investigating what caused the change in glibc behavior or _dl_stack_flags content in the first place.

JustinCappos:
> The fix was pretty simple though, I just removed the PF_X bit from the default stack flags so that stacks are no longer mapped with PROT_EXEC by default. That said, I am not sure if it might be worth investigating what caused the change in glibc behavior or _dl_stack_flags content in the first place.

maybe 'git bisect' will help here?

rennergade:
> Encountered a weird issue reported by @Yaxuan-w , fixed together in this PR.
> 
> Currently, on the main branch, thread creation fails because mprotect is called with PROT_EXEC for the thread stack. This behavior originates from nptl/allocatestack.c, where glibc checks a flag called _dl_stack_flags. If this flag contains PF_X, glibc attempts to assign PROT_EXEC permissions to the stack region.
> 
> The odd part is that this issue did not exist at least a month ago. Threading was working correctly, and _dl_stack_flags did not include PF_X, so glibc wasn’t applying PROT_EXEC to the stack. However, at some point in the past month, something merged into main changed this behavior, causing glibc to attempt executable stack mappings. This is strange because none of the recent PRs appear related to this area.
> 
> The fix was pretty simple though, I just removed the PF_X bit from the default stack flags so that stacks are no longer mapped with PROT_EXEC by default. That said, I am not sure if it might be worth investigating what caused the change in glibc behavior or _dl_stack_flags content in the first place.

I just replied in the issue Alice created but reposting here. This was definitely due to the addition of mprotect and the fail on PROT_EXEC we added, Alice confirmed that the error message that we added for debugging appears.

That being said I think your fix here is the appropriate one, and also this PR looks good to go at this point.

Yaxuan-w (approved):

PR #134, opened 2025-03-24
[Refactor] Add type conversion library

Yaxuan-w:
This PR sets up the basic framework for the typemap library, which will serve as a template for future extensions.

1. Includes only fundamental data type conversion functions (e.g., i32, u32, i64, etc.).
2. Provides minimal helper functions for fdtables and cage (e.g., fd translation and cageid validation).
3. Leaves more complex conversions (e.g., network data structures) to be added later when the corresponding syscalls are implemented, ensuring that type conversions evolve alongside the semantics of those syscalls.

**Note**
This PR focuses on establishing the foundation rather than completeness. Future PRs will incrementally add syscall-specific type conversion functions.


rennergade (commented):
Again really clean and well commented, looks awesome.

My only comment is that it might be worth discussing having the file names be clearer, for example fs_conv isnt immediately clear what the underlying functions are going to do. I'm not super picky about this but may be worth a quick brainstorm.

rennergade:
> Again really clean and well commented, looks awesome.
> 
> My only comment is that it might be worth discussing having the file names be clearer, for example fs_conv isnt immediately clear what the underlying functions are going to do. I'm not super picky about this but may be worth a quick brainstorm.

perhaps just fs_type_conversion etc is more clear here, I dont really see a problem with a bit longer filenames

Yaxuan-w:
> perhaps just fs_type_conversion etc is more clear here, I dont really see a problem with a bit longer filenames

Renamed.

src/typemap/src/fs_type_conversion.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[cfg(feature = "secure")]
+    {
+        if !validate_cageid(arg_cageid, cageid) {
+            return -1;
+        }
+    }
+    // Find corresponding virtual fd instance from `fdtable` subsystem
+    let wrappedvfd = fdtables::translate_virtual_fd(arg_cageid, virtual_fd);
+    if wrappedvfd.is_err() {
+        return -9;
```
rennergade, line 118: this and the -1 above seem like magic numbers, what do they signify? best way is macros but if not then comment

src/typemap/src/syscall_type_conversion.rs
```diff
@@ -0,0 +1,266 @@
...
+///
+/// ## Returns:
+/// Success: A converted u32
+/// Fail: panic
+pub fn get_u32(arg: u64, arg_cageid: u64, cageid: u64) -> u32 {
+    if !validate_cageid(arg_cageid, cageid) {
+        panic!("Invalide Cage ID");
+    }
+
+    if (arg & 0xFFFFFFFF_00000000) != 1 {
```
rennergade, line 135: same

rennergade:
Some things that would help here:

- it would be good to note what "sc" means in the comments for the functions

I think it makes sense to reorganize this so that we have 
- a file for path related helpers "path_conversion.rs"
- a file for data types "datatype_conversion.rs"
- and a file for cage and fd helper functions "cage_helpers.rs"

yzhang71 (approved):
The code looks good to me, and it’s clean and easy to follow.

PR #128, opened 2025-03-12
Enable tests in ci

m-hemmings:
Enables tests to be run automatically when a PR is opened.

The build will pull the branch being requested, build a container using it and run the test suite using the python script included.

The python script will be updated in a future PR to push the html somewhere that can be referenced directly in the return to the PR being raised, but that is not in this.

This PR Focuses on the following:

- Updating Bazel BUILD file for easier reading
- Updates Dockerfile used by cloud build to be used in both cloud and local environments
- Updates cloudbuild.yaml to build the container automatically when a PR is opened, building the branch submitted
- Executes the tests in wasmtestreport.py after building the container to ensure tests are passing that you expect
- Updates how lindtool.sh is finding its home directory for use in the cloud build script

PR #136, opened 2025-03-24
Patch to lindtool

qianxichen233:
A quick patch to the `lindtool`/`lind_config`. Previous [CI PR](https://github.com/Lind-Project/lind-wasm/pull/128) refactored the lindtool but missed the fresh updates in `lindtool` from the signal PR. This causes current `lindtool` unable to compile and run the test file correctly. (since the test file is compiled without signal support, which wasmtime would expect).

Cleaned up some unused code in `lindtool` by the way.

### DIFF
--- a/lind_config.sh
+++ b/lind_config.sh
@@ -5,22 +5,6 @@ rawposix_base="$script_dir/src/RawPOSIX"
 
 CC="${CLANG:=/home/lind/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/clang"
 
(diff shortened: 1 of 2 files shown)


m-hemmings (dismissed):
Consolidation and updates look good, this improves clarity of the tool a great deal.  Thank you.

ansalma6 (approved):

m-hemmings (approved):
Shebang added allows tests to run as before.  Still think the time taken needs to be addressed for test runs, but not in the scope of this.

ansalma6 (approved):

PR #153, opened 2025-04-06
Add signal interruption handling for syscalls

### LINKED ISSUE #118: Blocking Calls for Signals
We'll need to deal with anytime a signal is received by a cage but that cage is in a blocking call. There's various ways we can do this including syscall timeouts.

Some calls I know should block are: read, recv, connect/accept, select/poll, futex...

Let's start out by getting a full list of blocking system calls we currently implement. @ChinmayShringi can you help us generate this.

ChinmayShringi:
## Description

Fixes #118  (Blocking Calls for Signals)

This PR implements signal handling for the `select` syscall. Since we never actually block in the kernel for select/poll operations, we've added direct signal checks before and after the syscall to return EINTR when a signal is pending.

Key changes:
- Added signal checks using `lind_check_no_pending_signal` before and after select operation
- Return EINTR error code when a signal is detected
- Set default timeout to 100ms when no timeout is specified
- Simplified code by removing unnecessary comments

This change is required to support proper signal interruption behavior, particularly for PostgreSQL's pgbench which relies on signal interruption during select syscalls.

### Type of change

## How Has This Been Tested?
- `cargo build`

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/RawPOSIX/src/safeposix/syscalls/net_calls.rs
+++ b/src/RawPOSIX/src/safeposix/syscalls/net_calls.rs
@@ -555,22 +555,6 @@ impl Cage {
         mut errorfds: Option<&mut fd_set>,
         rposix_timeout: Option<RustDuration>,
     ) -> i32 {
-        // Check for pending signals before select
-        if interface::lind_check_no_pending_signal(self.cageid) {
-            return syscall_error(Errno::EINTR, "select", "Interrupted by signal");
-        }
-
-        let mut timeout = if let Some(t) = rposix_timeout {
-            libc::timeval {
-                tv_sec: t.as_secs() as i64,
-                tv_usec: t.subsec_micros() as i64,
-            }
-        } else {
-            libc::timeval {
-                tv_sec: 0,
-                tv_usec: 100_000,
-            }
-        };
 
         let orfds = readfds.as_mut().map(|fds| &mut **fds);
         let owfds = writefds.as_mut().map(|fds| &mut **fds);
@@ -621,11 +605,6 @@ impl Cage {
                 &mut timeout as *mut timeval,
             )
         };
-
-        // Check for pending signals after select
-        if interface::lind_check_no_pending_signal(self.cageid) {
-            return syscall_error(Errno::EINTR, "select", "Interrupted by signal");
-        }
 
         if ret < 0 {
             let errno = get_errno();
--- a/src/RawPOSIX/src/safeposix/syscalls/net_calls.rs
+++ b/src/RawPOSIX/src/safeposix/syscalls/net_calls.rs
@@ -26,6 +26,9 @@ use std::io::{Read, Write};
 use std::mem::size_of;
 use std::sync::Arc;
 use std::{os::fd::RawFd, ptr};
+use std::sync::mpsc;
+use std::thread;
+use std::time::{Duration};
 
 const FDKIND_KERNEL: u32 = 0;
 const FDKIND_IMPIPE: u32 = 1;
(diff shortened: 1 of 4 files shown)


src/RawPOSIX/src/safeposix/syscalls/net_calls.rs
```diff
@@ -532,19 +535,23 @@ impl Cage {
...
-            timeout = libc::timeval {
-                tv_sec: 0,
-                tv_usec: 0,
-            };
-        } else {
-            timeout = libc::timeval {
-                tv_sec: rposix_timeout.unwrap().as_secs() as i64,
-                tv_usec: rposix_timeout.unwrap().subsec_micros() as i64,
-            };
+        // Check for pending signals before select
```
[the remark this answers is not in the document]
ChinmayShringi: I have updated it could you please check if the changes are correct?

qianxichen233:
The approach used to do the syscall interruption here is not correct. For example, in select syscall, you only did one signal check before and after calling libc::select, this essentially does almost nothing since libc::select is still going to be blocked. The correct way to do the syscall interruption is to have a loop warpped over libc::select and we set the libc::select timeout to be like 100ms or so and manually check if timeout reached to the limit, or there is any signal happening here.

I have already finished the implementation of syscall interruption for `select` and `poll` in this [PR](https://github.com/Lind-Project/lind-wasm/pull/164/files#diff-542c44ea3dc0b1c441088134524717dba2fca5965af730942688e8b5b4ab559d), you can probably have a review about how I am doing syscall interruption and then you can apply the same approach to other syscalls

src/RawPOSIX/src/safeposix/syscalls/net_calls.rs
```diff
@@ -135,7 +138,7 @@ impl Cage {
      *   Get the kernel fd with provided virtual fd first
      *   connect() will return 0 when success and -1 when fail
      */
-    pub fn connect_syscall(&self, virtual_fd: i32, addr: &GenSockaddr) -> i32 {
+    pub fn connect_syscall(&self, virtual_fd: i32, addr: &GenSockaddr, timeout_ms: Option<i32>) -> i32 {
```
rennergade, line 141: With all of these we definitely don't want to add a new parameter to the function signature.

We also dont need to write our own timeout mechanisms since they are provided via setsockopt. Check here to see how that's done https://github.com/Lind-Project/safeposix-rust/blob/cd70d39a364bec626433ccf2f5b99788bb68bbdf/src/interface/comm.rs#L236

Basically you can just set this whenever a socket is created and then just loop to check if it times out within the syscalls.

src/RawPOSIX/src/safeposix/syscalls/sys_calls.rs
```diff
@@ -201,7 +201,11 @@ impl Cage {
      *   into the end of its parent's zombie list. Then when parent wants to wait for any of child, it could just check its
      *   zombie list and retrieve the first entry from it (first in, first out).
      */
-    pub fn waitpid_syscall(&self, cageid: i32, status: &mut i32, options: i32) -> i32 {
+    pub fn waitpid_syscall(&self, cageid: i32, status: &mut i32, options: i32, timeout_ms: Option<i32>) -> i32 {
```
rennergade, line 204: I dont think you need a timer here at all? You just need to check if a signal is set within the loop and thats it.

qianxichen233: Yeah, select has a timeout since this is indeed one of its argument. For other syscalls that do not have a timeout, you can just assume its "timeout" is infinity

src/RawPOSIX/src/safeposix/syscalls/sys_calls.rs
```diff
@@ -231,7 +235,16 @@ impl Cage {
...
                     drop(zombies);
                     // TODO: replace busy waiting with more efficient mechanism
-                    interface::lind_yield();
+                    // interface::lind_yield();
+                    // Check for timeout
+                    if interface::readtimer(start_time) > duration {
+                        return syscall_error(Errno::ETIMEDOUT, "waitpid", "Operation timed out");
+                    }
+                    // Check for signals
+                    if interface::signal_check_trigger(self.cageid) {
```
rennergade, line 244: this is the only part you need here.

PR #154, opened 2025-04-07
Testing Cloud build time reduction

ansalma6:


qianxichen233:
The test result does not look correct. Specifically, I saw a bunch of test failing due to `signal_callback export not found` which shouldn't happen. I kind forgot where this export comes from, it could be from `sigaction.c` from glibc, or from `wasm-opt`'s epoch injection

cloudbuild.yaml
```diff
@@ -9,34 +9,133 @@ steps:
...
+        CONTAINER_ID=$(docker ps -qf "name=lind_test_container")
+        echo "Running in container: $$CONTAINER_ID"
 
-        docker cp lind_test_container:/home/lind/lind-wasm/src/RawPOSIX/tmp/results.json /workspace/results.json &&
-        docker cp lind_test_container:/home/lind/lind-wasm/src/RawPOSIX/tmp/report.html /workspace/report.html &&
-        docker rm -fv lind_test_container
+        if [ "$$GLIBC_TAR_CACHE" = "true" ]; then
+          echo "Copying cached glibc"
+          tar xzf /workspace/glibc_build.tar.gz -C /workspace
+          docker cp /workspace/build "$$CONTAINER_ID":/src/glibc
```
m-hemmings, line 125: In your tests, did this improve build time a lot?

wasmtestreport.py
```diff
@@ -632,13 +633,23 @@ def main():
...
+    clean_results = args.clean_results
+
+    if clean_results:
+        if os.path.isfile(output_file):
+            os.remove(output_file)
+        if os.path.isfile(output_html_file):
+            os.remove(output_html_file)
+        print(Path(LIND_FS_ROOT))
+        for file in Path(LIND_FS_ROOT).iterdir():
+            file.unlink()
```
m-hemmings, line 645: If, in the future, this changes from just being files in this location to folders as well, will this work? perhaps each item could be checked to see if it is a file or a folder and removed appropriately depending on what it is?

m-hemmings (commented):
Other than the checking-for-type in wasmtestreport.py everything here looks good, and even that is only a suggestion to protect against future changes that may break the tests

BUILD
```diff
@@ -139,8 +139,8 @@ py_binary(
     data = [
         "tests",         
          "lindtool.sh",
-         "check.log",
-         "check_wasm.log",
+        #  "check.log",
+        #  "check_wasm.log",
```
rennergade, line 143: should this stuff be commented or removed?

rennergade (changes requested):
Minor comment added, but generally I think we'd appreciate a lot more comments explaining what each part is doing. The PR in general looks like a great effort though.

I'll let @m-hemmings and @yashaswi2000 look deeper into the technical side.

cloudbuild.yaml
ansalma6: From my testing, the build time reduces if there are no changes to the glibc files but since this doesnt affect testing, the test time still takes some time

m-hemmings:
@rennergade 
Could you tell me if b372292cb6570ba5e1825121d2b2f5497dd6a7c7 looks clearer now with the comments in cloudbuild.yaml?

@yashaswi2000 is putting changes to the docs to clearly outline the build, but these should help too

rennergade (dismissed):
Comments look great. Approved now on my end.

.devcontainer/Dockerfile
```diff
@@ -6,19 +6,19 @@ ENV DEV_MODE=${DEV_MODE:-false}
 ARG BRANCH_NAME
 ENV BRANCH_NAME=${BRANCH_NAME:-main}
 
+ARG GLIBC_TAR_CACHE
+
 # Install all the required dependencies
-RUN apt-get update && \
-    apt-get install -y build-essential git wget gcc-i686-linux-gnu g++-i686-linux-gnu \
+RUN apt-get -qq update && \
```
yashaswi2000, line 12: great enhancement, cleaner logs!

yashaswi2000 (approved):
LGTM

cloudbuild.yaml
```diff
@@ -9,41 +11,162 @@ steps:
       - 'USERNAME'
       - 'PASSWORD'
 
-  
+  # Step 2: Check if there were changes in `src/glibc` in the current branch.
+  # If so, attempt to find and download a cached `glibc_build.tar.gz` from GCS.
```
yashaswi2000, line 15: typo: if not

m-hemmings: Fixed

PR #157, opened 2025-04-12
[Refactor] Add system-related calls (excluding exec)

celinehoang177:
- Move all functions (excluding exec) in https://github.com/Yaxuan-w/lind-wasm-3i/blob/main/src/rawposix/src/syscalls/sys_calls.rs to main codebase


src/rawposix/src/sys_calls.rs
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+    let cage = get_cage(cageid).unwrap();
+
+    return cage.parent as i32;
+}
+
+/// Those functions are required by wasmtime to create the first cage. `verbosity` indicates whether
+/// detailed error messages will be printed if set
+pub fn lindrustinit(verbosity: isize) {
```
rennergade, line 390: These two function names are pretty bad (and they're my fault!) as relics from the really initial version of rustposix. maybe we change this to rawposix_initialize etc? makes it a bit better.

rennergade: they're also not really "sys system calls" is there a better file for them?

PR #158, opened 2025-04-16
Fix close syscall and incorrect path in gen_root.sh 

Yaxuan-w:
Currently, the main branch returns -errno on syscall failure instead of -1. This behavior is likely caused by two issues:

1. `lind_syscall.o` is not being compiled, which seems to be missing from `lindtool.sh`.
2. `gen_sysroot.sh` contains an incorrect path.

As a result, some test cases fail because they expect and compare the return value to be -1 on failure.

The bug in `close_syscall` is fixed and the path issue in `gen_sysroot.sh` is corrected in this PR. I've talked with `devops` team about updating `lindtool.sh`, so that should be included in another PR.

### DIFF
--- a/src/glibc/gen_sysroot.sh
+++ b/src/glibc/gen_sysroot.sh
@@ -34,7 +34,7 @@ fi
 mkdir -p "$sysroot_dir/include/wasm32-wasi" "$sysroot_dir/lib/wasm32-wasi"
 
 # Pack all found .o files into a single .a archive
-${CLANG:=/home/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/llvm-ar rcs "$output_archive" $object_files
+${CLANG:=/home/lind/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/llvm-ar rcs "$output_archive" $object_files
 "$CLANG/bin/llvm-ar" crs "sysroot/lib/wasm32-wasi/libpthread.a"
 
 # Check if llvm-ar succeeded
--- a/src/sysdefs/src/constants/fs_const.rs
+++ b/src/sysdefs/src/constants/fs_const.rs
@@ -25,6 +25,11 @@ pub const STDERR_FILENO: i32 = 2; // File descriptor for standard error
 // Source: include/dirent.h
 pub const DT_UNKNOWN: u8 = 0;
 
+pub const STARTINGFD: i32 = 0; // Starting file descriptor number
+pub const MAXFD: i32 = 1024; // Maximum number of file descriptors
+pub const STARTINGPIPE: i32 = 0; // Starting pipe descriptor number
+pub const MAXPIPE: i32 = 1024; // Maximum number of pipes
+
 // ===== File Access Permission Flags =====
 pub const F_OK: u32 = 0; // Test for existence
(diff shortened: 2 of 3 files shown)


robinyuan1002 (approved):
LGTM

src/sysdefs/src/constants/fs_const.rs
```diff
@@ -25,6 +25,11 @@ pub const STDERR_FILENO: i32 = 2; // File descriptor for standard error
 // Source: include/dirent.h
 pub const DT_UNKNOWN: u8 = 0;
 
+pub const STARTINGFD: i32 = 0; // Starting file descriptor number
```
rennergade, line 28: We removed these constants due to them being RustPOSIX specific. We definitely don't need to add back the pipe numbers since those will be a grate thing now.

If you need defines for MAXFD or STARTINGFD, my inclination is those should be integrated into FDTables and not these constant files.

src/glibc/gen_sysroot.sh
```diff
@@ -34,7 +34,7 @@ fi
 mkdir -p "$sysroot_dir/include/wasm32-wasi" "$sysroot_dir/lib/wasm32-wasi"
 
 # Pack all found .o files into a single .a archive
-${CLANG:=/home/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/llvm-ar rcs "$output_archive" $object_files
+${CLANG:=/home/lind/lind-wasm/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04}/bin/llvm-ar rcs "$output_archive" $object_files
```
rennergade, line 37: this seems to be a separate idea from the rest of the PR, and it seems you're indicating it needs companion changes to be effective? maybe it makes sense to lump this into a different PR where those other changes are made.

src/sysdefs/src/constants/fs_const.rs
Yaxuan-w: > If you need defines for MAXFD or STARTINGFD, my inclination is those should be integrated into FDTables and not these constant files.

I'm not sure how we want to handle constants. If the eventual goal is to separate all constants as standalone library, then would it make more sense to integrate all constants (fdtables / rawposix / threei / etc.) into one library to avoid re-definition and easy usage?

src/glibc/gen_sysroot.sh
Yaxuan-w: Yeah makes sense! Reverted changes in this PR and changed PR description

src/sysdefs/src/constants/fs_const.rs
rennergade: My opinion is that Linux/POSIX constants that were re-using should be in one place like we have now, but constants that are specific to a certain module (ie fdtables) should be packaged with that module.

qianxichen233:
Since you mentioned lind_syscall.o in the PR, could you also remove this file entirely as part of this change?

I've been always wanted to do this for a long time since `lind_syscall.o` is currently compiled from an very old version of `lind_syscall.c`. Deleting it will help avoid silent failures due to mismatched versions; the build will fail explicitly if the object file isn't recompiled, which is much safer than silently using the wrong binary.

rennergade (approved):
Looks good now, approved.

robinyuan1002 (approved):
LGTM

Issue #49, opened 2024-11-19
Tracking Modifications and Findings During Test Suite Porting -- RawPOSIX


Yaxuan-w:
# Description:

This issue is to document all the changes and discoveries made while porting the test suite from RustPOSIX to RawPOSIX. Below are the current modifications and fixes identified:

### `chmod`

1. Modify behaviors: cannot setting 0o777 so set to 0o755 instead
2. Modify behaviors: The extra bits are special permission bits in native linux

### `close` 

1. Fix bug in fdtable array

### `mmap` 

1. Fix error return 
2. Test fix: Native linux may default to treating the combination as one or the other. 
3. Test fix: Native linux will return EINVAL when offset is negative or not in range. 
4. Test fix: Native linux can mmap character device. 
5. Test fix: Native linux require specific flags to open a dir. 
6. Test fix: Native linux will return ENODEV when mapping a dir.
7. Test fix: The previous test cases created directories but did not remove them. Consequently, when subsequent test cases attempted to create directories with the same names, an error was thrown because the directories already existed. This fix ensures that the directories are properly removed after each test case, preventing these conflicts.
8. Modification: The test now checks for the generic error code (-1) before matching against EINVAL for invalid flag combinations.
9. Test fix: When handling memory-mapped files and syscalls like `getdents_syscall` that deal with directory entries stored in packed structures, accessing fields like `d_off` and `d_reclen` directly can result in alignment issues on architectures that enforce strict memory alignment. To prevent runtime errors, the fields from packed structures are copied into local variables before they are used. This approach avoids accessing misaligned memory directly, ensuring the program remains stable across different platforms. 

### `chdir` 

1. Fix bug that we should first check if desired path exists in native

### `fchdir` 

1. Test fix: Native linux require specific flags to open a dir. 
2. Fix bug that we should also updated the lind cwd (calling libc::cwd and then extract the working directory) after calling fchdir.

### `open` 

1. Fix the condition when opening large number of files
2. Fix the issue related to writing to directories:
   - Test fix: Ensure that when attempting to write to a directory with `O_WRONLY`, it returns `EISDIR` as expected.
   - Test fix: Ensure that opening a directory with `O_RDONLY` returns a valid file descriptor, and attempting to write to it results in `EBADF`.
   - Added cleanup to remove the directory after each test to prevent conflicts with subsequent tests.
3. **Fix for handling `st_nlink` and `st_size` values in directory tests**:
[...]

PR #167, opened 2025-04-23
Fix dup argument failure in RawPOSIX

Yaxuan-w:
I fixed the syscall signature of `dup_syscall` in PR #117 and forgot to update test suite, which cased the compilation failure. Other test failure I've mentioned on Issue #49 and will be fixed in separate PR.


robinyuan1002 (approved):
LGTM

PR #172, opened 2025-04-25
Remove dangling lind-wasm-docs submodule directory

lukpueh:
fixes #139

Other traces of the submodule have already been removed before.


Yaxuan-w:
The CI build failed due to 

```sh
fatal: Remote branch rm-dangling-submodule not found in upstream origin
```

I'm not sure the cause of that. Maybe @m-hemmings could provide some insights about this?

lukpueh:
I doubt it's related to my change. Unfortunately, I can't troubleshoot. When I click on the log I get:

> Permission denied. Review your IAM permissions or contact your admin."

I'm not familiar with Google Cloud builds, but is this even supposed to trigger on every PR from any contributor? Given that it's a paid service, I would assume such a policy can easily get expensive.

rennergade (approved):

rennergade:
Passing CI now that its fixed. Merging.

PR #175, opened 2025-04-26
Clean home folder

ansalma6:
Move all build and test scripts to /scripts folder


lukpueh (commented):
- Did you check, if the renames break anything? (see e.g. [lind_config.sh](https://github.com/Lind-Project/lind-wasm/blob/72917a618ee0b9de28e6d881ca416a15fc3207c5/lind_config.sh#L1) or various docs) Should we just move everything now and fix things later?
- Are all of these scripts still needed? Should we just move them first, and remove them later?
- Is wasmtest.sh not needed anymore? A small note like "Remove wasmtest.sh, because XZY" in a commit message and/or the PR description would be helpful.

lukpueh (changes requested):

ansalma6:
Did you check, if the renames break anything? (see e.g. [lind_config.sh](https://github.com/Lind-Project/lind-wasm/blob/72917a618ee0b9de28e6d881ca416a15fc3207c5/lind_config.sh#L1) or various docs) 
I'm testing it currently, the PR was raised to trigger a build, you got added prematurely for review, currently the cloud build searches for the cloudbuild at the root, the PR was to verify if it can detect cloudbuild.yaml files inside a folder

Are all of these scripts still needed? Should we just move them first, and remove them later?
No, like you said in the following comments, those files will be removed

Is wasmtest.sh not needed anymore? A small note like "Remove wasmtest.sh, because XZY" in a commit message and/or the PR description would be helpful.
Its not needed, it was replaced by the python script

Issue #193, opened 2025-04-30
Fix null status handling in `wait` and `waitpid` dispatcher


Yaxuan-w:
Calling `wait(NULL)` currently causes a panic due to a null address being passed into the [`interface::get_i32_ref(...)`](https://github.com/Lind-Project/lind-wasm/blob/main/src/RawPOSIX/src/interface/types.rs#L218) conversion. This happens during [syscall argument translation in `wait_syscall` and `waitpid_syscall`](https://github.com/Lind-Project/lind-wasm/blob/main/src/RawPOSIX/src/safeposix/dispatcher.rs#L1083), where we unconditionally attempt to get a mutable reference to a status address.

The fix should be simple by replacing code in dispatcher for status with something similar:

```rust
let status = if status_addr == 0 {
    &mut 0
} else {
    interface::get_i32_ref(status_addr).unwrap()
};
```

Yaxuan-w:
No. I think this should be fixed by Dennis's upcoming PR.

Issue #123, opened 2025-03-02
fix nested address translation in dispatcher


qianxichen233:
Some syscalls have address nested inside its struct. For example, `writev` syscall has an `iovec` struct with the definition of
```
pub struct iovec {
    pub iov_base: *mut c_void,
    pub iov_len: usize,
}
```
in which `iov_base` points to a string in guest memory and should also have its address translated. Currently `writev` syscall in dispatcher only translates its high level `iovec` struct address but forgot to translates its inner address. It might be possible that some other syscall has similar issues.

Combined with issue [121](https://github.com/Lind-Project/lind-wasm/issues/121) and [122](https://github.com/Lind-Project/lind-wasm/issues/122), I believe our dispatcher address translation is in a pretty mess state and need a careful review to make sure every syscall has its address translated correctly, no matter it is a nested translation, optional translation, conditional translation, or some other possible scenarios

PR #195, opened 2025-04-30
Fix wait null status handling

Yaxuan-w:
Fix the problem mentioned on issue #193

### DIFF
--- a/src/RawPOSIX/src/safeposix/dispatcher.rs
+++ b/src/RawPOSIX/src/safeposix/dispatcher.rs
@@ -1083,12 +1083,7 @@ pub fn lind_syscall_api(
             let cage = interface::cagetable_getref(cageid);
             // Convert user space buffer address to physical address
             let status_addr = translate_vmmap_addr(&cage, arg1).unwrap() as u64;
-            // Handle NULL explicitly to avoid panic when resolving status pointer from invalid address in `get_i32_ref`
-            let status = if status_addr == 0 {
-                &mut 0
-            } else {
-                interface::get_i32_ref(status_addr).unwrap()
-            };
+            let status = interface::get_i32_ref(status_addr).unwrap();
             cage.wait_syscall(status)
         }
 
@@ -1097,12 +1092,7 @@ pub fn lind_syscall_api(
             let cage = interface::cagetable_getref(cageid);
             // Convert user space buffer address to physical address
             let status_addr = translate_vmmap_addr(&cage, arg2).unwrap();
-            // Handle NULL explicitly to avoid panic when resolving status pointer from invalid address in `get_i32_ref`
-            let status = if status_addr == 0 {
-                &mut 0
-            } else {
-                interface::get_i32_ref(status_addr).unwrap()
-            };
+            let status = interface::get_i32_ref(status_addr).unwrap();
             let options = arg3 as i32;
 
             cage.waitpid_syscall(pid, status, options)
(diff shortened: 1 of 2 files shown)


rennergade (changes requested):
This looks good. I think it warrants a one line comment saying why were doing this.

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1084,7 +1083,12 @@ pub fn lind_syscall_api(
             let cage = interface::cagetable_getref(cageid);
             // Convert user space buffer address to physical address
             let status_addr = translate_vmmap_addr(&cage, arg1).unwrap() as u64;
-            let status = interface::get_i32_ref(status_addr).unwrap();
+            // Handle NULL explicitly to avoid panic when resolving status pointer from invalid address in `get_i32_ref`
```
rennergade, line 1086: So I think this is happening because when you add the base address to NULL (0) its not longer NULL because it has the preceding bits. Can we hint to to that here?

Yaxuan-w: [If we pass a `NULL` (0) to the dispatcher, it will return 0 after the vmmap translation](https://github.com/Lind-Project/lind-wasm/blob/main/src/RawPOSIX/src/interface/mem.rs#L776). 

Consequently, the existing logic passes this translated 0 as `status_addr` to `get_i32_ref`, causing [`get_i32_ref` to interpret it as a pointer to a `NULL` address](https://github.com/Lind-Project/lind-wasm/blob/main/src/RawPOSIX/src/interface/types.rs#L219), resulting in a panic.

rennergade: Oh! Thanks for clarifying. I didn't fully understand this then. I thought translate_vmmap_addr always added the top bits, but in fact it returns 0 if the address is NULL.

Is it worth us having get_i32_ref return &mut 0 instead of having to do an if statement every time we invoke it?

Yaxuan-w: Thanks!! I like that idea — I’ve updated the code in the latest commit to use this approach!

Yaxuan-w:
> Commit https://github.com/Lind-Project/lind-wasm/commit/76d50fb7b423cd63dea8eef827036a141ebc731b: Build Failed

Fixed this error on commit [c08c4c9](https://github.com/Lind-Project/lind-wasm/pull/195/commits/c08c4c9c0700db4e29f72c1f3f28008e05690107)

src/RawPOSIX/src/safeposix/dispatcher.rs
qianxichen233: The approach that makes the most sense to me is to have these type converter functions return an Option—returning None when the address is 0. This would better align with the interface of most syscalls, which also use Option for nullable pointers.

Adopting this approach would introduce a breaking change to the type converter interface, requiring updates in other parts of the dispatcher. However, there is already a high-priority task open to refactor the dispatcher's type conversion logic (see [issue #123](https://github.com/Lind-Project/lind-wasm/issues/123)), including improvements for NULL addresses, nested pointers, and conditional arguments. I'd really hope this issue could be fixed soon as well since our type conversion in dispatcher is indeed in a pretty messy state.

rennergade: libc also presents a null pointer if that makes sense at all: https://docs.rs/libc/latest/libc/constant.PT_NULL.html

rennergade:
Closing this for now per discussion last week, @yzhang71 is currently handling this problem in a wider PR

PR #198, opened 2025-05-01
Add initial memory documentation

rennergade:
Porting this PR over from the deprecated docs repo. I believe this was ready to merge but fell by the wayside.

After we merge this we still need to update this page to include some additional documentation on mprotect() and shared memory.

### DIFF
--- a/docs/internal/memory.md
+++ b/docs/internal/memory.md
@@ -1,0 +1,96 @@
+# Memory Management and Vmmap
+
+## What is a Vmmap?
+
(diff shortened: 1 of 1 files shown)


lukpueh:
see https://github.com/Lind-Project/lind-wasm-docs/pull/22 for original PR and reviews

JustinCappos (dismissed):
Clearer now.   I'm fine how it is.  If you want to consider moving "Why the Vmmap is Necessary" section up, it could also help with clarity.

docs/internal/memory.md
```diff
@@ -0,0 +1,96 @@
+# Memory Management and Vmmap
```
yzhang71, line 1: This looks good to me, but maybe we could make the title more informative, something like "POSIX-Compatible Memory Management in Lind: The Role of Vmmap."

yzhang71 (dismissed):

Issue #209, opened 2025-05-06
CI fails in "Handle Build Status" step


lukpueh:
see e.g. https://github.com/Lind-Project/lind-wasm/pull/207/checks?check_run_id=41711662802

I lack permissions to troubleshoot the issue. When I click on the log, I see the following warning on the Google Cloud dashboard:

> Permission denied. Review your IAM permissions or contact your admin."

lukpueh:
CI still fails for PRs from forks despite fix in #211.

See e.g. build log form https://github.com/Lind-Project/lind-wasm/pull/240

*Step 2: Detect Repo URL*
```
Already have image (with digest): gcr.io/cloud-builders/git
--
2 | Detected REPO_URL:
3 | https://github.com/Lind-Project/lind-wasm.git
```

*Step 3: Build Docker Image*
```
Step 24/30 : RUN if [ "$DEV_MODE" != "true" ]; then       echo "Cloning from $REPO_URL @ $BRANCH_NAME" &&       git clone --branch "$BRANCH_NAME" --single-branch "$REPO_URL" lind-wasm;     fi
1162 | ---> Running in 84c1752ff719
1163 | Cloning from https://github.com/Lind-Project/lind-wasm.git @ docker-build
1164 | Cloning into 'lind-wasm'...
1165 | warning: Could not find remote branch docker-build to clone.
1166 | fatal: Remote branch docker-build not found in upstream origin
1167 | The command '/bin/sh -c if [ "$DEV_MODE" != "true" ]; then       echo "Cloning from $REPO_URL @ $BRANCH_NAME" &&       git clone --branch "$BRANCH_NAME" --single-branch "$REPO_URL" lind-wasm;     fi' returned a non-zero code: 128
1168 |  
```

cc @m-hemmings

m-hemmings:
> Interestingly, it also fails for PRs that don't come from a fork, e.g.: [#239](https://github.com/Lind-Project/lind-wasm/pull/239).
> 
> Unfortunately, I cannot access the [build log](https://console.cloud.google.com/cloud-build/builds;region=global/45b3dcb7-880d-44ab-87af-2f00f8daf225?inv=1&invt=Ab0SvA&project=x-micron-451502-m7), even though I have the necessary permissions and it seems to have been built within the Google Cloud Build log retention period...
[...]

PR #211, opened 2025-05-06
Ci on forks fix

m-hemmings:
## Description

Fixes # [209](https://github.com/Lind-Project/lind-wasm/issues/209)

 

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

1. I forked the repo to my own github, created a new trigger that has the same information in CI
2. created a branch there with the changes in it
3. merged it, so the CI trigger would build.
The following occurred to show that the build was successful:
image
image
image

I expect the CI build here to fail, because the trigger is looking for the cloudbuild.yaml file in lind-wasm, that's why I replicated the testing above.

## Checklist:

### DIFF
--- a/scripts/cloudbuild.yaml
+++ b/scripts/cloudbuild.yaml
@@ -9,7 +9,7 @@ steps:
       - 'docker login --username=$$USERNAME --password=$$PASSWORD'
     secretEnv:
       - 'USERNAME'
-      - 'PASSWORD'
+      - 'PASSWORD'  
 
   # Step 2: Check if there were changes in `src/glibc` in the current branch.
   # If not, attempt to find and download a cached `glibc_build.tar.gz` from GCS.
@@ -40,6 +40,16 @@ steps:
           echo "GLIBC_TAR_CACHE=false" >> /workspace/env.txt
         fi
 
+  - name: 'gcr.io/cloud-builders/git'
+    id: 'Detect Repo URL'
+    entrypoint: 'bash'
+    args:
+      - '-c'
+      - |
+          cd /workspace
+          git config --get remote.origin.url > /workspace/repo_url.txt
+
+
   # Step 3: Build the Docker image from the .devcontainer directory.
   # - Loads variables from /workspace/env.txt set in Step 2.  
   # - Output is logged to /workspace/docker_build.log.
@@ -53,15 +63,18 @@ steps:
       - '-c'
       - |
           source /workspace/env.txt
+          REPO_URL=$(cat /workspace/repo_url.txt)
           (
             docker build \
               --build-arg BRANCH_NAME=$$BRANCH_NAME \
+              --build-arg REPO_URL=$$REPO_URL \
               --build-arg GLIBC_TAR_CACHE=$$GLIBC_TAR_CACHE \
               -t $$USERNAME/lind-wasm:latest ./.devcontainer
             echo $$? > /workspace/docker_build_exit_code
           ) | tee /workspace/docker_build.log
     env:
       - 'BRANCH_NAME=$BRANCH_NAME'
+      - ''
     secretEnv:
       - 'USERNAME'
 
--- a/scripts/cloudbuild.yaml
+++ b/scripts/cloudbuild.yaml
@@ -53,6 +53,8 @@ steps:
       - |
           cd /workspace
           git config --get remote.origin.url > /workspace/repo_url.txt
+          echo "Detected REPO_URL:"
+          cat /workspace/repo_url.txt
 
 
   # Step 4: Build the Docker image from the .devcontainer directory.
@@ -68,6 +70,8 @@ steps:
       - '-c'
       - |
           source /workspace/env.txt
+          echo "Reading REPO_URL from /workspace/repo_url.txt:"
+          cat /workspace/repo_url.txt
           REPO_URL=$(cat /workspace/repo_url.txt)
           (
             docker build \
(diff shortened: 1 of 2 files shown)


rennergade:
Question, do we have protections on "who" can trigger PR/CI builds? I'm realizing that we could be targeted in a possibly expensive manner if that's not setup properly.

m-hemmings:
> Question, do we have protections on "who" can trigger PR/CI builds? I'm realizing that we could be targeted in a possibly expensive manner if that's not setup properly.

@rennergade  There is an option in the CI trigger to require build approval, but it's not turned on right now

rennergade:
I think from our discussions it makes sense to merge this.

@m-hemmings seems to still be failing CI? Is this expected?

rennergade (dismissed):
based on discussion here I'm approving.

Would love @lukpueh to also review.

lukpueh:
I thought we had decided against building PRs from forks?

scripts/cloudbuild.yaml
```diff
@@ -9,7 +9,7 @@ steps:
       - 'docker login --username=$$USERNAME --password=$$PASSWORD'
     secretEnv:
       - 'USERNAME'
-      - 'PASSWORD'
+      - 'PASSWORD'
```
lukpueh, line 12: nit: 
```suggestion
      - 'PASSWORD'
```

scripts/cloudbuild.yaml
```diff
@@ -53,19 +70,23 @@ steps:
...
             docker build \
               --build-arg BRANCH_NAME=$$BRANCH_NAME \
+              --build-arg REPO_URL=$$REPO_URL \
               --build-arg GLIBC_TAR_CACHE=$$GLIBC_TAR_CACHE \
               -t $$USERNAME/lind-wasm:latest ./.devcontainer
             echo $$? > /workspace/docker_build_exit_code
           ) | tee /workspace/docker_build.log
     env:
-      - 'BRANCH_NAME=$BRANCH_NAME'
+      - 'BRANCH_NAME=$BRANCH_NAME'
```
lukpueh, line 85: nit:
```suggestion
      - 'BRANCH_NAME=$BRANCH_NAME'
```

scripts/cloudbuild.yaml
```diff
@@ -40,7 +40,24 @@ steps:
...
+  - name: 'gcr.io/cloud-builders/git'
+    id: 'Detect Repo URL'
+    entrypoint: 'bash'
+    args:
+      - '-c'
+      - |
+          cd /workspace
+          git config --get remote.origin.url > /workspace/repo_url.txt
+          echo "Detected REPO_URL:"
+          cat /workspace/repo_url.txt
```
lukpueh, line 57: This seems overkill. Can't I use something like [`$REPO_FULL_NAME`](https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values) below?

m-hemmings: Cloud Build reports REPO_FULL_NAME as the target, not the source, so it fails the same way as before

lukpueh: Thanks for the insight!

lukpueh (approved):
I only just now understood what this issue is about. 🙈  I thought that GCP couldn't fetch from the fork, but the culprit was actually the Dockerfile, where we cloned from the hardcoded URL. (Would have been lovely to mention this somewhere in the issue or PR description.)

The fix looks good!

 I wonder, though, if we couldn't instead of cloning a second time [`ADD`](https://docs.docker.com/reference/dockerfile/#add) the repo contents into the container?

m-hemmings:
> I only just now understood what this issue is about. 🙈 I thought that GCP couldn't fetch from the fork, but the culprit was actually the Dockerfile, where we cloned from the hardcoded URL. (Would have been lovely to mention this somewhere in the issue or PR description.)
> 
> The fix looks good!
> 
> I wonder, though, if we couldn't instead of cloning a second time [`ADD`](https://docs.docker.com/reference/dockerfile/#add) the repo contents into the container?

This is because the dockerfile is used elsewhere as well, not just in cloud build.  It's designed for users to be able to create a known-state image, which also gets used in the cloud process

lukpueh:
> This is because the dockerfile is used elsewhere as well, not just in cloud build. It's designed for users to be able to create a known-state image, which also gets used in the cloud process. 

Oh, I see. And how's the workflow for creating the image? Where does the user get the Dockerfile from? I'd assume this looks something like: `git clone <repo at known state> && docker build <repo>`. That would be compatible with the other use cases I've seen so far: CI, dev, tests. I think we could make the COPY/ADD scenario the default.

If we do need to support creating the image without repo context, we could maybe do a multi-stage build?

m-hemmings:
> > This is because the dockerfile is used elsewhere as well, not just in cloud build. It's designed for users to be able to create a known-state image, which also gets used in the cloud process.
> 
> Oh, I see. And how's the workflow for creating the image? Where does the user get the Dockerfile from? I'd assume this looks something like: `git clone <repo at known state> && docker build <repo>`. That would be compatible with the other use cases I've seen so far: CI, dev, tests. I think we could make the COPY/ADD scenario the default.
> 
> If we do need to support creating the image without repo context, we could maybe do a multi-stage build?

If we want to break it into a multi stage build or change the CI flow for it, I'm all for it, but I'm not sure it fits in the context of this PR which was to get the CI able to function for forks.  It feels like that should be its own issue/PR

lukpueh:
>If we want to break it into a multi stage build or change the CI flow for it, I'm all for it, but I'm not sure it fits in the context of this PR which was to get the CI able to function for forks. It feels like that should be its own issue/PR

I absolutely agree. Feel free to merge anytime, I already approved. And I can go ahead and submit an issue for the Dockerfile. :)

PR #221, opened 2025-05-12
Use standard rust style guide and rustfmt

### LINKED ISSUE #185: Do we need a custom (rust) style guide?
Style guides are important, but it takes an effort to create and maintain them, and also, for contributors, to adapt, if they already know a similar style guide. 

I think the following strategy generally works well, and doesn't include a custom guide:

- show contributors how to use auto-formatters
- enforce style with linters/auto-formatters in CI
- point to existing style guide  / standard for things that are not covered by linters / auto-formatters

lukpueh:
closes #185 (see issue description for rationale)


rennergade (approved):
This looks good to me.

Yaxuan-w (approved):

PR #227, opened 2025-05-19
Handle NULL buffer address in translate_vmmap_addr and write_syscall

yzhang71:
**Purpose:**
This commit updates the translate_vmmap_addr function to safely handle NULL pointers by returning None when the input address (arg) is 0. The write_syscall logic is updated to properly check for None and return an EINVAL error when the buffer address is NULL, preventing unsafe pointer conversion.

**Issue Reference:**
https://github.com/Lind-Project/lind-wasm/issues/123


src/RawPOSIX/src/interface/mem.rs
```diff
@@ -770,12 +770,12 @@ pub fn check_and_convert_addr_ext(
 ///
 /// # Returns
 /// * `Ok(u64)` - Translated physical memory address. NULL if user address is NULL.
```
Yaxuan-w, line 772: Should we also update comments for returns?

Yaxuan-w (commented):
This fix looks straightforward, and the overall logic makes sense to me. Left a minor suggestion regarding a comment update.

Just to confirm: as we discussed earlier, we're planning to first approve the modification in `translate_vmmap_addr` and this example fix for `write_syscall`, and then follow up with a larger PR (or multiple PRs) covering the remaining syscalls. Is that still the plan?

rennergade:
This seems to be failing build due to RawPOSIX compilation errors.

rennergade:
Fine with this general strategy, but also some syscalls rely on different behavior if the pointer is NULL, and wouldn't just return an error.

ie if readfds in select() is NULL, it doesnt return a syscall error, it just doesn't check readfds etc.

src/RawPOSIX/src/interface/mem.rs
```diff
@@ -770,12 +770,12 @@ pub fn check_and_convert_addr_ext(
...
-pub fn translate_vmmap_addr(cage: &Cage, arg: u64) -> Result<u64, Errno> {
-    // do not convert NULL pointer
+pub fn translate_vmmap_addr(cage: &Cage, arg: u64) -> Option<u64> {
```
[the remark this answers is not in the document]
yzhang71: I think the reason for using `u64` here is that in the `pub fn lind_syscall_api` function in the dispatcher, we define `arg2` as a `u64`, and then pass `arg2` as the input argument to `pub fn translate_vmmap_addr`. So the type of `arg` is `u64` to match.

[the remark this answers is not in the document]
qianxichen233: yes, but `arg` and `address` are for different purpose. The `arg` in `lind_syscall_api` is a general argument that could also possibly be an `u64`, but the function `translate_vmmap_addr` is specifically for translating `address`, a more concrete type of `arg`. It makes more sense to me that `translate_vmmap_addr` only take `u32`, and we do the type conversion (from `arg` to `address`), as we always do, before the handler

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -374,36 +418,51 @@ pub fn lind_syscall_api(
...
-            // 1. Monitor the specified file descriptors for activity
-            // 2. Modify the fd_sets to indicate which descriptors are ready
-            // 3. Return the number of ready descriptors or an error code
+
+            // Optional readfds
+            let mut readfds_raw = match translate_vmmap_addr(&cage, arg2) {
+                Some(addr) => Some(interface::get_fdset(addr).unwrap()),
+                None => None,
+            };
+            let readfds = readfds_raw.unwrap();
```
ChinmayShringi, line 429: This might panic if null

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -994,8 +1175,19 @@ pub fn lind_syscall_api(
             let fd = arg1 as i32;
             let cage = interface::cagetable_getref(cageid);
             // Convert user space buffer addresses to physical addresses
-            let name_addr = translate_vmmap_addr(&cage, arg2).unwrap();
-            let namelen_addr = translate_vmmap_addr(&cage, arg3).unwrap();
+            let name_addr = match translate_vmmap_addr(&cage, arg2) {
+                Some(addr) => addr,
+                None => {
+                    return syscall_error(Errno::EFAULT, "getsockname", "invalid address pointer")
```
ChinmayShringi, line 1181: I am not sure but should this be EINVAL?

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -994,8 +1175,19 @@ pub fn lind_syscall_api(
...
+                }
+            };
+
+            let namelen_addr = match translate_vmmap_addr(&cage, arg3) {
+                Some(addr) => addr,
+                None => {
+                    return syscall_error(Errno::EFAULT, "getsockname", "invalid length pointer")
```
ChinmayShringi, line 1188: same as previous

robinyuan1002 (changes requested):
I appreciate the effort put into this commit. I agree with Chinmay that in some cases, EFAULT should be replaced with EINVAL, specifically when the pointer itself is NULL. Other than this adjustment, everything else looks good to me.

rennergade:
My thought is that Dennis is correct with using efault but I need to look into it further. Will comment back tonight.

src/RawPOSIX/src/safeposix/dispatcher.rs
```diff
@@ -1074,23 +1304,41 @@ pub fn lind_syscall_api(
...
-            let req = translate_vmmap_addr(&cage, arg3).unwrap() as usize;
-            let rem = translate_vmmap_addr(&cage, arg4).unwrap() as usize;
+            let req = match translate_vmmap_addr(&cage, arg3) {
+                Some(addr) => addr as usize,
+                None => return syscall_error(Errno::EFAULT, "nanosleep", "invalid req address"),
+            };
+            let rem = match translate_vmmap_addr(&cage, arg4) {
+                Some(addr) => addr as usize,
+                None => return syscall_error(Errno::EFAULT, "nanosleep", "invalid rem address"),
+            };
```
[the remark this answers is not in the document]
yzhang71: I think req cannot be null, but yes, rem can be null.

req == NULL: ❌ Invalid — required to specify how long to sleep.
rem == NULL: ✅ Optional — if NULL, the remaining time (if interrupted) is not stored.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1707,16 +1710,21 @@ impl Cage {
         clockid: u32,
         flags: i32,
         req: usize,
-        rem: usize,
+        rem: Option<usize>,
     ) -> i32 {
+        let req_ptr = req as *const libc::timespec;
```
rennergade, line 1715: is it possible to use unwrap or zero here too then cast it?

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1688,10 +1688,13 @@ impl Cage {
         uaddr: u64,
         futex_op: u32,
         val: u32,
-        val2: usize,
+        timeout: Option<usize>, // <- updated type
```
qianxichen233, line 1691: Would it make more sense for this to be a PR comment rather than something directly in the code?

Issue #196, opened 2025-04-30
Update build-libc.md


Yaxuan-w:
Since we upgraded clang-16 to clang-18 mentioned on Issue #149 and clang-18 is now available, clang-related info in [build-libc section on our documentation and website](https://github.com/Lind-Project/lind-wasm/blob/main/docs/build/build-libc.md) should also be updated.

Issue #180, opened 2025-01-06
Update bazel.md


rennergade:
Let's describe how builds work more in depth here.

lukpueh:
I am happy to help with this. Can someone provide an overview of the state of the various build tools? Or tell me if below assessment is remotely accurate?


* **bazel** (BUILD, MODULE.bazel, WORKSPACE) is used to build parts of the custom toolchain developed in this repo -- i.e. clang with patched glibc, wasmtime, etc.? -- into a Docker image. This is used for development (.devcontainer/devcontainer.json)  or CI (cloudbuild.yaml), and in some revision hosted on dockerhub.
* Then, in a docker container, **lindtool.sh** can be used as wrapper for the previously built custom toolchain, to compile and run sandboxed programs.
*  **`wasmtestreport.py`** is an additional wrapper over `lindtool.sh`, to compile and run test programs in the sandbox and generate reports
* **`Makefile`** seems like a previous version of `lindtool.sh` and is probably obsolete.
* A few helper scripts/fixtures exist:
  - generate_token_and_upload_secret.py for cloudbuild.yaml
  - skip_test_cases.txt for wasmtestreport.py
  - lind_config.sh for lindtool.sh
  - setup_clang.sh ?? 
  - buildandmerge.yaml ??

lukpueh:
Bazel.md was removed in https://github.com/Lind-Project/lind-wasm/pull/248

I'll open a separate ticket to also remove bazel and close here.

PR #231, opened 2025-05-27
Improve Lind Toolchain docs

lukpueh:
Current documentation about the Lind toolchain used for building and running Lind programs can be found in several docs under: https://github.com/Lind-Project/lind-wasm/tree/997a29c89f1d26e400f7f18eca062914f78c62c9/docs/build

Some issues with the current documentation include:
- lack of bigger picture (how do these tools relate?)
- outdated
- overly detailed (e.g. libc)
- overly succinct (e.g. bazel)

This patch adds a new single document, which gives an **overview** of the full Lind toolchain, and how it is built step by step, with references to Internal Documentation for background information about the toolchain, and build scripts (Docker, bazel, shell scripts) for detailed build commands and options.

Note that I deliberately added the document under the *Usage* section, because I think it could fully replace all documents in the  *Build System* section, and then probably wouldn't need its own top-level section on the website. Moreover, the *Usage* section seems like a good fit given the current experimental state of the project, where usage very likely involves tinkering with the toolchain.

----

**Note to reviewer:**  Please check..
- correctness
- degree of detail (too much? too little?)
- typography (too much switching between bold, italic, code, link markup?)
- links

See [our add docs guide](https://lind-project.github.io/lind-wasm/contribute/add-docs/) for how to build the docs site locally.

related #180 #196

### DIFF
--- a/docs/use/toolchain.md
+++ b/docs/use/toolchain.md
@@ -1,0 +1,81 @@
+# Lind toolchain
+
+The toolchain to build and run lind programs consists of the following components:
+
+- *Clang* with *WebAssembly System Interface* (WASI) support to build *glibc*
+  and lind programs
+- A custom [*glibc*](../internal/libc.md) used as sysroot to build
+  lind programs
+- A custom `wasm-opt` binary to enable [multi-processing](../internal/multiprocess-support.md)
+  in lind programs
+- A custom *WebAssembly* runtime ([`wasmtime`](../internal/wasmtime.md)) based
+  on [*RawPOSIX*](../internal/rawposix.md) to run lind programs
+- Rust Cargo to build `wasmtime`
+
+This document gives an overview of how the toolchain is built. The build process
+is automated with *Docker*, *Bazel* and custom shell scripts. Please refer to
+the relevant files linked below for details about the build commands and used
+options.
+
+
+## Building the toolchain step by step
+
+1. __Install system dependencies__ *(see `apt` in [Dockerfile](https://github.com/Lind-Project/lind-wasm/blob/main/.devcontainer/Dockerfile))*
+
+2. __Download *Clang* and install builtins__ *(see `wget` and `cp` in [Dockerfile](https://github.com/Lind-Project/lind-wasm/blob/main/.devcontainer/Dockerfile))*
+
+    *Clang* supports the *WASI* cross-compilation target out of
+    the box, provided the necessary compiler runtime
+    [builtins](https://clang.llvm.org/docs/Toolchain.html#compiler-rt-llvm) and
+    a matching [sysroot](https://clang.llvm.org/docs/CrossCompilation.html).
+    See [*wasi-sdk* docs](https://github.com/WebAssembly/wasi-sdk) for details.
+
+    A pre-built *Clang* can be downloaded from the
+    [llvm-project releases](https://github.com/llvm/llvm-project/releases/) page.
+    Matching builtins are available in the *lind-wasm* repo under
+    [`src/glibc/wasi`](https://github.com/Lind-Project/lind-wasm/tree/main/src/glibc/wasi).
+
+3. __Build *glibc* and generate sysroot__ (see `make_glibc` in [BUILD](https://github.com/Lind-Project/lind-wasm/blob/main/BUILD) file)
+    1. Configure and compile *glibc* for the *WASI* target with *Clang*.  (see
+      [`wasm-config.sh`](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/wasm-config.sh))
+
+    2. Compile extra files:
+        - `nptl/pthread_create.c`
+        - `lind_syscall/lind_syscall.c`
+        - `csu/wasm32/wasi_thread_start.s`
+        - `csu/wasm32/set_stack_pointer.s`
+
+    3. Generate sysroot
+
+        Combine the built object files into a single archive file and copy
+        along with headers and a pre-built C runtime into a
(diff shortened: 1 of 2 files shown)


docs/use/toolchain.md
```diff
@@ -0,0 +1,81 @@
...
+    See [*wasi-sdk* docs](https://github.com/WebAssembly/wasi-sdk) for details.
+
+    A pre-built *Clang* can be downloaded from the
+    [llvm-project releases](https://github.com/llvm/llvm-project/releases/) page.
+    Matching builtins are available in the *lind-wasm* repo under
+    [`src/glibc/wasi`](https://github.com/Lind-Project/lind-wasm/tree/main/src/glibc/wasi).
+
+3. __Build *glibc* and generate sysroot__ (see `make_glibc` in [BUILD](https://github.com/Lind-Project/lind-wasm/blob/main/BUILD) file)
+    1. Configure and compile *glibc* for the *WASI* target with *Clang*.  (see
+      [`wasm-config.sh`](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/wasm-config.sh))
```
lukpueh, line 40: Let's adds some information about the used build flags here, like the current [build glibc doc](https://github.com/Lind-Project/lind-wasm/blame/997a29c89f1d26e400f7f18eca062914f78c62c9/docs/build/build-libc.md#L132-L145) does, but with better explanation of why these flags are are used with the given values.

Most notably, why do we use ["i686-linux-gnu" as `--host` and `--build` argument](https://github.com/Lind-Project/lind-wasm/blob/997a29c89f1d26e400f7f18eca062914f78c62c9/src/glibc/wasm-config.sh#L8)?

docs/use/toolchain.md
```diff
@@ -0,0 +1,81 @@
...
+
+    2. Compile extra files:
+        - `nptl/pthread_create.c`
+        - `lind_syscall/lind_syscall.c`
+        - `csu/wasm32/wasi_thread_start.s`
+        - `csu/wasm32/set_stack_pointer.s`
```
lukpueh, line 46: It would be great to add some extra information here, like a oneliner description of these files and why they need to be compiled separately?

I think it would also be worth to describe or give an overview of the [used compiler flags](https://github.com/Lind-Project/lind-wasm/blob/997a29c89f1d26e400f7f18eca062914f78c62c9/BUILD#L30-L76). Most notably, what's with all the includes?

lukpueh (commented):
Can someone fill the gaps below?  cc @Yaxuan-w, @rennergade, @qianxichen233 

You can comment here and I add it to the PR. Or we merge first, and edit in new PRs.

docs/use/toolchain.md
```diff
@@ -0,0 +1,81 @@
+# Lind toolchain
+
+The toolchain to build and run lind programs consists of the following components:
+
+- *Clang* with *WebAssembly System Interface* (WASI) support to build *glibc*
+  and lind programs
+- A custom [*glibc*](../internal/libc.md) used as sysroot to build
+  lind programs
+- A custom `wasm-opt` binary to enable [multi-processing](../internal/multiprocess-support.md)
```
m-hemmings, line 9: is the method of customization of either this or the following custom WebAssembly explained anywhere? Is it important to?

lukpueh: Good point! I think the `internal/*` documentation should and in parts already does elaborate on this.

docs/use/toolchain.md
```diff
@@ -0,0 +1,81 @@
...
+  in lind programs
+- A custom *WebAssembly* runtime ([`wasmtime`](../internal/wasmtime.md)) based
+  on [*RawPOSIX*](../internal/rawposix.md) to run lind programs
```
Yaxuan-w, line 12: I'm wondering if it might be more appropriate to say “wasmtime _with_ RawPOSIX” instead of “wasmtime based on RawPOSIX”, since RawPOSIX is implemented as an external library and not something wasmtime is built on top of. It feels more accurate to describe the relationship as composition rather than inheritance.

docs/use/toolchain.md
Yaxuan-w: - `lind_syscall/lind_syscall.c`: This file serves as the interface in glibc to redirect syscalls into Wasmtime. It’s specific to our lind-wasm setup and handles the syscall transition logic.
- `nptl/pthread_create.c`, `csu/wasm32/wasi_thread_start.s`, `csu/wasm32/set_stack_pointer.s`: These three are related to our thread support in lind-wasm. Some of the changes we made for enabling thread functionality involve these files.

I think @qianxichen233 could provide more detailed context here

docs/use/toolchain.md
m-hemmings: Maybe we can link to that in a footnote or something on this and expand on it in the appropriate place

rennergade: I think there's a bit more to this to do with signaling, hopefully @qianxichen233 can clarify.

docs/use/toolchain.md
rennergade: @qianxichen233

docs/use/toolchain.md
lukpueh: > What do you think about moving the link to wasm-opt and create an issue for mentioning it in the mutli-processing doc?

Just changed the link, will post an issue shortly.

lukpueh:
Thanks for your comments. This is ready for another round of review

Yaxuan-w (approved):

rennergade (approved):

PR #202, opened 2025-05-05
Clang 18

yzhang71:
## Description

Fixes # (issue)

 

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

- Test A - `lind_project/tests/test_cases/test_a.c`
- Test B - `lind_project/tests/test_cases/test_b.c`

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been added to a pull request and/or merged in other modules (native-client, lind-glibc, safeposix-rust)

PR #240, opened 2025-06-13
Add experimental Dockerfile for end-to-end testing  (WIP)

lukpueh:
The new Dockerfile is based on `.devcontainer/Dockerfile`, and simplified to allow optimzation for layer caching. Most notably, it
- introduces multi-stage builds, and
- copies sources from the build context, when they are used to invalidate caches only when relevant files are updated

The goal is to run the Docker build in a GitHub Action (e.g. https://github.com/docker/build-push-action), which requires strategic caching to not exceed allowed build minutes and cache sizes.

**Caveat**
This is an experimental/explorative patch and **does not fit in** with the rest of the toolchain. It:
* ignores `bazel` for a lighter toolchain and to focus on one caching mechanism
* moves all glibc-related build instructions into `gen_sysroot.sh` to reduce complexity

Depending on how well this works, we may replace or revert back to existing tools.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,13 @@
+name: End-to-end testing
+
+on:
+  push:
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build e2e
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
+        with:
+          platforms: linux/amd64
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,85 @@
+# syntax=docker/dockerfile:1.7-labs
+# (use non-stable syntax for convenient --parents option in COPY command)
+
+# Download clang
+# NOTE: chmod is required to extract as user below
+FROM scratch AS clang
+ADD --chmod=644 https://github.com/llvm/llvm-project/releases/download/llvmorg-16.0.4/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04.tar.xz /clang.tar.xz
+
+FROM ubuntu:latest
+
+#####################
+# SYSTEM DEPENDENCIES
+#####################
+# TODO: only install required deps, and only in the stage, where needed
+# TODO: https://docs.docker.com/build/building/best-practices/#apt-get
+RUN apt-get update && \
+    apt-get install -y -qq \
+        apt-transport-https \
+        bash \
+        binaryen \
+        bison \
+        build-essential \
+        curl \
+        g++ \
+        g++-i686-linux-gnu \
+        gawk \
+        gcc \
+        gcc-i686-linux-gnu \
+        git \
+        gnupg \
+        golang \
+        libssl-dev \
+        libxml2 \
+        openssl \
+        python3 \
+        sudo \
+        unzip \
+        vim \
+        wget \
+        zip
+
+#####################
+# USER SETUP
+#####################
+# TODO: do not require user and abspaths (in build/test scripts)
+RUN usermod --login lind --move-home --home /home/lind ubuntu && \
+    groupmod --new-name lind ubuntu
+RUN echo "lind ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+USER lind
+RUN mkdir /home/lind/lind-wasm
+WORKDIR /home/lind/lind-wasm
+
+
+###################
+# USER DEPENDENCIES
+###################
+# Install pinned rust nightly version (known to work)
+# TODO: Figure out why newer versions break the build and unpin
+# TODO: Beware of RUN layer caching: cache not invalidated by remote change
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
+    sh -s -- -y --default-toolchain nightly-2025-06-01
+ENV PATH="/home/lind/.cargo/bin:${PATH}"
+
+# Extract Clang
+# see best practices for downloading and extracting large files
+# https://docs.docker.com/build/building/best-practices/#add-or-copy
+RUN --mount=from=clang,target=/clang tar xf /clang/clang.tar.xz
+
+###################
+# Build GLIBC
+###################
+COPY --chown=lind:lind src/glibc src/glibc
+RUN ./src/glibc/gen_sysroot.sh
+
+###################
+# Build WASMTIME
+###################
+COPY --chown=lind:lind --parents src/wasmtime src/RawPOSIX src/fdtables src/sysdefs .
+RUN cargo build --manifest-path src/wasmtime/Cargo.toml
+
+###################
+# Run TESTS
+###################
+COPY --chown=lind:lind --parents scripts tests tools .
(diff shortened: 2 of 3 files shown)


lukpueh:
https://github.com/Lind-Project/lind-wasm/pull/240/commits/ff6aaf4f882a1e0d7b378aeda26fffa4297d1721 adds a GHA workflow to build the dockerfile on GitHub.  [Resulting .dockerbuild archives](https://github.com/lukpueh/lind-wasm/actions/runs/15614173396#summary-43982545210) can be [imported and inspected with Docker Desktop](https://docs.docker.com/build/ci/github-actions/build-summary/). 

With cold caches the build takes ~1h.  Surprisingly, the glibc build including sysroot generation takes only ~5 minutes, but running wasmtestreport.py takes more than 50 minutes. Locally on my mac, each takes around 20 minutes.

This means that caching (clang, glibc, sysroot) won't help us much, if the tests are the bottleneck. Will look into improving test times ...

lukpueh:
8255a7c48776e79ef67369621bc4f9de8df8775a adds a substantial speed gain only by compiling `wasmtime` with `--release` flag.

`wasmtestreport.py` run time just dropped from 50m to 10m. 🚀

lukpueh:
https://github.com/Lind-Project/lind-wasm/pull/240/commits/2876d61456918414fcbb40c831d16cf503a435d2 enables caching. This works well out of the box, see e.g. build times for
- [no build-relevant change](https://github.com/lukpueh/lind-wasm/actions/runs/15727196668): **29s**
- ["unit test" change](https://github.com/lukpueh/lind-wasm/actions/runs/15727224261): **5m 25s** 

The cache layer sizes are still massive (~4GB total; quota is 10GB), so we should definitely look into trimming those. But maybe we can do that in a follow-up PR. What do others think? cc @Lind-Project/lind-team 

(NOTE: Before I mark his for review, I need to refactor gen_sysroot.sh so that it does not break existing tooling)

lukpueh:
This PR is ready for review!
- see updated [PR description](https://github.com/Lind-Project/lind-wasm/pull/240#issue-3142852937), and
- passing [CI](https://github.com/Lind-Project/lind-wasm/actions/runs/15871731200/job/44749692510?pr=240) (relevant build logs can be found in the "Build e2e" step).

 cc @Yaxuan-w, @rennergade, @m-hemmings

scripts/make_glibc_and_sysroot.sh
```diff
@@ -0,0 +1,150 @@
...
+../configure \
+  --disable-werror \
+  --disable-hidden-plt \
+  --disable-profile \
+  --with-headers=/usr/i686-linux-gnu/include \
+  --prefix=$GLIBC/target \
+  --host=i686-linux-gnu \
+  --build=i686-linux-gnu \
+  CFLAGS=" -matomics -mbulk-memory -O2 -g" \
+  CC="$CC --target=wasm32-unkown-wasi -v -Wno-int-conversion"
```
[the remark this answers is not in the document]
lukpueh: This is just copy-pasted from the original build (configure) shell script:
https://github.com/Lind-Project/lind-wasm/blob/ae2f3871fe1a8e7c5e839f1a2bc4f2bbf8fb8434/src/glibc/wasm-config.sh#L10

It would be good to find out why the glibc build passes despite the typo (related conversation: https://github.com/Lind-Project/lind-wasm/pull/202#discussion_r2121892005). If this option is ignored than the proper fix would be to remove the flag.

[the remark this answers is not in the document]
Yaxuan-w: > If this option is ignored than the proper fix would be to remove the flag.

This flag is necessary because it explicitly informs Clang that the target binary is in WebAssembly format and supports the WASI ABI.

> It would be good to find out why the glibc build passes despite the typo

Clang appears to be tolerant in parsing the target triple. It likely matches on the key components: the architecture (`wasm32`) and the environment (`wasi`), and ignores or defaults the middle vendor component (`unknown`).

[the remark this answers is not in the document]
lukpueh: Okay, that makes sense. I'll commit you suggestion then. Thanks!

scripts/Dockerfile.e2e
```diff
@@ -0,0 +1,76 @@
+# syntax=docker/dockerfile:1.7-labs
+# (use non-stable syntax for convenient --parents option in COPY command)
+
+# Download clang
+FROM scratch AS clang
+ADD https://github.com/llvm/llvm-project/releases/download/llvmorg-16.0.4/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04.tar.xz /clang.tar.xz
+
+FROM ubuntu:latest
```
m-hemmings, line 8: Is using ubuntu:latest a good idea? It feels like this could be brittle and could be better done by pinning a specific build here like ubuntu:22.04 or ubuntu:24.04

scripts/Dockerfile.e2e
```diff
@@ -0,0 +1,76 @@
...
+
+#####################
+# DEPENDENCIES
+#####################
+# Install misc build dependencies
+# TODO: only install required hard requirements, see best practices
+# https://docs.docker.com/build/building/best-practices/#apt-get
+RUN apt-get update && \
+    apt-get install -y -qq \
```
m-hemmings, line 17: Is there a reason to not do a --no-install-recommends here?

scripts/Dockerfile.e2e
```diff
@@ -0,0 +1,76 @@
...
+        golang \
+        libssl-dev \
+        libxml2 \
+        openssl \
+        python3 \
+        sudo \
+        unzip \
+        vim \
+        wget \
+        zip
```
m-hemmings, line 40: We could save a little bit of room here if we clean up after the updates with `rm -rf /var/lib/apt/lists/*`

scripts/Dockerfile.e2e
lukpueh: I have a branch with `--no-install-recommends` and with quite a few deps removed: https://github.com/lukpueh/lind-wasm/blob/273d8ba44f274ae3f5c633f3b7c0849c60ca6d5f/Dockerfile#L13-L32

But I didn't have time yet to confirm that it doesn't break anything.

scripts/make_glibc_and_sysroot.sh
```diff
@@ -0,0 +1,150 @@
+#!/bin/bash
```
rennergade, line 1: if this file replaces other scripts functionality should those scripts be removed?

rennergade (commented):
I really love everything you've done here, thank you very much for your efforts.

I still don't really understand the details of how the caching works and would like to, plus it may be better to document this all even more fully now. Would it be possible to also add something in docs/ going into some detail?

scripts/make_glibc_and_sysroot.sh
lukpueh: This needs some planning. `lindtool.sh` depends on those scripts, and we depend on `lindtool.sh` via `wasmtestreport.py`. I will make a proposal to untangle these local build tools.

lukpueh:
> Would it be possible to also add something in docs/ going into some detail?

Absolutely! Let me add a ticket.

PR #258, opened 2025-06-30
ci: fix e2e cache re-use between branches

### LINKED ISSUE #256: GHA cache not shared between PRs
Looks like the speed-up through caching promised in #240 does not quite work (yet!), at least not across PRs. I just submitted  two PRs (#254, #255), both of which should have completely re-used the caches from #240, because none of the relevant sources changed, but they didn't use any cache at all.

The problem seems to be that caches are [not shared across sibling-branches](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache).
The solution, I think, is to prime caches on the main branch.

lukpueh:
Prior to this PR e2e builds were only triggered und thus cached on PR. This allows re-using caches for new pushes to a PR, but not for new PRs from sibling branches. This is because GitHub only shares caches between branches, if one branch is based off another.

As a fix, we also trigger and cache on push to main, so that PRs based on main can re-use that cache.

fixes #256

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -2,6 +2,9 @@ name: End-to-end testing
 
 on:
   pull_request:
+  push:
+    branches:
+      - main
 
 permissions: {}
 


rennergade (approved):
Great!

PR #270, opened 2025-07-03
ci: misc optimizations in Dockerfile.e2e

### LINKED ISSUE #251: build: optimize Docker caching
#240 adds a build pipeline to run lind-wasm end-to-end tests locally and on GitHub Actions in a Docker container, making use of Docker [build caching](https://docs.docker.com/build/cache/).

A full build takes almost half of the cache space allotted by GitHub (4 of 10GB), but there is some room for optimization:

* install  only what (see [POC](https://github.com/lukpueh/lind-wasm/blob/273d8ba44f274ae3f5c633f3b7c0849c60ca6d5f/Dockerfile#L13-L31)) and where (see multi-stage builds below) it is needed 
* adopt multi-stage builds  for independent steps (1) [build glibc+gen sysroot](https://github.com/Lind-Project/lind-wasm/blob/85d77114f5729e60df11883d697f613a1fbf9b3c/scripts/Dockerfile.e2e#L51-L55), (2) [build wasmtime](https://github.com/Lind-Project/lind-wasm/blob/85d77114f5729e60df11883d697f613a1fbf9b3c/scripts/Dockerfile.e2e#L57-L65), and (3) [final image ](https://github.com/Lind-Project/lind-wasm/blob/85d77114f5729e60df11883d697f613a1fbf9b3c/scripts/Dockerfile.e2e#L68-L76). Benefits are:
   * parallel builds, might improve build time
   * more targeted build environment (e.g. we only need to install rust for compiling wasmtime, but not for building, or running tests)
   * fewer cache invalidation (e.g. wasmtime is currenlty rebuilt, when glibc is rebuilt, because caches in non multi-stage builds are incremental, but that is not necessary and not the case with multi-stage builds)
   * *BONUS: multi-stage builds also allow us to create a smaller final image, which only contains what is needed (clang + sysroot and wasmtime binaries), and discards everything else (sources, build dependencies, etc.). This does not decrease cache sizes, because pevious layers are still cached, even when left out of the final image, but it is good practice and a good idea, if we want to publish the final image, e.g. on Dockerhub*

* review extent of [clang](https://github.com/Lind-Project/lind-wasm/blob/85d77114f5729e60df11883d697f613a1fbf9b3c/scripts/Dockerfile.e2e#L6) and [wasmtime](https://github.com/Lind-Project/lind-wasm/blob/85d77114f5729e60df11883d697f613a1fbf9b3c/scripts/Dockerfile.e2e#L64) installations (do we need all of their features?)
* consider splitting up `make_glibc_and_sysroot.sh` into multiple `COPY`+`RUN` layers for more fine-grained caching. This only makes sense, if files used by later instructions are more likely to change, than those used earlier in the script, e.g. if we expect files used by [glibc's `configure && make`](https://github.com/Lind-Project/lind-wasm/blob/85d77114f5729e60df11883d697f613a1fbf9b3c/scripts/make_glibc_and_sysroot.sh#L81-L92) to change less frequently, than the [extra build files](https://github.com/Lind-Project/lind-wasm/blob/85d77114f5729e60df11883d697f613a1fbf9b3c/scripts/make_glibc_and_sysroot.sh#L96-L115)

lukpueh:
Optimize Dockerfile.e2e for faster builds and smaller caches:
- Groom system deps installed with apt
- Uses multi-stage build for targeted cache invalidation: Independent build stages 'build-wasmtime' and 'build-glibc' do not invalidate each other's caches
- Mount build artifacts for use in later stages instead of copying them.

Caveat: The Dockerfile is meant to run the e2e test pipeline on `docker build` (locally or in CI). It does not produce a usable image. The resulting image, does not include all build artifacts for caching optimization reasons mentioned above.

closes #251

### DIFF
--- a/scripts/Dockerfile.e2e
+++ b/scripts/Dockerfile.e2e
@@ -1,76 +1,79 @@
 # syntax=docker/dockerfile:1.7-labs
-# (use non-stable syntax for convenient --parents option in COPY command)
+# NOTE: use non-stable syntax for convenient --parents option in COPY command
 
-# Download clang
-FROM scratch AS clang
-ADD https://github.com/llvm/llvm-project/releases/download/llvmorg-16.0.4/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04.tar.xz /clang.tar.xz
+# Dockerfile for lind-wasm end-to-end testing
+#
+# - Install build dependencies
+# - Build wasmtime, glibc and sysroot for clang cross-compilation
+# - Run tests using wasmtestreport.py
+#
+# Usage:
+#   docker build --platform=linux/amd64 -f scripts/Dockerfile.e2e .
+#
+# Caveat:
+#   This Dockerfile is meant for end-to-end testing via `docker build`.
+#   It employs several optimizations for targeted cache invalidation
+#   and minimal cache size:
+#   - scoped COPYs
+#   - independent build stages
+#   - minimal number of layers
+#   - minimal layer sizes
+#   As part of this not all build artifacts are copied into the final image,
+#   and thus not available on `docker run`.
+#   TLDR: The resulting docker image is not meant for distribution.
 
-FROM ubuntu:latest
-
-#####################
-# DEPENDENCIES
-#####################
-# Install misc build dependencies
-# TODO: only install required hard requirements, see best practices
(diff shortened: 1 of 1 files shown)


Yaxuan-w (approved):
Thanks for doing this! Comments are really clear and detailed explaining reason of each flag.

Issue #269, opened 2025-07-02
Deprecate lindtool.sh


rennergade:
It would be great to get rid of this swiss army knife script that's confusing. My suggestion would be to move everything with system builds to a top level Makefile.

As far as its compilation options it would be great to separate this out and alias so we essentially have 'lind-clang' which we can call with options.

The option to run files can also just be alliased as "lind-wasm"

rennergade:
I think the first step here is migrating all of the system build steps to a Makefile, that way we can remove the bazel commands and proceed with the clang PR.

@m-hemmings can you take a look at this

lukpueh:
@m-hemmings, I have a PR in the queue, which I want to submit later today. I'd appreciate your feedback!

m-hemmings:
> [@m-hemmings](https://github.com/m-hemmings), I have a PR in the queue, which I want to submit later today. I'd appreciate your feedback!

I approved it this morning, I think it's waiting on a second review

PR #272, opened 2025-07-07
clean some of the commands, updated the path of lindtool.sh

xhanghe:
Clarified docs/use/run-programs.md and added compilation first. Also modified some paths from the personal path to the general path. Modified the lindtool path to scripts/lindtool.sh

### DIFF
--- a/docs/use/getting-started.md
+++ b/docs/use/getting-started.md
@@ -35,6 +35,8 @@ EOF
 **3. Compile and run**
 
 The use `lindtool.sh` to compile and run your program in the Lind Sandbox.
+Run cpwasm command first to compile the environment before compiling the
+example program. 
 
 ```bash
 ./scripts/lindtool.sh cpwasm
--- a/docs/use/getting-started.md
+++ b/docs/use/getting-started.md
@@ -37,8 +37,9 @@ EOF
 The use `lindtool.sh` to compile and run your program in the Lind Sandbox.
 
 ```bash
-./lindtool.sh compile_test hello
-./lindtool.sh run hello
+./scripts/lindtool.sh cpwasm
+./scripts/lindtool.sh compile_test hello
+./scripts/lindtool.sh run hello
 ```
 
 *Here is what happens under the hood:*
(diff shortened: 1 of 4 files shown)


lukpueh (commented):
Thanks! We are actually trying to deprecate lindtool.sh, so I don't think it's worth fixing related docs at the moment (see #269).

Yaxuan-w:
> We are actually trying to deprecate lindtool.sh, so I don't think it's worth fixing related docs at the moment (see https://github.com/Lind-Project/lind-wasm/issues/269).

I'm also considering this.. We have lots of new people join recently, so I'm thinking what's the better solution here

docs/use/getting-started.md
```diff
@@ -37,8 +37,9 @@ EOF
 The use `lindtool.sh` to compile and run your program in the Lind Sandbox.
 
 ```bash
-./lindtool.sh compile_test hello
-./lindtool.sh run hello
+./scripts/lindtool.sh cpwasm
```
Yaxuan-w, line 40: Should we mention somewhere on the `getting-started.md` that people needs to compile the runtime environment before their first run?

docs/use/run-programs.md
```diff
@@ -1,8 +1,19 @@
...
 Run the `.wasm` file, modify the wasmtime path to your own
 
+Here is an example to run the `printf.c` with wasmtime
+
 ```
-/home/lind-wasm/wasmtime/target/debug/wasmtime add.wasm
+cd $HOME/lind-wasm
+./scripts/lindtool.sh cpwasm
+./scripts/lindtool.sh cptest tests/unit-tests/file_tests/deterministic/printf
+src/wasmtime/target/debug/wasmtime run --wasi threads=y --wasi preview2=n \
```
Yaxuan-w, line 15: I think this command is equivalent to `./scripts/lindtool.sh run <test>`?

Yaxuan-w (changes requested):
Awesome thanks for updating! Few suggestions

rennergade (approved):
I'm fine with whatever changes @Yaxuan-w approves for the short term here since this is going to be completely overhauled in a week.

Yaxuan-w (approved):

PR #273, opened 2025-07-07
Fix: close_syscall panics on negative fd

stupendoussuperpowers:
`close_syscall` panics when a negative file descriptor is passed. This occurs because there is no bound check for fd's being negative. A common case where this occurs is when `open(...)` fails. 

This PR adds a simple check for this. 

```
int fd = open("/path/doesnt/exist", O_RDONLY);    // fd = -1
...
close(fd);    // POSIX behaviour: return -1, errno set to EBADF. 
```

Current lind behaviour:
```
thread 'main' panicked at crates/rawposix/src/fdtables/dashmapvecglobal.rs:379:15:
index out of bounds: the len is 1024 but the index is 18446744073709551614
```

### DIFF
--- a/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
+++ b/src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
@@ -905,6 +905,8 @@ impl Cage {
     /// ## Returns:
     ///     Return 0 when success; -1 along with errno when fail.
     pub fn close_syscall(&self, virtual_fd: i32) -> i32 {
+        // We only check for negative fd's here since fdtables uses u64.
+        // Upper-bound checks are handled by the close_virtualfd function.
         if virtual_fd < 0 {
             return syscall_error(Errno::EBADF, "close", "Bad File Descriptor");
         }


src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
```diff
@@ -905,6 +905,10 @@ impl Cage {
     /// ## Returns:
     ///     Return 0 when success; -1 along with errno when fail.
     pub fn close_syscall(&self, virtual_fd: i32) -> i32 {
+        if virtual_fd < 0 {
```
Yaxuan-w, line 908: I think checking here is because u64 in fdtables won't have negative values and we've added check on fdtables for virtual_fd not exceeding the max_fd. That'd be great to have a one line comment here explaining why we only checks the negative value here and the reason we perform checks here not on fdtables.

Yaxuan-w (changes requested):
Thanks for fixing this! Only one minor suggestion.

src/RawPOSIX/src/safeposix/syscalls/fs_calls.rs
Yaxuan-w: Also curious, does the same problem exist on other syscalls?

stupendoussuperpowers: I'll add the comment. Not sure if there's an easy way to check other syscalls since this was very `close` specific issue as the panic only happens because of an indexing error on the fdtables. I can go over the rest to see if similar issues exist for other syscalls.

rennergade (approved):
Looks good. Thank you!

Issue #230, opened 2025-05-22
Do not build rawposix independently


lukpueh:
[lindtool make_all](https://github.com/Lind-Project/lind-wasm/blob/997a29c89f1d26e400f7f18eca062914f78c62c9/scripts/lindtool.sh#L285) builds rawposix and wasmtime separately. But the wasmtime build also includes rawposix via its Cargo workspace.

*Proposed fix*
- Remove `//:make_rawposix` target in `lindtool make_all`
- Remove build targets to build rawposix independently from lindtool (`compile_rawposix|cpraw`) and bazel BUILD (make_rawposix). In this context of lind-wasm, raw posix should always be built as part of wasmtime.

PR #276, opened 2025-07-09
build: replacement for lindtool.sh

lukpueh:
Add **shell scripts** and **Makefile** to replace lindtool.sh (see #269) and use in wasmtestreport.py and Dockerfile.e2e:

* Replacements
  - `lindtool.sh compile_test` → `scripts/lind_compile`
  - `... run` → `scripts/lind_run`
  - `... compile_wasmtime` → `make wasmtime`
  - `... compile_src` + `... make_glibc` → `make sysroot`
* Does not add replacement for `... rawposix` because of [#230](https://github.com/Lind-Project/lind-wasm/issues/230)
* Replacement for `... debug` may be added in follow-up PR, if needed
* Adds new `make test` target, and slightly restructures Dockerfile.e2e
* Please review commit by commit and see commit messages for details!
* IMPORTANT NOTE: default usage of wasmtestreport.py is backwards incompatible (see  commit message 88da8e17f31695a44f8f8451464a76528ee6fa90)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ sysroot:
 
 .PHONY: wasmtime
 wasmtime:
+	# Build wasmtime with `--release` flag for faster runtime (e.g. for tests)
 	cargo build --manifest-path src/wasmtime/Cargo.toml --release
 
 .PHONY: test
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+
+.PHONY: all
+all:
+	@echo "Run targets individually!"
+
+.PHONY: sysroot
+sysroot:
+	./scripts/make_glibc_and_sysroot.sh
+
+.PHONY: wasmtime
+wasmtime:
+	cargo build --manifest-path src/wasmtime/Cargo.toml --release
+
+.PHONY: test
+test:
+	# NOTE: `grep` workaround required for lack of meaningful exit code in wasmtestreport.py
+	LIND_WASM_BASE=. LIND_FS_ROOT=src/RawPOSIX/tmp \
+	./scripts/wasmtestreport.py && \
+	cat results.json && \
+	! grep '"number_of_failures": [^0]' results.json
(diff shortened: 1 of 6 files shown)


scripts/Dockerfile.e2e
```diff
@@ -43,43 +43,46 @@ RUN apt-get update && \
...
-# Install clang, build glibc and generate sysroot
+# Build glibc and generate sysroot
 FROM base AS build-glibc
-# NOTE: We `curl | tar` in spite of Docker best practices to save cache layers
-RUN curl -sL https://github.com/llvm/llvm-project/releases/download/llvmorg-16.0.4/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04.tar.xz | \
-       tar -xvJ
-COPY --parents scripts/make_glibc_and_sysroot.sh src/glibc .
-RUN ./scripts/make_glibc_and_sysroot.sh
+COPY --parents scripts/make_glibc_and_sysroot.sh src/glibc Makefile .
+RUN make sysroot
```
lukpueh, line 70: Maybe using (and therefor COPYing) Makefile here (and in the build-wasm stage) is not such a good idea. Now, potentially unrelated changes in the Makefile will invalidate the cache layer and re-trigger the costly build. If this turns out to be a problem, we might want to revert to calling the build commands directly here. Should we add an issue, or code comment?

Makefile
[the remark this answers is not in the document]
lukpueh: Good idea. I won't add one in this PR, because it's a bit tricky. Each target generates quite a few build artifacts in different places. I suggest you create a separate PR on top of this one.

scripts/lind_run
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
rennergade, line 1: can we name this "lind-wasm" or at least alias it to that? probably both?

scripts/lind_compile
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
rennergade, line 1: can we name this to "lind-clang" or alias? both?

Makefile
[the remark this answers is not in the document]
m-hemmings: Sounds good, resolving.

m-hemmings (approved):

scripts/lind_compile
lukpueh: I chose a generic name, because the script runs `clang`, `wasm-opt` and `wasmtime compile`. Wouldn't `lind-clang` be misleading?

Would you mind adding aliases in follow-up PRs?

rennergade: Yeah thats fine with me.

rennergade (dismissed):

m-hemmings (approved):

rennergade (approved):

PR #278, opened 2025-07-09
build: improve local development experience

### LINKED ISSUE #253: Upgrade local unit test experience and docs
Right now its pretty confusing how to run unit tests locally.

It was packaged in with Bazel I believe but we've just removed that. I'm not sure if we're shifting back to some Makefiles? A top-level makefile with something like "make test" directing to the python suite would be easy for most to understand. We should mention how to do this in the README and the docs as well.

Also it could be nice to have some sort of in terminal HTML viewer installed to read the output since the HTML version is much easier to parse.

lukpueh:
Update Dockerfile.e2e to improve local development experience. That is, to support the following usage scenario:

- build base image
- run container with local source code repository mounted into container
  - this allows syncing source code changes between host and container
- run `make` on the container sheel to build toolchain (libc/sysroot, wasmtime) from local sources and test with wasmtestreport.py

The PR also adds code comments and updates contributor guidelines to document the usage.

See commit messages for details. 

(Note that the changes in `make_glibc_and_sysroot.sh` seem unrelated but are necessary to use the script on arbitrary paths, e.g. when mounted on `/lind` as suggested in the updated docs.

fixes #253

### DIFF
--- a/docs/contribute/unit-tests.md
+++ b/docs/contribute/unit-tests.md
@@ -21,30 +21,23 @@ cd lind-wasm
 ```
 3. Build Docker Image 
 ```
-docker build -t testing_image -f .devcontainer/Dockerfile --build-arg DEV_MODE=true --platform=linux/amd64 .
+docker build --platform=linux/amd64 -f scripts/Dockerfile.e2e -t dev --target base .
 ```
 4. Run the image 
 ```
-docker run -it testing_image /bin/bash
+# NOTE: this mounts the source code repo into the container
+docker run --platform=linux/amd64 -v $(PWD):/lind -w /lind -it dev /bin/bash
 ```
 5. Build toolchain (glibc and wasmtime)
 ```
-bazel build //:make_glibc //:make_wasmtime
+# this may take a while ...
+make wasmtime sysroot
 ```
 6. Run the test suite 
 ```
-bazel run //:python_tests
+./scripts/wasmtestreport.py
 ```
-(This will run the whole test suite.  Use `scripts/wasmtestreport.py --help` to
-list available arguments and flags)
-Note: Pass test suite arguments using
-```
-bazel run //:python_tests -- <wasmtestreport arguments>
-```
-For example: 
-```
-bazel run //:python_tests -- --timeout 10
-```
+Run `scripts/wasmtestreport.py --help` to list available usage options.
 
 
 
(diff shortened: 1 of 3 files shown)


scripts/make_glibc_and_sysroot.sh
```diff
@@ -12,13 +12,16 @@
...
+# - expects `clang` and other llvm binaries on $PATH
+# - expects GLIBC source in $PWD/src/glibc
+#
 set -x
 
-# TODO: Fix absolute paths (consider using $PWD, if necessary)
-GLIBC="/src/glibc"
-CLANG="/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04"
+CC="clang"
+GLIBC="$PWD/src/glibc"
```
m-hemmings, line 23: will `$PWD` correctly return the path necessary if it's accessed from outside the scripts folder?
I had thought something like:
```
# Set source path for script location and lind base for reference as one folder up from it
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
LIND_BASE="$(cd "${SCRIPT_DIR}/.." && pwd)"

# Set things as referenced by the base install of LIND
GLIBC="${LIND_BASE}/src/glibc"
CLANG="${LIND_BASE}/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04"
```

That way it works reliably regardless of where it’s called from. This can help avoid surprises during CI runs or automation too.  What do you think?

docs/contribute/unit-tests.md
```diff
@@ -21,30 +21,23 @@ cd lind-wasm
 ```
 3. Build Docker Image 
 ```
-docker build -t testing_image -f .devcontainer/Dockerfile --build-arg DEV_MODE=true --platform=linux/amd64 .
+docker build --platform=linux/amd64 -f scripts/Dockerfile.e2e -t dev --target base .
 ```
 4. Run the image 
 ```
-docker run -it testing_image /bin/bash
+# NOTE: this mounts the source code repo into the container
```
m-hemmings, line 28: Could we expand this a tiny bit to note that it mounts all the contents of the current folder including the source code repo into the container?

m-hemmings (changes requested):

PR #294, opened 2025-07-15
build: add lint/format/docs targets to Makefile (Fixes #280)

### LINKED ISSUE #280: build: add make targets for lint, format, docs, ...
#276 added a new basic project-level Makefile. Let's use it for other targets, e.g.:
- [ ] lint with [cargo fmt](https://github.com/Lind-Project/lind-wasm/blob/642dae071e44e0451c270ee8919eeff205b7d9da/.github/workflows/lint.yml#L20) and [clippy](https://github.com/Lind-Project/lind-wasm/blob/642dae071e44e0451c270ee8919eeff205b7d9da/.github/workflows/lint.yml#L35-L42)
- [ ] format (like `cargo fmt` above, but without `--check`)
- [ ] [publish docs](https://github.com/Lind-Project/lind-wasm/blob/642dae071e44e0451c270ee8919eeff205b7d9da/.github/workflows/docs.yml#L37)
- [ ]  [serve docs locally](https://github.com/Lind-Project/lind-wasm/blame/642dae071e44e0451c270ee8919eeff205b7d9da/docs/contribute/add-docs.md#L22)

J7-7-7:
Closes #280, adds lint/format/docs targets to Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,17 +46,14 @@ docs-serve:
 
 .PHONY: clean
 clean:
-	@echo "Removing glibc artifacts"
+	@echo "glibc artifacts"
 	$(RM) -r src/glibc/build src/glibc/sysroot src/glibc/target
-
 	@echo "cargo clean (wasmtime)"
 	cargo clean --manifest-path src/wasmtime/Cargo.toml
 
-	@echo "Deleting test outputs"
+.PHONY: distclean
+distclean: clean
+	@echo "removing test outputs & temp files"
 	$(RM) -f results.json report.html
 	$(RM) -r src/RawPOSIX/tmp/testfiles || true
-
-	@echo "Purging compiled test artifacts"
 	find tests -type f \( -name '*.wasm' -o -name '*.cwasm' -o -name '*.o' \) -delete
-
-	@echo "Clean done."
--- a/Makefile
+++ b/Makefile
@@ -35,10 +35,7 @@ lint:
 .PHONY: format
 format:
 	cargo fmt --all --manifest-path src/wasmtime/Cargo.toml
-
-.PHONY: docs-publish
-docs-publish:
-	mkdocs gh-deploy --force 
+ 
 
 .PHONY: docs-serve
 docs-serve:
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,27 @@ test:
 	./scripts/wasmtestreport.py && \
 	cat results.json && \
 	! grep '"number_of_failures": [^0]' results.json
+
+.PHONY: lint
+lint:
+	cargo fmt --check --all --manifest-path src/wasmtime/Cargo.toml
+	cargo clippy \
+	    --manifest-path src/wasmtime/Cargo.toml \
+	    --all-features \
+	    --keep-going \
+	    -- \
+	    -A warnings \
+	    -A clippy::not_unsafe_ptr_arg_deref \
+	    -A clippy::absurd_extreme_comparisons
+
+.PHONY: format
+format:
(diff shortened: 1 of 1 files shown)


rennergade:
I'm confused at why other changes besides the Makefile are in this PR (ie docs stuff)

J7-7-7:
Didn't notice it, sorry about that. Corrected it.

Makefile
```diff
@@ -19,3 +19,27 @@ test:
...
+	    -A clippy::not_unsafe_ptr_arg_deref \
+	    -A clippy::absurd_extreme_comparisons
+
+.PHONY: format
+format:
+	cargo fmt --all --manifest-path src/wasmtime/Cargo.toml
+
+.PHONY: docs-publish
+docs-publish:
+	mkdocs gh-deploy --force
```
m-hemmings, line 41: This already auto-updates when docs are merged into main, do we need this here? It seems dangerous

m-hemmings (changes requested):

Yaxuan-w:
I think we can also add compilation of wasmtime and lind-glibc for `make all` target for now.

rennergade:
> I think we can also add compilation of wasmtime and lind-glibc for `make all` target for now. Do we want to include this fix in this PR or in a separate PR?

I think my comment here may clear that up: https://github.com/Lind-Project/lind-wasm/issues/280#issuecomment-3079929244

I think its worth adding a build target here

rennergade:
It also probably makes sense to wrap this issue into this PR: https://github.com/Lind-Project/lind-wasm/issues/296

J7-7-7:
Added "make build" and "make clean".  Should we add "make all" to "all: build lint test"?

Makefile
```diff
@@ -19,3 +23,41 @@ test:
...
+	cargo fmt --all --manifest-path src/wasmtime/Cargo.toml
+ 
+
+.PHONY: docs-serve
+docs-serve:
+	mkdocs serve
+
+.PHONY: clean
+clean:
+	@echo "▶ Removing glibc artefacts"
```
rennergade, line 50: *artifacts (spelling)

rennergade:
> Added "make build" and "make clean". Should we add "make all" to "all: build lint test"?

According to the Make manual apparently "all" should mean build the system its pointing to, so I think Lukas's questioning was actually not right in that case. (https://stackoverflow.com/a/22735335)

So from looking at that, I think Alice is right for requesting "all" to be the same as build (can point to it), but doesn't call the other stuff (lint, docs etc)

rennergade (dismissed):
This looks great to me now!

Makefile
```diff
@@ -19,3 +22,41 @@ test:
...
+
+	@echo "cargo clean (wasmtime)"
+	cargo clean --manifest-path src/wasmtime/Cargo.toml
+
+	@echo "Deleting test outputs"
+	$(RM) -f results.json report.html
+	$(RM) -r src/RawPOSIX/tmp/testfiles || true
+
+	@echo "Purging compiled test artifacts"
+	find tests -type f \( -name '*.wasm' -o -name '*.cwasm' -o -name '*.o' \) -delete
```
Yaxuan-w, line 60: Most of time developers maybe don't want to cleanup test files since they are testing behaviors for runtime. Could we split the entire `clean` target into 3 sub-target? 

1. `make clean-test`: only cleanup test files
2. `make clean`: remove glibc & wasmtime
3. `make distclean`: remove all

Thanks!

Yaxuan-w (commented):
Thanks for doing this! This helps a lot. A minor suggestions to make life easier during development.

PR #308, opened 2025-07-23
Add glibc 3i

Yaxuan-w:


stupendoussuperpowers:
We need to modify the make_glibc_and_sysroot scripts so that `cp_data_between_cages.h` and `register_handler.h` headers are added to sysroot includes.

src/glibc/lind_syscall/register_handler.c
```diff
@@ -0,0 +1,6 @@
+#include <stdint.h> // For uint64_t definition
+#include <syscall-template.h> // For make_syscall definition
+
+int register_handler(uint64_t targetcage, uint64_t targetcallnum, uint64_t handlefunc_index_in_this_grate, uint64_t this_grate_id) {
```
stupendoussuperpowers, line 4: Do we still need the `handlefunc_index_in_this_grate` argument? From my understanding syscall dispatch now happens by looking for an exported function with the `<syscall name>_grate` name.

Yaxuan-w: Good point!! I'll resolve this suggestion after our first round testing :D

rennergade (approved):
Added one minor comment but this also looks good.

rennergade:
This is good to merge from my end whenever.

src/glibc/lind_syscall/lind_syscall.c
```diff
@@ -54,3 +55,77 @@ int lind_syscall (unsigned int callnumber, unsigned long long callname, unsigned
...
+    // is returning an 32-bit address, which may overflow the int type (i32)
+    // luckily we can handle this easily because the return value of mmap is always
+    // multiple of pages (typically 4096) even when overflow, therefore we can distinguish
+    // the errno and mmap result by simply checking if the return value is
+    // within the valid errno range
+    if(ret < 0 && ret > -256)
+    {
+        errno = -ret;
+        return -1;
+    }
```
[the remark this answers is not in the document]
Yaxuan-w: Thanks for pointing out! Updated

Issue #291, opened 2025-07-11
test: revise wasmtestreport.py


lukpueh:
A few observations from working with wasmtestreport.py:

- Improve output and debugging options. It should be easy to see the stdout/stderr from the tools you're running (gcc, lind_compile, lind_run). The JSON/HTML is nice, but seems like an additional feature. Just printing to the log would go a long way.
- Get rid of absolute paths. It shouldn't only work on a computer, which has a `/home/lind` user directory.
- Go easy on the `chdir`s. It's prone to inconsistent state (see [discussion](https://github.com/Lind-Project/lind-wasm/pull/285#issuecomment-3061496871)).
- Consider using a temporary directory for all test artifacts and remove that instead of individual files. You can also add an option to keep it for troubleshooting.
- Revise test setup and tear down. There's a lot of test file copying. Do we always need all of them? Also, cleaning up all files created during tests afterwards should be the default (see temp dir above).
- Allow running individual test files.
- Revise programming style. The code has a lot of nesting and repitition, which could be streamlined quite a bit. E.g. look at the similarly shaped try/except blocks, repetitive invocations of add_test_result, duplication in deterministic vs. non-deterministic test functions, etc.

rennergade:
@DyanB PR's have finished all of this. Thanks!

PR #315, opened 2025-07-30
wasmtestreport.py Test suite revision

DyanB:
Reference issue: [Link](https://github.com/Lind-Project/lind-wasm/issues/291)

This PR takes care of some of the issues listed in Issue 291
 - Removed chdir calls by using a wrapper for subprocess.run
 - Added Logging for the subprocess calls
 - Can run multiple individual test files now


J7-7-7 (commented):
Looks good to me, I don't see any major issues

Karan-05 (dismissed):
Looks good to me

Suggestion 
We could use logger.debug instead of print ?? as logging adds timestamps, severity labels etc

scripts/wasmtestreport.py
```diff
@@ -622,6 +611,58 @@ def compare_test_results(file1, file2):
...
+        if DEBUG_MODE:
+            print(f"[{label}] TIMEOUT after {timeout}s")
+        raise
+    except Exception as e:
+        if DEBUG_MODE:
+            print(f"[{label}] EXCEPTION: {str(e)}")
+        raise
+
 def main():
     os.chdir(LIND_WASM_BASE)
```
m-hemmings, line 667: I like the general move away from `os.chdir`, and I wonder if this one is still necessary?

scripts/wasmtestreport.py
```diff
@@ -260,7 +258,7 @@ def test_single_file_non_deterministic(source_file, result, timeout_sec=DEFAULT_
         else:
             if retcode == 0:
                 add_test_result(result, str(source_file), "Success", None, output)
-            elif retcode == 134 or retcode == 139:
+            elif retcode in (134, 139):
```
m-hemmings, line 261: This check is done more than once with the same check, would it be possible to abstract it to a function

scripts/wasmtestreport.py
```diff
@@ -296,7 +294,7 @@ def test_single_file_deterministic(source_file, result, timeout_sec=DEFAULT_TIME
     expected_output_file = source_file.parent / EXPECTED_DIRECTORY / f"{source_file.stem}.output"
 
     native_output = source_file.parent / f"{source_file.stem}.o"
-    native_compile_cmd = f"gcc {source_file} -o {native_output}"
+    native_compile_cmd = ["gcc", str(source_file), "-o", str(native_output)]
```
m-hemmings, line 297: This is nice change away from a string based command

scripts/wasmtestreport.py
```diff
@@ -622,6 +611,58 @@ def compare_test_results(file1, file2):
...
+#   timeout : Timeout for command execution.
+#
+# - Output:
+#   The result of subprocess.run()
+#
+# Raises:
+#   subprocess.TimeoutExpired - If the command exceeds the timeout.
+#   Exception - For all other unexpected execution errors.
+# ----------------------------------------------------------------------
+def run_subprocess(cmd, label="", cwd=None, shell=False, timeout=None):
```
m-hemmings, line 636: Could there be a bit of a guard rail here to prevent strange errors from calling this in the future? Something like expressly checking that if Shell=False, cmd must be a list whereas if Shell=True, cmd must be a string.  I know right now we're only doing one, but it would protect us later if the other was added.

m-hemmings (changes requested):

DyanB:
**Tasks done:**
Ref: [Issue 291](https://github.com/Lind-Project/lind-wasm/issues/291)
Allow running individual test files.
Go easy on the chdirs.
Improve output and debugging options.

**Ongoing tasks:**
Get rid of absolute paths.
Revise programming style.
Temporary directory for all test artifacts
Revise test setup and tear down.

PR #132, opened 2025-03-24
[Refactor] Syscall number unification in glibc

Yaxuan-w:

PR #326, opened 2025-08-13
[Refactor] Update wasmtime syscall num to align with linux version

Yaxuan-w:
This PR changes syscall nums in wasmtime to align with versions used in PR #132


rennergade:
I think this PR got screwed up some how. Theres a huge amount of commits here and i think it should only be the latest one?

PR #333, opened 2025-08-15
Add condition to prevent running on draft PRs

m-hemmings:
### PURPOSE

Prevent unnecessary runs of the e2e testing workflow on draft PRs to conserve CI resources and prevent spam sent out without cause.
This PR updates `.github/workflows/e2e.yml`  and `.github/lint.yml` to add a conditional check that skips the job when a draft PR is created.

### ISSUES
-  Avoids burning compute time on unnecessary e2e and lint tests
- reduces queue times for PRs ready for testing and review

### NOTES FOR REVIEWERS
- Draft PRs will still trigger the workflow, but should be marked as 'skipped' 
- No changes to the build steps or test setups, just an exception to avoid extra runs

Shows results as 'skipped' from a draft PR
image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -10,6 +10,12 @@ permissions: {}
 
 jobs:
   docker:
+    # If the workflow was triggered by anything other than a pull_request event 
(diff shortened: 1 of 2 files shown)


rennergade (approved):
Thanks!

PR #164, opened 2025-04-21
Various misc fixes for rawposix/glibc/wasmtime

yzhang71:
This pull request merges Qianxi's development branch, which includes cleanup of debugging code and redundant components, along with several important feature enhancements. These include the implementation of vmmap, support for signal handlers, a fix for locale-related issues, and expanded support for additional syscalls.

PR #342, opened 2025-08-18
rawposix - signal interruption for select and poll

qianxichen233:

This PR focuses on spliting out the signal interruption implementation for select and poll

The signal interruption for select and poll works by invoking system select and poll syscall with fixed timeout (100ms) to be able to periodically check for signal. It wakes up every 100ms once system select/poll is timed out, it then checks if it reached user timeout and signals.

### DIFF
--- a/src/RawPOSIX/src/interface/misc.rs
+++ b/src/RawPOSIX/src/interface/misc.rs
@@ -304,3 +304,44 @@ pub struct AdvisoryLock {
     advisory_lock: RustRfc<Mutex<i32>>,
     advisory_condvar: Condvar,
 }
+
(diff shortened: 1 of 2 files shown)


src/RawPOSIX/src/interface/misc.rs
```diff
@@ -304,3 +304,44 @@ pub struct AdvisoryLock {
...
+
+    let mut timeout;
+    if end_time > Duration::from_millis(100) {
+        timeout = 100;
+    } else {
+        timeout = timeout_ms;
+    }
+
+    (end_time, timeout)
+}
```
Yaxuan-w, line 347: Add new line

Yaxuan-w (changes requested):
Would be great to have more comments about implementation

Issue #78, opened 2025-01-06
Figure out ways to resolve function pointer casting in wasm


qianxichen233:
In python (and probably some other softwares), it is using function pointer casting to store the python method function pointers. This sometimes forcely changed the signature of the function. One example is `dict_keys` function, which has the signature of `PyObject *dict_keys(PyDictObject *mp)`. It is forcely casted to `PyObject *(*PyCFunction)(PyObject *, PyObject *)` in python and python invokes the `dict_keys` in form of `(*meth)(self, NULL)`. This causes wasm to raise `function type mismatch error`.

Calling a function pointer with extra argument is an undefined behavior according to c specification, but it looks like many developer does not care about this as long as this looks like working fine. But apparently wasm will not allow this behavior so we are facing some issues here. There is an [article](https://blog.pyodide.org/posts/function-pointer-cast-handling/) that also discussed exactly the same issue here. So looks like we currently have two options here:
1. Use emulated function pointer casting. This would basically rewrite every occurence of `indirect_call` instruction by making it always calling a function that takes 60 arguments. However, according to the article, this method could introduce a huge overhead to the wasm module
2. Another method would be just modifying the source code and fixing the undefined behavior here. This could be manually or automatically. Though having a general-purpose tool automatically fixing the function pointer casting would probably be pretty hard.

rennergade:
> How does WASM handle this normally? I can't believe we're the first to run into this.

They go over that in the article link Qianxi provided. Emscripten deals with it with option 1.

rennergade:
I've posted on Zulip asking the wasmtime folks if they deal with this in any explicit way.

rennergade:
https://bytecodealliance.zulipchat.com/#narrow/channel/206238-general/topic/Function.20pointer.20casting

It seems like we can use fpcast-emu with binaryen to deal with this, though it comes with the code bloat that Qianxi described.

rennergade:
This also might be useful: https://bytecodealliance.zulipchat.com/#narrow/channel/206238-general/topic/Function.20pointer.20casting/near/492409381

PR #346, opened 2025-08-20
glibc - qsort fix

qianxichen233:

This PR focuses on spliting out the fix for qsort function in glibc for wasm compatibility

In C, it is legal to cast a function pointer to another type (potentially with fewer parameters) and then invoke it with the “wrong” number of arguments. The qsort implementation in glibc relies on this trick. However, this behavior violates Wasm’s strict function signature rules and results in a runtime error under Wasm. More discussion [here](https://github.com/Lind-Project/lind-wasm/issues/78)

While it’s possible to work around this with wasm-opt’s emulated function pointer casting, that approach introduces significant overhead across the entire Wasm binary. Instead, this PR directly modifies glibc’s source to remove the qsort function pointer trick, avoiding the need for binary rewriting.

The main change is duplicating the qsort implementation with a function signature that has one fewer parameter. This eliminates the function pointer casting and makes qsort work in a Wasm environment.

### DIFF
--- a/src/glibc/stdlib/qsort.c
+++ b/src/glibc/stdlib/qsort.c
@@ -685,6 +685,23 @@ __qsort_internal (void *const pbase, size_t total_elems, size_t size,
     free (buf);
 }
 
+/*
+ * lind-wasm:
+ * In C, it's legal to cast a function pointer to another type (even with fewer
+ * parameters) and call it with the "wrong" number of arguments. glibc's qsort
+ * implementation relies on this trick.
+ *
+ * However, WebAssembly enforces strict function signatures, so this behavior
+ * causes a runtime error under Wasm.
+ *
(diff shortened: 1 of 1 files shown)


rennergade:
this needs comments explaining the reasons for updates in the code, and a bit more full description here

Yaxuan-w (changes requested):
It seems moving qsort operations to entirely in user space. I noticed the reason of this listed as "wasm compatibility". I'm not sure if the reason is callback function signature mismatch. Could you make it a bit detailed?

qianxichen233:
> It seems moving qsort operations to entirely in user space. I noticed the reason of this listed as "wasm compatibility". I'm not sure if the reason is callback function signature mismatch. Could you make it a bit detailed?

See updated PR description

PR #347, opened 2025-08-20
fix migration - glibc/wasmtime - libc fatal debug

qianxichen233:

This PR focuses on spliting out the temporary debug solution for glibc fatal. Previously, when glibc crushes, there is no debug information getting printed out, which is pretty inconvenient. The reason behind it is that the `libc_assert_fail` and `malloc_printerr` function in glibc is not working correctly. This PR provides an temporary solution to have the error message directly passed to wasmtime to print the error

### DIFF
--- a/src/wasmtime/crates/lind-common/src/lib.rs
+++ b/src/wasmtime/crates/lind-common/src/lib.rs
@@ -218,6 +218,33 @@ pub fn add_to_linker<
             wasmtime_lind_multi_process::signal::signal_handler(&mut caller);
         },
     )?;
+    
+    // a temporary solution to have libc_assert_fail correctly working
+    linker.func_wrap(
+        "debug",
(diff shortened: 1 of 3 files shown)


src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -218,6 +218,33 @@ pub fn add_to_linker<
             wasmtime_lind_multi_process::signal::signal_handler(&mut caller);
         },
     )?;
+    
+    // a temporary solution to have libc_assert_fail correctly working
+    linker.func_wrap(
```
rennergade: these are both basically write() syscalls, instead of adding these host functions can't we just use write? Once the addr translation PR is in that should be easy right? @qianxichen233 @drapl0n

PR #349, opened 2025-08-20
fix migration - wasmtime - asyncify & signal interruption

qianxichen233:

This PR focuses on spliting out the signal interruption implementation on wasmtime part. Added logic to have Asyncify supporting signal interruption on syscall

### DIFF
--- a/src/wasmtime/crates/lind-common/src/lib.rs
+++ b/src/wasmtime/crates/lind-common/src/lib.rs
@@ -4,7 +4,7 @@ use anyhow::Result;
 use rawposix::safeposix::dispatcher::lind_syscall_api;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


rennergade:
This is failing some tests, do we think its related to this PR: https://github.com/Lind-Project/lind-wasm/pull/350

Also need to fix for linter

src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -68,17 +68,54 @@ impl LindCommonCtx {
...
+                    call_name,
+                    arg1,
+                    arg2,
+                    arg3,
+                    arg4,
+                    arg5,
+                    arg6,
+                );
+
+                // Assumption: lind_syscall_api will not switch asyncify state, which holds true for now
```
rennergade, line 98: if we have an assumption like this we need to explain why its true

Issue #345, opened 2025-08-20
Unit test for waitpid with WNOHANG flag is failing


ssannkkallpp:
The unit test waitpid_wnohang.c under process unit tests fails with the following error:

`signal_callback export not foun`

We need to add this export during compilation.

rennergade:
This is probably resolved with the crt build being re-enabled. Do we have an easy way to confirm this?

PR #350, opened 2025-08-20
Pass signal_callback export during lind compilation

ssannkkallpp:
Regarding issue : https://github.com/Lind-Project/lind-wasm/issues/345

We need to explicitly pass the signal callback export to the compilation as a flag. This PR includes said change. 

@qianxichen233 @rennergade Would love your input on this.


rennergade:
@qianxichen233 would love your thoughts on this

Issue #334, opened 2025-08-15
Dockerfile.e2e - Misconfigured LIND_ROOT causes all fs unit tests to fail with ENOENT.


stupendoussuperpowers:
Currently, all unit tests that try to access a file fail with the syscalls returning an ENOENT.

LIND_ROOT is [hardcoded](https://github.com/Lind-Project/lind-wasm/blob/0e818d6f70cef5228847e6e3a062a2219613fdeb/src/sysdefs/src/constants/fs_const.rs#L11) to be `/home/lind/lind-wasm/src/RawPOSIX/tmp`. However when we use Dockerfile.e2e to build our test container this path doesn't exist on the container, and LIND_ROOT should instead be `/src/RawPOSIX/tmp`.

Due to this, regardless of what file you open inside a lind cage during a test, it returns a ENOENT. Even when you try to access "." or "/" (Since these convert to `/home/lind/lind-wasm/src/RawPOSIX/tmp/.` instead of `/src/RawPOSIX/tmp/.`)

We could either - 

- Add a way to specify a custom LIND_ROOT while building wasmtime.

- Follow what we do for the `Dockerfile.dev` image, where we create a `lind` user so that these paths are configured correctly.

PR #351, opened 2025-08-25
tests: get rid of absolute paths and add usage of temp directory to store test artifacts when needed

DyanB:
Reference issue: [Link](https://github.com/Lind-Project/lind-wasm/issues/291)

This PR takes care of some of the issues listed in Issue 291

Removed Absolute Paths
Test artifacts can now be retained in a specific directory of your choice / default temp directory

### DIFF
--- a/scripts/wasmtestreport.py
+++ b/scripts/wasmtestreport.py
@@ -459,7 +459,7 @@ def generate_html_report(report):
         html_content.append(f'<tr><td>Number of Successes</td><td>{test_result.get("number_of_success", 0)}</td></tr>')
         html_content.append(f'<tr><td>Number of Failures</td><td>{test_result.get("number_of_failures", 0)}</td></tr>')
         for error_type in error_types:
-            html_content.append(f'<tr><td>Number of {error_types[error_type]}</td><td>{test_result.get(f"number_of_{error_types[error_type]}", 0)}</td></tr>')
+            html_content.append(f'<tr><td>Number of {error_types[error_type]}</td><td>{test_result.get(f"number_of_{error_type}", 0)}</td></tr>')
         html_content.append('</table>')
 
     for test_type, test_result in report.items():


stupendoussuperpowers (dismissed):
Changes look good to me. This should solve #334.

scripts/wasmtestreport.py
```diff
@@ -764,27 +779,47 @@ def main():
...
-            native_file.unlink()
-    
-    shutil.rmtree(TESTFILES_DST) # removes the test files from the lind fs root
+    # Remove testfiles directory from lind fs root
+    try:
+        shutil.rmtree(TESTFILES_DST)
+    except FileNotFoundError:
+        pass
+
+    # Remove artifacts directory if it was temp and not requested to keep
```
m-hemmings, line 818: If the script crashes between creating the temp and here, will this linger?

scripts/wasmtestreport.py
```diff
@@ -725,6 +730,16 @@ def main():
         "non_deterministic": get_empty_result()
     }
 
+    # Prepare artifacts root
+    created_temp_dir = False
+    if artifacts_dir_arg:
+        artifacts_root = artifacts_dir_arg.resolve()
```
m-hemmings, line 736: if the artifacts dir is requested but not writable, this will behave unexpectedly.  Perhaps a try/except would help on it to make sure?

scripts/wasmtestreport.py
```diff
@@ -612,6 +613,8 @@ def parse_arguments():
     parser.add_argument("--clean-results", action="store_true", help="Flag to clean up result files")
     parser.add_argument("--testfiles", type=Path, nargs = "+", help="Run one or more specific test files")
     parser.add_argument("--debug", action="store_true", help="Enable detailed stdout/stderr output for subprocesses")
+    parser.add_argument("--artifacts-dir", type=Path, help="Directory to store build artifacts (default: temp dir)")
+    parser.add_argument("--keep-artifacts", action="store_true", help="Keep artifacts directory after run for troubleshooting")
```
m-hemmings, line 617: It seems to me that clean-results and keep-artifacts are mutually exclusive, conceptually, but there's nothing that prevents a user from asking for both right now.

DyanB: Have added a check so that the user does not provide contradictory flags.

scripts/wasmtestreport.py
DyanB: Have handled script crashing issues by encompassing the cleanup logic in the finally block.

PR #354, opened 2025-08-25
Fixes #352

### LINKED ISSUE #352: Fix `make clean` to avoid removing required build/target directories
While working with @ssannkkallpp about 3i integration, we found that current `make clean` in [Makefile](https://github.com/Lind-Project/lind-wasm/blob/main/Makefile#L48) breaks any future running. After we dived into it, the error is caused by `make clean` / `make distclean` removes both the build and target directories. However, these directories are still needed during the process of building glibc. Removing them breaks subsequent builds.

Instead, the clean target should only remove sysroot and the generated files, since these are the true intermediate artifacts created during compilation.

I am not yet 100% sure whether this list (sys root + .o files) fully covers all intermediate build outputs. Further verification may be needed by reviewing the build logs and identifying every file generated during compilation.

Temporary fix:

```Makefile
.PHONY: clean
clean:
	@echo "Cleaning glibc artifacts"
	$(RM) -r src/glibc/sysroot src/glibc/**/*.o
	@echo "Cleaning cargo (wasmtime)"
	cargo clean --manifest-path src/wasmtime/Cargo.toml
```

J7-7-7:
Temporary fix for #352, figuring out the paths for all intermediate build outputs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,8 @@ docs-serve:
 
 .PHONY: clean
 clean:
-	@echo "glibc artifacts"
-	$(RM) -r src/glibc/build src/glibc/sysroot src/glibc/target
+	@echo "cleaning glibc artifacts"
+	$(RM) -r src/glibc/sysroot src/glibc/**/*.o
 	@echo "cargo clean (wasmtime)"
 	cargo clean --manifest-path src/wasmtime/Cargo.toml
 


ssannkkallpp:
Looks fine to me. I did a quick sanity test on main with 
`make clean`
`make all`

And it seemed to fail during the rebuild. Not sure if it was something on my end but @Yaxuan-w @J7-7-7 let me know if you guys face a similar problem.

Makefile
```diff
@@ -46,8 +46,8 @@ docs-serve:
 
 .PHONY: clean
 clean:
-	@echo "glibc artifacts"
-	$(RM) -r src/glibc/build src/glibc/sysroot src/glibc/target
+	@echo "cleaning glibc artifacts"
+	$(RM) -r src/glibc/sysroot src/glibc/**/*.o
```
Yaxuan-w, line 50: I'm not sure if `globstar` has been enabled in our system. We can use `find` along with `rm` to clean. This should be more natural

Yaxuan-w (changes requested):

PR #357, opened 2025-08-27
glibc - missing syscalls fix

qianxichen233:

This PR focuses on spliting out the fix for missing syscalls on glibc side

These are some syscalls that some application is trying to use but we forgot to modify previously


rennergade:
These changes are ported to #458, closing this pr

PR #358, opened 2025-08-27
glibc - workaround for getpwuid functions

qianxichen233:

This PR focuses on spliting out the fix for getpwuid functions

`getpwuid` in glibc is used for retrieveing some system information. This function does not work for some reason. Here is a temporary workaround to make some applications happy


rennergade:
I'm going to close this PR since we definitely shouldn't solve this in this way.

I believe this should work for us as long as we have /etc/passwd and /etc/nsswitch.conf loaded correctly into our filesystem root.

PR #367, opened 2025-09-05
Add documentation for In Memory File System

stupendoussuperpowers:
This adds a brief design documentation for the in memory file system.

### DIFF
--- a/docs/internal/in-memory-file-system.md
+++ b/docs/internal/in-memory-file-system.md
@@ -1,14 +1,14 @@
-## In Memory File System 
-
-### Overview 
+# In Memory File System 
+ 
+## Overview 
 
 The In Memory File System (IMFS) provides a self-contained implementation of a POSIX-like FS backed by memory. It serves as a backbone that can later be integrated as a grate to sandbox any FS calls made by a cage. IMFS exposts POSIX-like APIs and maintains its own inode and file descriptor tables to provide an end-to-end FS interface.
 
-IMFS is independent of the Lind and can run natively on Linux, with the limitation that it only supports a single process. New features are usually developed and tested natively before being integrated and tested within Lind. 
+IMFS is independent of Lind and can run natively on Linux, with the limitation that it only supports a single process. New features are usually developed and tested natively before being integrated and tested within Lind. 
 
-### File System APIs
+## File System APIs
 
-IMFS mirrors POSIX system calls with the the added `cageid` paramter. For example:
+IMFS mirrors POSIX system calls with an added `cageid` parameter. For example:
 
 ```
 open(const char* pathname, int flags, mode_t mode)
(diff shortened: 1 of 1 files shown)


docs/internal/in-memory-file-system.md
```diff
@@ -0,0 +1,50 @@
+## In Memory File System 
+
+### Overview 
+
+The In Memory File System (IMFS) provides a self-contained implementation of a POSIX-like FS backed by memory. It serves as a backbone that can later be integrated as a grate to sandbox any FS calls made by a cage. IMFS exposts POSIX-like APIs and maintains its own inode and file descriptor tables to provide an end-to-end FS interface.
+
+IMFS is independent of the Lind and can run natively on Linux, with the limitation that it only supports a single process. New features are usually developed and tested nateively before beign integrated and tested within Lind.
```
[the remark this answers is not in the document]
stupendoussuperpowers: cageid can be any integer as long as the same one is used across all the calls. When testing natively I just pass 0.

docs/internal/in-memory-file-system.md
```diff
@@ -0,0 +1,50 @@
+## In Memory File System 
+
+### Overview 
+
+The In Memory File System (IMFS) provides a self-contained implementation of a POSIX-like FS backed by memory. It serves as a backbone that can later be integrated as a grate to sandbox any FS calls made by a cage. IMFS exposts POSIX-like APIs and maintains its own inode and file descriptor tables to provide an end-to-end FS interface.
+
+IMFS is independent of the Lind and can run natively on Linux, with the limitation that it only supports a single process. New features are usually developed and tested natively before being integrated and tested within Lind.
```
Yaxuan-w, line 7: If we figure out a way to convert fdtables to C binaries (Rust code to C), do you think it can be hooked with current IMFS?

docs/internal/in-memory-file-system.md
```diff
@@ -0,0 +1,50 @@
...
+
+IMFS mirrors POSIX system calls with the the added `cageid` paramter. For example:
+
+```
+open(const char* pathname, int flags, mode_t mode)
+->
+imfs_open(int cageid, const char* pathname, int flags, mode_t mode)
+```
+
+The behaviour of these APIs closely matches those of Linux's system calls, including types, return values, and error codes. This ensure that IMFS can be a drop-in replacement for an actual filesystem.
```
Yaxuan-w, line 19: > The behaviour of these APIs closely matches those of Linux's system calls, including types, return values, and error codes.

Does this mean IMFS APIs align with man page?

docs/internal/in-memory-file-system.md
```diff
@@ -0,0 +1,50 @@
...
+load_file(char *path); // Load a single file into IMFS at 'path', recursively creating any required folders. 
+
+dump_file(char *path, char *actual_path); // Copies IMFS file at 'path' to the host filesystem at 'actual_path'
+
+preloads(char *preload_files); // 'preload_files' is expected to be a ':' separated list of filenames that are each passed to load_file()
+```
+
+Typically, `preloads` or `load_file` is called at the start of a grate's lifecycle to stage the required files, and `dump_file` is called at the end to persist any results back to the host. 
+
+### Implementation
```
Yaxuan-w, line 35: Maybe adding a sub-title for each could be clearer

docs/internal/in-memory-file-system.md
```diff
@@ -0,0 +1,50 @@
...
+
+Each cage has its own array of `struct FileDesc` objects that represent a file descriptor. The file descriptors returned and used by the FS functions are indices into this array. FD allocations start at descriptor 3, the implementation for standard descriptors (stdin, stdout, stderr) are left to the enclosing grate. FileDescs are allocated through `open` or `openat` calls. Upon allocation, the FileDesc stores a pointer to the Node that was opened, the read offset, and the open flags. 
+
+### Testing 
+
+Testing for POSIX-compliance is checked through using `pjdfstest` which is a standard test suite used for BSD and Linux file systems. The tests are run natively on Linux which required modifying `pjdfstest` to have a persistent test runner to maintain state. `pjdfstest` provides a comprehensive list of one-liner assertions that are helpful to discover edge-cases and violations of the POSIX standards.
```
Yaxuan-w, line 43: Could we also have an example to run test here? (ie: command lines)

docs/internal/in-memory-file-system.md
Yaxuan-w, line 7: > Also, if you do run it natively, what does the cageID mean and how does it get integrated into code?

Adding your answer of Justin's question here should be helpful too

Yaxuan-w (commented):
Awesome!

stupendoussuperpowers:
Have made the suggested changes and polished up the structure a bit. I've also added more comments on the IMFS source code.

rennergade:
I think this has been ported to a new repo so lets close this PR.

PR #385, opened 2025-09-18
Create report script for clippy warnings

### LINKED ISSUE #384: Create report script for clippy warnings
In advance of re-enabling the clippy warnings, and before they become part of CI, add a script to provide a sample HTML report so everyone can see what will start showing in all CI after this is implemented.

tashabits:
### PURPOSE 
Add `clippy_report_generator.sh` script to generate a styled HTML report from `cargo clippy` output.
This script provides an easier way to review warnings/errors in a browser rather than parsing raw JSON or terminal logs.

- Resolve manifest path automatically relative to script location
- Support --manifest-path and --out flags for overrides
- Clean up temp files after run
- Generate a self-contained HTML report

### ISSUES
- Without this script, developers have to manually parse raw JSON or terminal output from `cargo clippy`, which is hard to review
- No existing HTML reporting mechanism exists for this

### NOTES FOR REVIEWERS
- Verified defaults work when from from project root
- Report is written to clippy_report.html at project root by default
- Script exits with Clippy's return code
- **REQUIRES** `jq`, `sed`, and `grep` to be present in the environment

### DIFF
--- a/scripts/clippy_report_generator.sh
+++ b/scripts/clippy_report_generator.sh
@@ -1,5 +1,36 @@
 #!/usr/bin/env bash
+
 # scripts/clippy_report_html.sh
+#
+# -----------------------------------------------------------------------------
+# Clippy → HTML Report Generator
+#
+# USAGE (from project root):
+#   ./scripts/clippy_report_generator.sh [--manifest-path <path>] [--out <file>]
+#
+# DEFAULTS (resolved relative to the script’s directory):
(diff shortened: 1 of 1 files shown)


J7-7-7 (approved):

DyanB (approved):
looks good to me!

rennergade:
This is helpful and we'll definitely use it when cleaning up the clippy errors. I'm not sure we should host it in our repo though since its mostly unrelated to Lind.

Could be a cool tool for you to host somewhere that we'll use?

PR #386, opened 2025-09-18
Add support to upload an artifact of the html report from wasmtestrep…

tashabits:
### PURPOSE
Expand the GHA workflow for e2e testing to surface the generated HTML test report as an artifact.
This maes the wasm test results easy to download and view from the Actions run summary page.
The actions it takes are:
- Copy the generate `report.html` file out of the e2e Docker image
- Store it locally on the runner as `test-reports/wasm-e2e-report.html`
- Upload the file as a named artifact "`wasm-e2e-report`"

### ISSUES
- Without this, the HTML report created inside the container remains inaccessible

### NOTES FOR REVIEWERS
- Only the CI workflow was updated, no changes to source code or logic
- Artifact retention was set to 7 days; can be tuned as needed
- Report artifact appears under "Artifacts" in the GitHub Ations run details

### DIFF
--- a/scripts/templates/e2e_comment.md.j2
+++ b/scripts/templates/e2e_comment.md.j2
@@ -4,49 +4,7 @@
 <details>
   <summary><strong>HTML preview</strong></summary>
 
-<style>
-/* Success rows: bold and readable */
-tr[style*="lightgreen"] td,
-tr[style*="lightgreen"] th,
-tr[style*="lightgreen"] pre {
-  font-weight: 700;
-  color: #000;
-}
 
-/* Output cells — fix dark grey backgrounds */
-td pre,
-td[style*="#333"],
-td[style*="rgb(51"] {
-  background: none !important;   /* strip the dark box in light mode */
-  color: #000 !important;
-}
-
-/* Adjust for dark mode */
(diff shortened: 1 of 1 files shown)


m-hemmings:
> If I'm seeing this right we still need to download a zip and open the enclosed html ourselves right? there's no way to directly view?

That's right, you do.  I can try and find a free host, but what we were doing before won't work, because there isn't just an open bucket to post it to.  The normal solution is to use github pages, but that's being used for docs, and the documentation from it says it would overwrite the docs if we used it for this

J7-7-7 (commented):
If we can't use github pages, is this the only way? Are there any alternatives?

m-hemmings:
> If we can't use github pages, is this the only way? Are there any alternatives?

I can check if I can do a comment on the PR with a summary and a link, maybe

rennergade:
This seems to be failing e2e

m-hemmings:
> This seems to be failing e2e

Yes, I'm actively working on it.  It works without the comment being pushed, I'm trying to add that for you

m-hemmings:
> First cut, comment here

This comment was made by the GHA, I'll update it to be more useful, but didn't want this to be confusing

Issue #365, opened 2025-09-05
Add new syscall constants in sysdefs library


Yaxuan-w:
Per discussion in https://github.com/Lind-Project/lind-wasm/pull/237#discussion_r2323256119, we could add a constant file in sysdef library and use macro in syscall table instead of using hardcoded numbers. 

After this we should also remove the [lind_syscall_num.rs file in wasmtime](https://github.com/Lind-Project/lind-wasm/blob/threei-integration/src/wasmtime/crates/lind-utils/src/lind_syscall_numbers.rs) and replace all imports in wasmtime by same constant library.

rennergade:
I believe we can use https://github.com/rust-lang/rust-bindgen to do most of this

PR #388, opened 2025-09-18
 [Refactor] Implement Filesystem Syscalls

ChinmayShringi:
## PURPOSE

Implements 9 essential filesystem syscalls to achieve feature parity between tmp-main-2 and main branches. All implementations follow established patterns and include comprehensive documentation.

**Syscalls Added:**
- ACCESS (21), UNLINK (87), LINK (86), RENAME (82)
- XSTAT (4), FSYNC (74), FDATASYNC (75) 
- SYNC_FILE_RANGE (277), READLINKAT (267)

**Implementation Details:**
- All syscalls follow consistent conversion patterns (sc_convert_path_to_host, convert_fd_to_host)
- Logic matches original main branch implementations exactly
- Added enterprise-grade documentation with Linux man page references
- Proper syscall number mappings in threei syscall table

**Files Modified:**
- `src/rawposix/src/fs_calls.rs` - 9 new syscall implementations
- `src/threei/src/syscall_table.rs` - syscall mappings and imports

**Testing:** All implementations use established error handling and validation patterns. Linter checks pass.


src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+    let virtual_fd = sc_convert_sysarg_to_i32(fd_arg, fd_cageid, cageid);
+
+    // Validate unused args
+    if !(sc_unusedarg(arg2, arg2_cageid)
+        && sc_unusedarg(arg3, arg3_cageid)
+        && sc_unusedarg(arg4, arg4_cageid)
+        && sc_unusedarg(arg5, arg5_cageid)
+        && sc_unusedarg(arg6, arg6_cageid))
+    {
+        return syscall_error(Errno::EFAULT, "fsync", "Invalid Cage ID");
```
JustinCappos, line 1171: Is this the right error?  It seems like at a minimum, you want to differentiate this from the case below...

As above, I would even argue that any error of this type should be an internal error / panic (for the time being).

(This comment applies throughout your code here.)

One of "Justin's rules of good coding" is that from a fatal error trace, you should know exactly where something failed.

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+    // Type conversion
+    let _vers = sc_convert_sysarg_to_i32(vers_arg, vers_cageid, cageid);
+    let path = sc_convert_path_to_host(path_arg, path_cageid, cageid);
+
+    // Validate unused args
+    if !(sc_unusedarg(arg4, arg4_cageid)
+        && sc_unusedarg(arg5, arg5_cageid)
+        && sc_unusedarg(arg6, arg6_cageid))
+    {
+        return syscall_error(Errno::EFAULT, "xstat", "Invalid Cage ID");
```
JustinCappos, line 1092: How do you know this is a problem with the Cage ID?

I would even argue that any error of this type should be an internal error / panic (for the time being).

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+        // Check if the fd is AT_FDCWD
+        let cage_arc = get_cage(cageid).unwrap();
+        let cwd_container = cage_arc.cwd.read();
+        path = format!("{}/{}", cwd_container.to_str().unwrap(), path);
+        // Convert the path from relative path (lind-wasm perspective) to real kernel path (host kernel
+        // perspective)
+        let relpath = normpath(convpath(&path), cageid);
+        let relative_path = relpath.to_str().unwrap();
+        let full_path = format!("{}{}", LIND_ROOT, relative_path);
+        let c_path = std::ffi::CString::new(full_path).unwrap();
```
JustinCappos, line 1375: I dislike that we're doing this in trusted code because there really isn't a reason to do so.  I'd much prefer this lived in a grate.  However that grate currently doesn't exist.  So, if need be, we can fix this later by moving the code then.

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+            return syscall_error(Errno::EFAULT, "readlinkat", "Invalid Cage ID");
+        } else if kernel_fd == -9 {
+            return syscall_error(Errno::EBADF, "readlinkat", "Bad File Descriptor");
+        }
+        // Convert the path from relative path (lind-wasm perspective) to real kernel path (host kernel
+        // perspective)
+        let relpath = normpath(convpath(&path), cageid);
+        let relative_path = relpath.to_str().unwrap();
+        let full_path = format!("{}{}", LIND_ROOT, relative_path);
+        let c_path = std::ffi::CString::new(full_path).unwrap();
```
JustinCappos, line 1397: Same here.  I'd prefer all of this path munging moved to a grate.

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+    arg5: u64,
+    arg6: u64,
+    arg7: u64,
+    arg8: u64,
+    arg9: u64,
+    arg10: u64,
+    arg11: u64,
+    arg12: u64,
+) -> i32 {
+    let oldpath = sc_convert_path_to_host(oldpath_arg, oldpath_cageid, cageid);
```
JustinCappos, line 1468: Do we handle the case where this path conversion exceeds the filename, etc. limit?

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+    pathname_arg: u64,
+    pathname_cageid: u64,
+    flags_arg: u64,
+    flags_cageid: u64,
+    _arg7: u64,
+    _arg8: u64,
+    _arg9: u64,
+    _arg10: u64,
+    _arg11: u64,
+    _arg12: u64,
```
JustinCappos, line 1565: Shouldn't you be checking for unused args like above?

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+            )
+        }
+    } else {
+        // Case 2: Specific directory fd
+        let kernel_fd = convert_fd_to_host(virtual_fd as u64, dirfd_cageid, cageid);
+        if kernel_fd == -1 {
+            return syscall_error(Errno::EFAULT, "unlinkat", "Invalid Cage ID");
+        } else if kernel_fd == -9 {
+            return syscall_error(Errno::EBADF, "unlinkat", "Bad File Descriptor");
+        }
```
JustinCappos, line 1587: what is this magic?  What are -1 and -9?

src/threei/src/syscall_table.rs
```diff
@@ -15,11 +16,13 @@ pub const SYSCALL_TABLE: &[(u64, Raw_CallFunc)] = &[
     (1, write_syscall),
     (2, open_syscall),
     (3, close_syscall),
+    (4, stat_syscall),
```
JustinCappos, line 19: Do we have macros for these?  It seems logical to me we would.

JustinCappos (changes requested):
Initial comments / request for fixes.  A few here are worth @rennergade, @Yaxuan-w , and @qianxichen233 looking at.

src/rawposix/src/fs_calls.rs
Yaxuan-w, line 1375: Should we do path conversion by using type conversion function?

src/rawposix/src/fs_calls.rs
Yaxuan-w, line 1397: Same with previous. Using type conversion function

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+    arg9: u64,
+    arg10: u64,
+    arg11: u64,
+    arg12: u64,
+) -> i32 {
+    let oldpath = sc_convert_path_to_host(oldpath_arg, oldpath_cageid, cageid);
+    let newpath = sc_convert_path_to_host(newpath_arg, newpath_cageid, cageid);
+
+    let old_c_path = std::ffi::CString::new(oldpath.to_str().unwrap()).unwrap();
+    let new_c_path = std::ffi::CString::new(newpath.to_str().unwrap()).unwrap();
```
Yaxuan-w, line 1472: I'm a bit confused here... Why we need to do this?

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+    arg8: u64,
+    arg9: u64,
+    arg10: u64,
+    arg11: u64,
+    arg12: u64,
+) -> i32 {
+    let pathname = sc_convert_path_to_host(pathname_arg, pathname_cageid, cageid);
+    let mode = sc_convert_sysarg_to_i32(mode_arg, mode_cageid, cageid);
+
+    let c_path = std::ffi::CString::new(pathname.to_str().unwrap()).unwrap();
```
Yaxuan-w, line 1520: Same here

src/threei/src/syscall_table.rs
Yaxuan-w: The `syscall_table.rs` is the mappings between syscallnum to function ptr in rawposix, so we couldn't actually use them as constants in other places. 

Here's an issue #365 for unifying syscall num in a constant file under sysdefs library, which would also unify syscall constant file in wasmtime to one place.

src/rawposix/src/fs_calls.rs
```diff
@@ -987,3 +989,616 @@ pub fn fcntl_syscall(
...
+ *   - The path is converted from the RawPOSIX perspective to the host kernel perspective
+ *     using `sc_convert_path_to_host`, which handles the LIND_ROOT prefixing and path normalization.
+ *   - The statbuf buffer is converted from WASM address to host address using `sc_convert_addr_to_host`.
+ *   - The underlying libc::stat() is called and results are copied to the user buffer.
+ *
+ *  ## Return Value:
+ *   - `0` on success.
+ *   - `-1` on failure, with `errno` set appropriately.
+ */
+pub fn stat_syscall(
```
Yaxuan-w, line 1068: We need to have an additional type conversion function to convert the `stat_arg` to our defined stat struct. 

After receiving the results from kernel, we need to cp+paste results from kernel struct to our defined stat struct

Yaxuan-w (changes requested):

src/rawposix/src/fs_calls.rs
ChinmayShringi: Added better error message

src/rawposix/src/fs_calls.rs
```diff
@@ -131,6 +145,9 @@ pub fn read_syscall(
 ) -> i32 {
     // Convert the virtual fd to the underlying kernel file descriptor.
     let kernel_fd = convert_fd_to_host(vfd_arg, vfd_cageid, cageid);
+    // convert_fd_to_host returns negative errno values on error:
+    // -1 = -EINVAL (cage ID validation failed)
+    // -9 = -EBADF (virtual fd not found in fdtables or invalid)
     if kernel_fd == -1 {
```
Yaxuan-w, line 151: Could we use constants from sysdefs and replace -1 & -9?

src/rawposix/src/fs_calls.rs
ChinmayShringi: Added a panic wrapper please let me know if that works

src/rawposix/src/fs_calls.rs
ChinmayShringi: Added conversion

src/rawposix/src/fs_calls.rs
ChinmayShringi: Updated

src/rawposix/src/fs_calls.rs
ChinmayShringi: Added conversion

src/rawposix/src/fs_calls.rs
ChinmayShringi: Updated to use sysdef constatns

src/rawposix/src/fs_calls.rs
```diff
@@ -1,14 +1,30 @@
...
 use sysdefs::constants::fs_const::{STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO, O_CLOEXEC, MAP_ANONYMOUS, MAP_FIXED, MAP_PRIVATE, MAP_SHARED, PROT_EXEC, PROT_NONE, PROT_READ, PROT_WRITE, PAGESHIFT, PAGESIZE};
-use sysdefs::constants::lind_platform_const::{FDKIND_KERNEL, MAXFD};
+use sysdefs::constants::lind_platform_const::{FDKIND_KERNEL, MAXFD, LIND_ROOT};
 use sysdefs::constants::sys_const::{DEFAULT_UID, DEFAULT_GID};
+use sysdefs::data::fs_struct::StatData;
 use typemap::cage_helpers::*;
 use cage::{round_up_page, get_cage, HEAP_ENTRY_INDEX, MemoryBackingType, VmmapOps};
 use fdtables;
 
+/// Panic wrapper for unused argument validation failures
```
rennergade, line 14: If this is helpful we should have this hosted in a different file and be used in other PRs. @Yaxuan-w thoughts?

src/rawposix/src/fs_calls.rs
```diff
@@ -18,7 +34,7 @@ use fdtables;
 /// This function is registered in `fdtables` when creating the cage
 pub fn kernel_close(fdentry: fdtables::FDTableEntry, _count: u64) {
     let kernel_fd = fdentry.underfd as i32;
-
+
```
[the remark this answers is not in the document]
Yaxuan-w: Updated

rennergade (commented):
Only concern is same as other PR where we seem to deal with paths sometimes but not others

Yaxuan-w (approved):

PR #391, opened 2025-09-21
Handle NULL Buffer Address in Threei Impl

yzhang71:
Purpose:
Update `translate_vmmap_addr` to safely handle NULL pointers by returning `None` when the input address (`arg`) is 0. 

Additionally, update related code paths to consistently handle the `None` case, ensuring system calls gracefully return appropriate errors (e.g., `EINVAL`) where applicable.


robinyuan1002 (approved):
LGTM

PR #393, opened 2025-09-22
docs: end to end testing 

### LINKED ISSUE #250: docs: describe CI and local end-to-end testing in contributor docs
#240 adds a new build pipeline to run lind-wasm end-to-end tests locally and on GitHub Actions using Docker. We should document how it works (e.g. wrt caching) and how to use it locally. 

The PR description has plenty of details.

J7-7-7:
Documentation for end to end testing, closes #250. Let me know if any additional information should be included.

### DIFF
--- a/docs/contribute/e2e-testing.md
+++ b/docs/contribute/e2e-testing.md
@@ -1,0 +1,88 @@
+# End-to-End testing
+
+Multi-stage **.e2e** flow for lind-wasm **end-to-end testing and image creation**.
+
+- Installs build dependencies
+- Builds **wasmtime**, **glibc**, and a **sysroot** for clang cross-compilation
+- **A.** Runs end-to-end tests (default)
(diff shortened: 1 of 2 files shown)


m-hemmings (commented):
Suggestion: expand the docs with a bit more detail in two areas:

###  Build steps

In the `make sysroot section`, note that multiple components compile and some .o files are used to build a library.

In the make test section, explain how `wasmtestreport.py` chooses which tests to run and how it generates the report.

### Caching

Clarify that `cache-from: type=gha` pulls from the GitHub Actions hosted cache, while `cache-to: type=gha,mode=max` saves all usable layers (not just the final image).

Mention that the first cold run builds all layers, but after that only changed layers are rebuilt, dramatically speeding up CI runs.

Note that GHA may evict cache layers outside our control; if that happens, a run will start cold but still pass the same way.

PR #395, opened 2025-09-22
Selective file copying and code restructure for wasmtestreport test suite

DyanB:
Reference issue: https://github.com/Lind-Project/lind-wasm/issues/291

Final PR to handle all of the remaining issues listed in Issue 291

1. Added selective file copying (All testfiles were copied previously irrespective what set of tests are run).
2. Code restructure and improvements.


scripts/wasmtestreport.py
```diff
@@ -404,17 +490,53 @@ def test_single_file_deterministic(source_file, result, timeout_sec=DEFAULT_TIME
...
+        os.makedirs(TESTFILES_DST, exist_ok=True)
+        shutil.copytree(TESTFILES_SRC, TESTFILES_DST, dirs_exist_ok=True)
+    
+    # Always create the readlinkfile symlink
     readlinkfile_path = TESTFILES_DST / "readlinkfile.txt"
     symlink_path = TESTFILES_DST / "readlinkfile"
     open(readlinkfile_path, 'a').close()
-    os.symlink(readlinkfile_path, symlink_path)
+    if not symlink_path.exists():
+        os.symlink(readlinkfile_path, symlink_path)
```
m-hemmings, line 539: Later on in this, you have a place where you use os.symlink with a fallback for OSError in case of problems, is it worth mimicing that here? This is the snippet I mean, from `setup_test_file_in_artifacts`

```
# Create symlink or copy file
    if not dest_source.exists():
        try:
            dest_source.symlink_to(original_source)
        except OSError:
            shutil.copy2(original_source, dest_source)
```

scripts/wasmtestreport.py
```diff
@@ -695,6 +817,109 @@ def run_subprocess(cmd, label="", cwd=None, shell=False, timeout=None):
...
+        
+        dest_source = setup_test_file_in_artifacts(original_source, artifacts_root)
+        
+        # Determine test type and run appropriate test
+        parent_name = original_source.parent.name
+        if parent_name == DETERMINISTIC_PARENT_NAME:
+            test_single_file_deterministic(dest_source, results["deterministic"], timeout_sec)
+        elif parent_name == NON_DETERMINISTIC_PARENT_NAME:
+            test_single_file_non_deterministic(dest_source, results["non_deterministic"], timeout_sec)
+
```
m-hemmings, line 922: If a test file lives in a folder other than deterministic or non_deterministic, it silently does nothing. This can make mis-tagged or malformed tests vanish without any feedback.

I’d recommend adding an else branch that logs a warning but continues execution.

Maybe just log it as an unknown file and keep going?

scripts/wasmtestreport.py
```diff
@@ -154,6 +154,129 @@ def add_test_result(result, file_path, status, error_type, output):
...
+# Function: compile_and_run_native
+#
+# Purpose:
+#   Compile and run native version of a test
+#
+# Variables:
+# - Input: source_file - path to the .c file, timeout_sec - timeout for execution
+# - Output: (success, output, returncode, error_type) tuple
+# ----------------------------------------------------------------------
+def compile_and_run_native(source_file, timeout_sec=DEFAULT_TIMEOUT):
```
m-hemmings, line 221: There are two spots here where `run_subprocess` is called with file paths.  If they're always absolute, no issues, but if a relative path sneaks through somehow, running with `cwd=LIND_FS_ROOT` could lead to some confusing issues where it can't find the file.

Might be worth adding an early check to see if the path is absolute or not and failing if it isn't?

m-hemmings (changes requested):

J7-7-7 (commented):
Matt covered the important changes. A couple small follow-ups that would help, filter the skip list before staging so we don’t copy/log files we’ll ignore anyway, de-duplicate the selected files (multiple globs can hit the same path) and use shutil.copy2 instead of copy so mtimes are preserved and downstream caching behaves better.

scripts/wasmtestreport.py
DyanB: Added the fallback case! Thanks for pointing it out!

DyanB:
> Matt covered the important changes. A couple small follow-ups that would help, filter the skip list before staging so we don’t copy/log files we’ll ignore anyway, de-duplicate the selected files (multiple globs can hit the same path) and use shutil.copy2 instead of copy so mtimes are preserved and downstream caching behaves better.

setup_test_environment() function takes care of the skip list. I'm already using a set to handle duplicates and i'm using shutil.copy2 throughout!

m-hemmings (approved):
My change requests were addressed, looks good to me now

Issue #242, opened 2025-06-17
wasmtime clippy build fails starting in nightly-2025-06-09


lukpueh:
When building wasmtime with cargo clippy nightly-2025-06-09 or above, it fails with below errors. (in nightly-2025-06-08 they don't) 

```bash
$ rustup default nightly-2025-06-09
info: using existing install for 'nightly-2025-06-09-x86_64-unknown-linux-gnu'
info: default toolchain set to 'nightly-2025-06-09-x86_64-unknown-linux-gnu'

  nightly-2025-06-09-x86_64-unknown-linux-gnu unchanged - rustc 1.89.0-nightly (6ccd44760 2025-06-08)
```

```bash
$ cargo clippy --manifest-path src/wasmtime/Cargo.toml --all-features --keep-going -- -A warnings -A clippy::not_unsafe_ptr_arg_deref -A clippy::absurd_extreme_comparisons

[...]

error: implicit autoref creates a reference to the dereference of a raw pointer
   --> crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs:107:18
    |
107 |         unsafe { (*self.memory.as_ptr()).len() }
    |                  ^^--------------------^^^^^^^
    |                    |
    |                    this raw pointer has type `*mut [u8]`
    |
    = note: creating a reference requires the pointer target to be valid and imposes aliasing requirements
note: autoref is being applied to this expression, resulting in: `&[u8]`
   --> crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs:107:18
    |
107 |         unsafe { (*self.memory.as_ptr()).len() }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^
note: method calls to `len` require a reference
   --> /rustc/6ccd4476036edfce364e6271f9e190ec7a2a1ff5/library/core/src/slice/mod.rs:115:5
    = note: `#[deny(dangerous_implicit_autorefs)]` on by default
help: try using a raw pointer method instead; or if this reference is intentional, make it explicit
    |
107 |         unsafe { (&(*self.memory.as_ptr())).len() }
    |                  ++                       +

error: implicit autoref creates a reference to the dereference of a raw pointer
   --> crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs:152:23
    |
152 |             let len = (*self.memory.as_ptr()).len();
    |                       ^^--------------------^^^^^^^
    |                         |
    |                         this raw pointer has type `*mut [u8]`
    |
[...]

PR #397, opened 2025-09-23
fixes wasmtime clippy build fails starting in nightly-2025-06-09

J7-7-7:
Fixes issue #242, wasmtime clippy build fails starting in nightly-2025-06-09

### DIFF
--- a/Docker/Dockerfile.dev
+++ b/Docker/Dockerfile.dev
@@ -64,7 +64,7 @@ WORKDIR /home/${USERNAME}
 
 # --- Install Rust as user ---
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
-    sh -s -- -y --default-toolchain nightly-2025-06-01
+    sh -s -- -y --profile minimal
 
 # --- Clone the repo as user ---
 RUN git clone --branch "${BRANCH_NAME}" --single-branch https://github.com/Lind-Project/lind-wasm.git
--- a/Docker/Dockerfile.dev
+++ b/Docker/Dockerfile.dev
@@ -63,6 +63,7 @@ USER ${USERNAME}
 WORKDIR /home/${USERNAME}
 
 # --- Install Rust as user ---
+# --profile minimal tells rustup to install only the essentials for the toolchain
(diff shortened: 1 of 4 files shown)


m-hemmings (approved):
Looks good, passes e2e

DyanB (dismissed):

Yaxuan-w (commented):
GitHub doesn't allow me to comment on the file... might be a GitHub problem, so I'm going to left my question here:

I saw there're changes in wasmtime which just adding `&`. I'm curious about the reason of this change. Is this cause of new rust feature/ etc.?

rennergade:
> The current use of `(&*self.memory.as_ptr()).len()` makes sense. However, since [`memory` is defined on same file line 11](https://github.com/Lind-Project/lind-wasm/blob/main/src/wasmtime/crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs#L11) as `SendSyncPtr<[u8]>`, and in Wasmtime [`SendSyncPtr::as_ptr` expands to return a `*mut T`](https://github.com/Lind-Project/lind-wasm/blob/main/src/wasmtime/crates/wasmtime/src/runtime/vm/send_sync_ptr.rs#L23), in this case it resolves to `*mut [u8]` according to [Wasmtime `SendSyncPtr` definition](https://github.com/Lind-Project/lind-wasm/blob/main/src/wasmtime/crates/wasmtime/src/runtime/vm/send_sync_ptr.rs#L9). For slice raw pointers, Rust provides a safe `.len()` method that reads the length metadata without dereferencing the pointer.
> 
> This means we can simplify the expression to `self.memory.as_ptr().len()`. In fact, the latest official release of Wasmtime also uses `self.memory.as_ptr().len()` on both [`fn len`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs#L136) and [`fn drop`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs#L231).
> 
> To stay aligned with upstream and ease future maintenance, I would recommend using that form here.

That's a great catch, I agree

Yaxuan-w (approved):

rennergade (approved):

PR #161, opened 2025-04-16
Network-related syscall typemap

robinyuan1002:

PR #160, opened 2025-04-16
[Refactor] Add network syscalls in rawposix [socket / connect / bind / listen / accept / setsockopt / send]

robinyuan1002:
This PR provide socket / connect / bind / listen / accept / setsockopt / send syscalls implementation.

PR #412, opened 2025-09-25
[Refactor] Add basic network syscalls to 3i (socket / connect / bind / listen / accept / setsockopt / send)

Yaxuan-w:
This PR consolidates PR #160 and PR #161 into a single change set and addresses all review comments raised on those PRs. The work was done in collaboration with @robinyuan1002.

- Implement: socket, connect, bind, listen, accept, setsockopt, and send
- our minimal sockaddr definition plus the conversion helpers
- simple single-thread test cases. 

All changes have been validated on the threei-integration-test branch.


src/rawposix/src/lib.rs
rennergade, line 1: confused about this empty file

src/typemap/src/network_helpers.rs
```diff
@@ -0,0 +1,133 @@
...
+use sysdefs::constants::net_const::{AF_UNIX};
+use cage::{get_cage, memory::memory::translate_vmmap_addr};
+use crate::datatype_conversion::validate_cageid;
+use std::os::raw::{c_void, c_char};
+use std::ptr;
+
+/// `unix_len_from_sun_path` computes the effective `socklen_t` for an `AF_UNIX` 
+/// `sockaddr_un` by applying the Linux length rules to its `sun_path`. On Linux, 
+/// `offsetof(sockaddr_un, sun_path)` is 2 because `sa_family_t` is a 16-bit field; 
+/// this fixed offset is treated as the base. If the first byte of `sun_pat` is zero,
```
rennergade, line 18: path*

src/typemap/src/network_helpers.rs
```diff
@@ -0,0 +1,133 @@
...
+/// the address is in the abstract namespace and the effective length is the base 
+/// plus the index of the last non-zero byte plus one, with no trailing NULL added. 
+///
+/// If the first byte is non-zero, the address is a pathname and the effective length 
+/// is the base plus the strlen of the path plus one to include the terminating NULL, 
+/// provided there is room in the 108-byte array; if the array is completely full 
+/// with no NUL, the length becomes the base plus 108. The function assumes the caller
```
rennergade, line 25: NULL*

maybe run this comment through a checker.

src/typemap/src/network_helpers.rs
```diff
@@ -0,0 +1,133 @@
...
+use sysdefs::data::net_struct::SockAddr;
+use libc::{sockaddr, strlen};
+use sysdefs::constants::lind_platform_const::LIND_ROOT;
+use sysdefs::constants::net_const::{AF_UNIX};
+use cage::{get_cage, memory::memory::translate_vmmap_addr};
+use crate::datatype_conversion::validate_cageid;
+use std::os::raw::{c_void, c_char};
+use std::ptr;
+
+/// `unix_len_from_sun_path` computes the effective `socklen_t` for an `AF_UNIX`
```
rennergade, line 15: It may be fine but I've been staring at this function even after reading the comments and having a pretty difficult time parsing it, may be better to have some of the comments inline.

rennergade (changes requested):
Left a few comments, not in bad shape

src/rawposix/src/lib.rs
Yaxuan-w: Thanks for pointing out! Updated.

src/rawposix/src/net_calls.rs
```diff
@@ -0,0 +1,404 @@
...
+/// ## Input:
+///     - cageid: current cageid
+///     - domain_arg: communication domain (e.g., AF_INET, AF_UNIX)
+///     - socktype_arg: socket type (e.g., SOCK_STREAM, SOCK_DGRAM)
+///     - protocol_arg: protocol to be used (usually 0)
+///
+/// ## Return:
+///     - On success: a newly allocated virtual file descriptor within the current cage
+///     - On failure: a negative errno value indicating the syscall error
+pub fn socket_syscall(
```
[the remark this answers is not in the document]
Yaxuan-w: We won't store nonblock info in fdtables I think. Added setting cloexec flag logic. Also updated comments.

src/typemap/src/network_helpers.rs
```diff
@@ -0,0 +1,133 @@
...
+/// the address is in the abstract namespace and the effective length is the base 
+/// plus the index of the last non-zero byte plus one, with no trailing NULL added. 
+///
+/// If the first byte is non-zero, the address is a pathname and the effective length 
+/// is the base plus the strlen of the path plus one to include the terminating NULL, 
+/// provided there is room in the 108-byte array; if the array is completely full 
+/// with no NUL, the length becomes the base plus 108. The function assumes the caller 
+/// provides a valid 108-byte buffer taken from a properly laid-out `sockaddr_un`.
+unsafe fn unix_len_from_sun_path(sun_path: &[i8; 108]) -> libc::socklen_t {
+    // offsetof(sockaddr_un, sun_path) == 2 (sa_family_t is u16)
```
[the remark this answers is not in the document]
Yaxuan-w: This's not the actual code, just explain the logic is that: the offset of (sockaddr_un, sun_path) is 2, because type of `sa_family_t` is `u16`. Thanks for pointing this out! Updated comments

src/rawposix/src/net_calls.rs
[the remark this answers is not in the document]
rennergade: I thought we added perfdinfo for flags?

rennergade (approved):
This looks good to me. Now I'm still a little confused about how we deal with blocking.

tests/unit-tests/networking_tests/deterministic/socket_cloexec.c
```diff
@@ -0,0 +1,31 @@
...
+    if (flags < 0) {
+        perror("fcntl(F_GETFD)");
+        close(fd);
+        return 1;
+    }
+
+    if (flags & FD_CLOEXEC) {
+        printf("SOCK_CLOEXEC is set.\n");
+    } else {
+        printf("SOCK_CLOEXEC is NOT set.\n");
```
stupendoussuperpowers, line 26: Should the failed case also return with exit code 0? Are we relying on comparing output from the native case here?

tests/unit-tests/networking_tests/deterministic/tcp_connect_single.c
```diff
@@ -0,0 +1,54 @@
...
+        return 7;
+
+    // ---- 4) send ----
+    static const char msg1[] = "ping";
+    if (send(s_client, msg1, sizeof(msg1)-1, 0) < 0)
+        return 8;
+
+    static const char msg2[] = "pong";
+    if (send(s_conn, msg2, sizeof(msg2)-1, 0) < 0)
+        return 9;
```
stupendoussuperpowers, line 50: Should we also perror() here so that we can see what these called failed with? Much like we use it in the cloexec case.

src/threei/src/syscall_table.rs
```diff
@@ -40,6 +43,15 @@ pub const SYSCALL_TABLE: &[(u64, RawCallFunc)] = &[
     (33, dup2_syscall),
     (35, nanosleep_time64_syscall),
     (39, getpid_syscall),
+    (41, socket_syscall),
+    (42, connect_syscall),
+    (43, accept_syscall),
+    (46, send_syscall),
```
stupendoussuperpowers, line 49: Should rename these functions. For e.g. 44 is sendto, 46 is sendmsg.

src/rawposix/src/net_calls.rs
```diff
@@ -0,0 +1,522 @@
...
+///     - cageid: current cageid
+///     - fd_arg: virtual file descriptor indicating the socket to send data on
+///     - buf_arg: pointer to the message buffer in user memory
+///     - buflen_arg: length of the message to be sent
+///     - flags_arg: bitmask of flags influencing message transmission behavior
+///
+/// ## Return:
+///     - On success: number of bytes sent
+///     - On failure: a negative errno value indicating the syscall error
+pub fn send_syscall(
```
stupendoussuperpowers, line 378: I believe this is `sendto` since it involves flags, `send` is `sendto` with flags -> 0. 

From the linux syscall table here - https://filippo.io/linux-syscall-table/, send is not a separate syscall and it's sendto instead. 

I am not sure how glibc handles this, whether send() just calls sendto() or if it calls write() as the man page mentions.

src/typemap/src/network_helpers.rs
```diff
@@ -0,0 +1,133 @@
...
+    let base: libc::socklen_t = 2;
```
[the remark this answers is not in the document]
stupendoussuperpowers: I think libc::socklen_t might be a type, we could import this explicitly if we want it to look cleaner.

stupendoussuperpowers (approved):
The changes look good to me, but need some clarity on:

1. send vs sendto
2. How we handle failures in unit-tests.

src/rawposix/src/net_calls.rs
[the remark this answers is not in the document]
Yaxuan-w: Yeah, my understanding is that in-memory pipes will handle such cases in their own grate?

tests/unit-tests/networking_tests/deterministic/tcp_connect_single.c
Yaxuan-w: Updated

src/threei/src/syscall_table.rs
Yaxuan-w: Great catch! Actually we should redirect `send` to `sendto` in glibc, since from man page:

```txt
the following call

           send(sockfd, buf, size, flags);

       is equivalent to

           sendto(sockfd, buf, size, flags, NULL, 0);
```

`sendmsg` is another syscall with signature `ssize_t sendmsg(int sockfd, const struct msghdr *msg, int flags);`

I removed `send_syscall` in rawposix and renamed 46 to sendmsg in glibc syscall table

PR #424, opened 2025-10-01
fix pthread

qianxichen233:
This PR addresses issues in the pthread implementation that were preventing thread-related tests from running.

The first issue was pretty trival: there is a `MAKE_SYSCALL` invocation in the `pthread_join` routine was not replacing the syscall number correctly. This has been fixed.

The second issue is more complex and still under consideration. In the `pthread_create` routine, an `unwind_buf` is used to simulate a `try-catch` block in C via setjmp. Since glibc is compiled as a 32-bit target, `unwind_buf` is only 4-byte aligned.
However, wasmtime’s setjmp implementation stores a 64-bit hash directly into the user-provided buffer. On a 64-bit host, dereferencing this as a u64 requires 8-byte alignment. This mismatch can lead to alignment issues.

The draft solution is to enforce 8-byte alignment on `unwind_buf` via an attribute. This resolves the pthread issue and allows tests to run again, but it may not be the ideal long-term solution.

I am considering an alternative approach that modifies the setjmp implementation itself to avoid relying on stricter alignment in user-provided buffers.

### DIFF
--- a/src/glibc/nptl/pthread_create.c
+++ b/src/glibc/nptl/pthread_create.c
@@ -471,7 +471,7 @@ start_thread (void *arg)
      the saved signal mask), so that is a false positive.  */
   DIAG_IGNORE_NEEDS_COMMENT (11, "-Wstringop-overflow=");
 #endif
-  not_first_call = setjmp ((struct __jmp_buf_tag *) &unwind_buf.cancel_jmp_buf);
+  not_first_call = setjmp ((struct __jmp_buf_tag *) unwind_buf.cancel_jmp_buf);
   DIAG_POP_NEEDS_COMMENT;
 
   /* No previous handlers.  NB: This must be done after setjmp since the
--- a/src/glibc/nptl/pthread_create.c
+++ b/src/glibc/nptl/pthread_create.c
@@ -40,6 +40,7 @@
 #include <clone3.h>
 #include <futex-internal.h>
 #include <syscall-template.h>
+#include <lind_syscall_num.h>
 
 #include <shlib-compat.h>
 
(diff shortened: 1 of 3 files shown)


src/glibc/nptl/pthread_create.c
```diff
@@ -459,7 +460,7 @@ start_thread (void *arg)
 
   /* This is where the try/finally block should be created.  For
      compilers without that support we do use setjmp.  */
-  struct pthread_unwind_buf unwind_buf;
+  struct pthread_unwind_buf __attribute__((aligned(8))) unwind_buf;
```
rennergade, line 463: add comment explaining this change

src/glibc/sysdeps/x86/bits/setjmp.h
```diff
@@ -27,13 +27,13 @@
 
 #ifndef _ASM
 
-# if __WORDSIZE == 64
+// # if __WORDSIZE == 64
```
rennergade, line 30: lets not commit commented code, add a comment here that this was removed and why

rennergade:
> This PR addresses issues in the pthread implementation that were preventing thread-related tests from running.
> 
> The first issue was pretty trival: there is a `MAKE_SYSCALL` invocation in the `pthread_join` routine was not replacing the syscall number correctly. This has been fixed.
> 
> The second issue is more complex and still under consideration. In the `pthread_create` routine, an `unwind_buf` is used to simulate a `try-catch` block in C via setjmp. Since glibc is compiled as a 32-bit target, `unwind_buf` is only 4-byte aligned. However, wasmtime’s setjmp implementation stores a 64-bit hash directly into the user-provided buffer. On a 64-bit host, dereferencing this as a u64 requires 8-byte alignment. This mismatch can lead to alignment issues.
> 
> The draft solution is to enforce 8-byte alignment on `unwind_buf` via an attribute. This resolves the pthread issue and allows tests to run again, but it may not be the ideal long-term solution.\

Great, thanks for adding this and the detailed explanation.

> 
> I am considering an alternative approach that modifies the setjmp implementation itself to avoid relying on stricter alignment in user-provided buffers.

This does seem like a bug that should be fixed before production but isn't blocking our path to a paper. My opinion is that we fix up and merge what you have here and post the stricter alignment project as a future issue.

PR #426, opened 2025-10-01
[Refactor] Threei integration patch: conflict fixes, and cageid and unused arg panic enforcement

Yaxuan-w:
This patch addresses issues and follow-ups after merging the previous PR for the threei integration. It consists of three main parts:

1. Missing/incorrect code after merge: Restore code and imports that were either lost or incorrectly resolved during conflict resolution in the last merge.
2. Cageid and unused arg panic enforcement: As discussed in the previous https://github.com/Lind-Project/lind-wasm/pull/387#discussion_r2392658967, unused cage IDs should now trigger a panic. At the time we made this decision, only the filesystem calls were pending; this patch updates system calls and network calls accordingly

### DIFF
--- a/src/rawposix/src/net_calls.rs
+++ b/src/rawposix/src/net_calls.rs
@@ -1452,8 +1452,6 @@ pub fn recvfrom_syscall(
             let errno = get_errno();
             return handle_errno(errno, "recvfrom");
         }
-
-        return ret;
     }
     // Case 2: both non-NULL → caller wants src_addr + addrlen filled
     else if !(addr_nullity || addrlen_nullity) {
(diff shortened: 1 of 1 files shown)


rennergade:
@Yaxuan-w can you split this into 2 PRs, one with the docker/testing changes and one with the platform changes

Yaxuan-w:
> one with the docker/testing changes

These changes are only merging from main branch. I can revert those changes, and leave them to be solved in final merge, if this's clearer for you.

rennergade:
Ok that makes sense

rennergade (approved):
I really think all the panics! should have a wrapper function or a macro but not sure if the plan is to do that in a later PR. Otherwise this is fine.

PR #427, opened 2025-10-01
Add shared memory syscalls to 3i

Yaxuan-w:
This PR adds the implementation of shared memory–related system calls. During testing, I found that shared memory was not working even on the original main branch. The initial issue was that the dispatcher on main branch was missing the shmctl syscall entirely. After adding it, the test files still failed. The current expansion also triggers errors in the 3i codebase, but this version compiles successfully and provides a baseline for further debugging.


rennergade (approved):
I think expanding the top line comments for each of the syscalls to be a bit more descriptive (ie recipe type layout) would be helpful since this is our best chance to actually make this better. I know I wrote this code originally years ago so if you need help with that let me know.

Otherwise, I think this looks fine.

ssannkkallpp (approved):

PR #433, opened 2025-10-06
tests: changing stream buffering type to line-buffer

drapl0n:
Updated `wasmtimereport.py`'s stream buffering type to line-buffering. This prevents output duplication while executing `mmap_shared` test natively using the automated script.


DyanB (approved):
Looks good!

PR #439, opened 2025-10-07
Make `disable_signals` opt-in, not default

### LINKED ISSUE #438: disable_signal feature incorrectly propogates for multi-processing
It seems like the signal_disable feature is being turned on by default in the multiprocess crate which is incorrect. It should only be enabled when enabled from the base build.

J7-7-7:
This fixes #438 and `disable_signals` is enabled when enabled from the base build.

### DIFF
--- a/skip_test_cases.txt
+++ b/skip_test_cases.txt
@@ -63,7 +63,6 @@ process_tests/deterministic/sem_forks.c
 process_tests/deterministic/setsid.c
 process_tests/deterministic/sigalrm.c
 process_tests/deterministic/sigchld.c
-process_tests/deterministic/signal-fork.c
 process_tests/deterministic/signal-simple.c
 process_tests/deterministic/sigpipe.c
 process_tests/deterministic/sigprocmask.c


src/wasmtime/Cargo.toml
```diff
@@ -55,13 +55,13 @@ wasmtime-wasi = { workspace = true, default-features = true, optional = true }
 wasmtime-wasi-nn = { workspace = true, optional = true }
 wasmtime-wasi-threads = { workspace = true, optional = true }
 wasmtime-lind-3i-vmctx = { path = "crates/lind-3i-vmctx" }
-wasmtime-lind-multi-process = { path = "crates/lind-multi-process", optional = false, features = ["disable_signals"]}
+wasmtime-lind-multi-process = { path = "crates/lind-multi-process", optional = false }
```
rennergade, line 58: doesn't this just remove the feature entirely?

J7-7-7: You can run it like this, cargo build --manifest-path src/wasmtime/Cargo.toml -p wasmtime-lind-multi-process --features disable_signals

PR #441, opened 2025-10-08
migrate wait syscall into lind-glibc

qianxichen233:
This PR deletes wait syscall in rawposix and have wait syscall in lind-glibc calling waitpid directly. Previously we are using same syscall number for both wait and waitpid, which could cause huge problem since both `wait` and `waitpid` would only go into the implementation of `wait_syscall` in rawposix, which means `waitpid` will have its argument misaligned and missed


rennergade:
Needs to pass linter, please run cargo fmt

Also might be a good opportunity to add a simple wait() test.

rennergade (changes requested):
Left some feedback

PR #465, opened 2025-10-14
Vmmap unit tests

### LINKED ISSUE #400: Add comprehensive unit tests for vmmap
As discussed on Monday, we're going to opt to deprecate the rawposix inner test suite but switch to unit tests for the individual modules (fdtables, vmmap, etc)

@Yaxuan-w mentioned she found a few bugs in the vmmap so this would be super helpful. Would be great to mention those situations here so we can add tests for them.

I think using the normal guidelines for unit tests in Rust here is appropriate: https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html

Shounak-Ghosh:
This PR creates unit tests for vmmap.rs and closes #400. Documentation of the unit tests and the key findings have been added as part of the cage lib README.

These are the results when running the tests:

failures:
    memory::vmmap::tests::test_change_prot_end_of_region
    memory::vmmap::tests::test_change_prot_exact_boundaries
    memory::vmmap::tests::test_change_prot_middle_of_region
    memory::vmmap::tests::test_change_prot_multiple_times
    memory::vmmap::tests::test_change_prot_spanning_multiple_regions
    memory::vmmap::tests::test_change_prot_to_none
    memory::vmmap::tests::test_change_prot_to_same_value
    memory::vmmap::tests::test_find_map_space_with_hint_alignment_in_pages

test result: FAILED. 14 passed; 8 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

### DIFF
--- a/src/cage/src/memory/vmmap.rs
+++ b/src/cage/src/memory/vmmap.rs
@@ -1239,7 +1239,10 @@ mod tests {
         assert_eq!(entry.prot, PROT_READ | PROT_WRITE);
 
         let count = vmmap.entries.overlapping(ie(100, 110)).count();
-        assert_eq!(count, 1, "Region should remain as a single entry (no fragmentation)");
+        assert_eq!(
+            count, 1,
+            "Region should remain as a single entry (no fragmentation)"
+        );
     }
 
     /// Test: Change protection on overlapping boundary


src/cage/README.md
```diff
@@ -0,0 +1,55 @@
...
+These tests clarify parameter expectations and search behavior for finding available address space.
+
+**Key Insight**: All parameters to `find_map_space_with_hint` are in PAGE UNITS, not byte addresses. This includes:
+- `hint`: starting page number for search
+- `npages`: number of pages needed
+- `pages_per_map`: alignment requirement in pages
+- Return value: interval of page numbers
+
+
+## Important Clarifications from Unit Tests
```
rennergade: this section is a bit weirdly phrased. we do want to know how this is structured but its phrased like we just found this out?

perhaps cross-reference the docs that talks about memory and make sure its consistent

Issue #435, opened 2025-10-06
Move Address Translation from RawPOSIX to libc


rennergade:
We'd like to translate 32 bit pointers to 64-bit pointers in libc before they're sent to 3i. This should help us become more platform agnostic as well as streamline the rawposix interface.

1. One way to do this would be simply to call into the vmmap's existing address translate infrastructure from glibc before we call MAKE_SYSCALL. However this would require a either a 2nd host function and extra overhead, or added to occur immediatey before MAKE_SYSCALL is triggered on the host side. 

Pros: less clobbering of libc, less engineering effort, doesn't expose base pointer to user space, doesnt require grates to do explicit vmmap checks

Cons: either more overhead or slightly more messy MAKE_SYSCALL interface

2. The other way to do this would require storing the base pointer in libc and adding conversion mechanisms for all pointers within libc, we'd later still check the vmmap for validity in RawPOSIX or from grates

Pros: clean 3i/MAKE_SYSCALL interface, low overhead
Cons: exposes base pointer to user space, requires grates to implicitly call to vmmap if they want to validate addresses

rennergade:
I just did some refresher on WASM's SFI model and I think that exposing the base address like this is fine but we probably do need to justify this security wise in both the docs and the paper since I'm sure it will raise some eyebrows/needs more than a handwave.

On another note, I may have missed this before but then the glibc were maintaining is supposed to be WASM specific?

rennergade:
Plan that Sankalp and I discussed:

1. Create libc address translation file. Add base_address variable,  function get_base_addr(), and translate_ptr_to_host() which adds base_address to any pointer.

2. Route get_base_addr() as a host function that is called on libc startup. This probably needs to be triggered in the crt before main. @qianxichen233 would appreciate any ideas here on where to place this function.

3. Wrap all pointers in translate_ptr_to_host() before MAKE_SYSCALL in their individual stubs.

rennergade:
This functionality will also need to be simultaneously removed from the vmmap.

rennergade:
@qianxichen233 This PR will fix the NULL pointer conversion problems as well. Do you have a list of tests that currently fail to that so we can enable them here?

qianxichen233:
> [@qianxichen233](https://github.com/qianxichen233) This PR will fix the NULL pointer conversion problems as well. Do you have a list of tests that currently fail to that so we can enable them here?

[...]

PR #469, opened 2025-10-15
Export base memory pointer from Lind and perform address translation in glibc

ssannkkallpp:
This PR is addressing the issue mentioned in https://github.com/Lind-Project/lind-wasm/issues/435.

**The changes include:**

- Addition of TRANSLATE_GUEST_POINTER_TO_HOST macro in glibc via addr_translation header file/c file
- Export lind-get-memory-base and lind-get-cage-id functions from lib.rs in Lind
- Wrap pointer arguments with TRANSLATE_GUEST_POINTER_TO_HOST macro before sending it through MAKE_SYSCALL
- Remove calls to translate_vmmap_addr() in typemap crate

### DIFF
--- a/src/rawposix/src/fs_calls.rs
+++ b/src/rawposix/src/fs_calls.rs
@@ -3572,12 +3572,12 @@ pub fn ioctl_syscall(
 
     // Only support FIONBIO and FIOASYNC. Return error for unsupported requests.
     if req_arg != FIONBIO as u64 && req_arg != FIOASYNC as u64 {
-        lind_debug_panic("Error in syscall: ioctl - EINVAL: Unsupported ioctl request");
+        lind_debug_panic("Lind unsupported ioctl request - FIONBIO and FIOASYNC only");
     }
 
     let wrappedvfd = fdtables::translate_virtual_fd(cageid, vfd_arg);
     if wrappedvfd.is_err() {
-        lind_debug_panic("Error in syscall: ioctl - EBADF: Bad File Descriptor");
+        return syscall_error(Errno::EBADF, "ioctl", "Bad File Descriptor");
     }
 
     let vfd = wrappedvfd.unwrap();


src/typemap/src/datatype_conversion.rs
```diff
@@ -305,9 +298,7 @@ pub fn sc_convert_addr_to_epollevent<'a>(
         }
     }
 
-    let cage = get_cage(cageid).unwrap();
-    let addr = translate_vmmap_addr(&cage, arg).unwrap();
-    let pointer = addr as *mut EpollEvent;
+    let pointer = arg as *mut EpollEvent;
```
Yaxuan-w: Could we directly pass epoll related data structure to kernel as ptr directly? I mean some ops like `arg as *mut libc::epoll_event`

src/glibc/sysdeps/unix/sysv/linux/epoll_ctl.c
```diff
@@ -20,13 +20,14 @@
 #include <sysdep.h>
 #include <syscall-template.h>
 #include <lind_syscall_num.h>
+#include <addr_translation.h>
 
 libc_hidden_proto (epoll_ctl)
 
 int
 epoll_ctl (int __epfd, int __op, int __fd,
```
rennergade: epoll_event is a weird struct here, we probably need to convert that here manually

src/glibc/sysdeps/unix/sysv/linux/ioctl.c
```diff
@@ -31,7 +32,7 @@ __ioctl (int fd, unsigned long int request, ...)
   void *arg = va_arg (args, void *);
   va_end (args);
 
-        return MAKE_SYSCALL(IOCTL_SYSCALL, "syscall|ioctl", (uint64_t) fd, (uint64_t) request, (uint64_t) arg, NOTUSED, NOTUSED, NOTUSED);
+        return MAKE_SYSCALL(IOCTL_SYSCALL, "syscall|ioctl", (uint64_t) fd, (uint64_t) request, (uint64_t) TRANSLATE_GUEST_POINTER_TO_HOST(arg), NOTUSED, NOTUSED, NOTUSED);
```
rennergade: ioctl is extremely weird and uses var args, i dont think this is handling that correclty

src/glibc/sysdeps/unix/sysv/linux/writev.c
```diff
@@ -26,7 +26,13 @@ ssize_t
 __writev (int fd, const struct iovec *iov, int iovcnt)
 {
   // Dennis Edit
-  return MAKE_SYSCALL(WRITEV_SYSCALL, "syscall|writev", (uint64_t) fd, (uint64_t)(uintptr_t) iov, (uint64_t) iovcnt, NOTUSED, NOTUSED, NOTUSED);
+  struct iovec *host_iov = malloc(iovcnt * sizeof(struct iovec));
```
rennergade: if we malloc a struct like this we'll need to free as well.

src/glibc/sysdeps/unix/sysv/linux/i386/libc_sigaction.c
```diff
@@ -16,111 +16,113 @@
...
+     // we do the manual translation between glibc sigaction struct and rawposix sigaction struct here
+     struct rawposix_sigaction rawposix_act, rawposix_oact;
+     // check for NULL pointer
+     if (act)
+     {
+       rawposix_act.handler = act->sa_handler;
+       rawposix_act.sa_mask = act->sa_mask.__val[0];
+       rawposix_act.sa_flags = act->sa_flags;
+     }
+     int retval = MAKE_SYSCALL(SIGACTION_SYSCALL, "syscall|sigaction", (uint64_t) sig, (uint64_t) TRANSLATE_GUEST_POINTER_TO_HOST(act ? &rawposix_act : NULL), (uint64_t) TRANSLATE_GUEST_POINTER_TO_HOST(oact ? &rawposix_oact : NULL), NOTUSED, NOTUSED, NOTUSED);
```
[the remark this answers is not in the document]
qianxichen233: I think this should be fine? This struct is handled in rawposix and it is basically the same with how we handle other structs. even if it does not go into kernel, we still need to translate the address somewhere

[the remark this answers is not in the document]
qianxichen233: also I think this file has an extra space at the beginning of every line for some reason

src/glibc/sysdeps/unix/sysv/linux/setsockopt.c
```diff
@@ -90,7 +90,9 @@ setsockopt32 (int fd, int level, int optname, const void *optval,
 int
 __setsockopt (int fd, int level, int optname, const void *optval, socklen_t len)
 {
-  int r = setsockopt_syscall (fd, level, optname, optval, len);
+  /* Translate only the guest-provided pointer on the direct path. */
+  const void *translated_optval = TRANSLATE_GUEST_POINTER_TO_HOST(optval);
+  int r = setsockopt_syscall (fd, level, optname, translated_optval, len);
```
qianxichen233: This is not the root level of the syscall invocation, why we are doing the address translation here instead of the place where `MAKE_SYSCALL` is used?

src/glibc/sysdeps/unix/sysv/linux/i386/libc_sigaction.c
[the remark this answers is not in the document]
rennergade: > I think this should be fine? This struct is handled in rawposix and it is basically the same with how we handle other structs. even if it does not go into kernel, we still need to translate the address somewhere

ok so these are currently translated to 64bit host pointers in rawposix?

rennergade:
> I think we should add address checking on rawposix side as well? Now we are directly using the address passed from userspace, which is quite dangerous

We'll still be checking the VMMap once in trusted space. We're now just converting from 32 to 64 bit host pointers in libc. Realistically the kernel still has checks on any pointer supplied to it, so at worst case were not doing anything worse than a scenario where a user calls a syscall with a bogus pointer.

src/glibc/sysdeps/unix/sysv/linux/i386/libc_sigaction.c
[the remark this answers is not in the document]
ssannkkallpp: Done.

src/glibc/sysdeps/unix/sysv/linux/munmap.c
```diff
@@ -4,17 +4,18 @@
...
   // return 0;
 }
 
 int munmap(void *addr, size_t len)
 {
-  return MAKE_SYSCALL(MUNMAP_SYSCALL, "syscall|munmap", (uint64_t)(uintptr_t) addr, (uint64_t) len, NOTUSED, NOTUSED, NOTUSED, NOTUSED);
+  return MAKE_SYSCALL(MUNMAP_SYSCALL, "syscall|munmap", (uint64_t) TRANSLATE_GUEST_POINTER_TO_HOST(addr), (uint64_t) len, NOTUSED, NOTUSED, NOTUSED, NOTUSED);
 }
 
 weak_alias(__GI___munmap, __munmap)
```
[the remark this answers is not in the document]
ssannkkallpp: Done.

src/glibc/sysdeps/unix/sysv/linux/writev.c
```diff
@@ -26,7 +26,13 @@ ssize_t
 __writev (int fd, const struct iovec *iov, int iovcnt)
 {
   // Dennis Edit
```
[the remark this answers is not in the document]
ssannkkallpp: Done.

src/glibc/sysdeps/unix/sysv/linux/ioctl.c
rennergade: Ok so from discussion we just want to handle FIONBIO and FIOASYNC which both need pointer translation. 

lets assert and fail for other cases so we know they come up

qianxichen233:
I also believe some syscalls are missing since they are not using `MAKE_SYSCALL` but a slightly different macro. For example, `futex` syscalls are implemented inside `glibc/nptl/futex-internal.c`, which uses `MAKE_RAW_SYSCALL` macro instead. There are also other variants of `MAKE_SYSCALL`, as you can check in `glibc/sysdeps/unix/syscall-template.h`, I am not sure if those variants are used somewhere, but it's worth to check as well

rennergade:
> > We mentioned in the last meeting that we'd like to deprecate the syscall strings in MAKE_SYSCALL ie ("syscall|bind"). This PR seems like a decent time to do that. Can we lump that in here?
> 
> This will also influence wasmtime and 3i. I'd suggest to do that after grates' PR being merged or in a separate PR

ok agreed

src/glibc/elf/sotruss-lib.c
```diff
@@ -77,7 +77,7 @@ init (void)
      yes, whether we should print into a file.  */
   const char *which_process = getenv ("SOTRUSS_WHICH");
 
-  // Dennis Edit
```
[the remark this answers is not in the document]
rennergade: Yeah

src/glibc/nptl/pthread_create.c
```diff
@@ -40,7 +40,11 @@
 #include <clone3.h>
 #include <futex-internal.h>
 #include <syscall-template.h>
+<<<<<<< HEAD
```
rennergade: still merge conflict

ssannkkallpp: Fixed.

scripts/make_glibc_and_sysroot.sh
```diff
@@ -100,6 +100,12 @@ $CC $CFLAGS $WARNINGS $EXTRA_FLAGS \
     -o $BUILD/lind_syscall.o \
     -c $GLIBC/lind_syscall/lind_syscall.c
 
+# Compile address translation module
+$CC $CFLAGS $WARNINGS $EXTRA_FLAGS \
```
rennergade: i think this needs a docs update too

src/glibc/sysdeps/unix/sysv/linux/ioctl.c
```diff
@@ -22,20 +22,40 @@
...
   va_start (args, request);
-  void *arg = va_arg (args, void *);
+
+  /* Use unsigned long to safely capture either pointer or integer values */
+  unsigned long raw = va_arg (args, unsigned long);
   va_end (args);
 
-        return MAKE_SYSCALL(IOCTL_SYSCALL, "syscall|ioctl", (uint64_t) fd, (uint64_t) request, (uint64_t) arg, NOTUSED, NOTUSED, NOTUSED);
+  /* Only support FIONBIO and FIOASYNC.  Fail fast otherwise. */
+  assert (request == FIONBIO || request == FIOASYNC);
```
rennergade: will this assert give us enough debug info if it fails?

src/glibc/lind_syscall/addr_translation.c
```diff
@@ -0,0 +1,22 @@
...
+__lind_init_addr_translation (void)
+{
+  if (__lind_base != 0ULL && __lind_cageid != 0ULL)
+    {
+      return; // idempotent
+    }
+  // Query the host for the base address of this cage's linear memory
+  __lind_base = (uint64_t) __imported_lind_get_memory_base ();
+  // Query the host for the cage id (pid) for this instance
+  __lind_cageid = (uint64_t) __imported_lind_get_cage_id ();
```
qianxichen233: Should we use the syscall routine (i.e. getpid) instead of an independent routine to get the cage id for 3i compatibility and code consistency? @Yaxuan-w

qianxichen233: We definitely need `__lind_cageid` stored in glibc for efficiency and/or other design considerations. But I mean `__imported_lind_get_cage_id` introduces a separate routine to retrieve the cage ID instead of using the normal syscall path.
We should document the reasoning behind choosing `__imported_lind_get_cage_id` over `getpid_syscall` (or vice versa). Additionally, note that `__imported_lind_get_cage_id` returns the pid stored in wasmtime, rather than the pid from rawposix, which is another key behavioral difference from `getpid_syscall`.

rennergade: This seems like a good discussion. @Yaxuan-w can you clarify what we want to store the cageid in libc for? May be easier to understand that way.

@qianxichen233 I think one reason to have this here vs harnessing getpid() is that getpid could have a grate attached to return some sort of masked value, while this would always return the correct cageid.

Yaxuan-w: > can you clarify what we want to store the cageid in libc for? May be easier to understand that way.

The cageid. Could we store the cageid as a symbol in glibc? (Only call wasmtime once)

JustinCappos:
The point is that we need it for some of the system calls we make and
having it as something that is set once upon initialization (or reset on
fork) instead of reading it in an expensive manner repeatedly.

glibc could technically set it and store it.  However, setting it upon cage
load is even better.  This variable will be used a lot.

scripts/make_glibc_and_sysroot.sh
rennergade: https://lind-project.github.io/lind-wasm/contribute/toolchain/

Yaxuan-w (approved):

src/glibc/sysdeps/nptl/lowlevellock-futex.h
```diff
@@ -65,62 +66,85 @@
...
 
-# define lll_futex_timed_wait(futexp, val, timeout, private)     \
-  lll_futex_syscall (4, futexp,                                 \
-		     __lll_private_flag (FUTEX_WAIT, private),  \
-		     val, timeout)
+# define lll_futex_timed_wait(futexp, val, timeout, private)            \
+  ({                                                                    \
+  uint64_t __translated_futexp = TRANSLATE_GUEST_POINTER_TO_HOST(futexp); \
+  uint64_t __translated_timeout = TRANSLATE_GUEST_POINTER_TO_HOST(timeout); \
+  lll_futex_syscall (4, __translated_futexp,                          \
```
qianxichen233: this is not our syscall entry, I do not think we need the address translation here?

src/glibc/posix/fork.c
```diff
@@ -83,6 +84,10 @@ __libc_fork (void)
   if (pid == 0)
     {
       fork_system_setup ();
+      // Reinitialize address translation for child processes
+      __lind_base = 0ULL;
+      __lind_cageid = 0ULL;
+      __lind_init_addr_translation ();
```
qianxichen233: this would be better to move to `src/glibc/sysdeps/unix/sysv/linux/i386/clone3.c`

src/glibc/sysdeps/unix/sysv/linux/fcntl.c
qianxichen233: could you remove the formatter change for this file

src/glibc/sysdeps/unix/sysv/linux/ioctl.c
qianxichen233: currently no I think, the glibc's error message seems to be a little buggy and the error message will not show up

qianxichen233: btw, why we are doing assert here in glibc? I think this is likely to be a soft panic case that should appear in host?

src/glibc/sysdeps/nptl/lowlevellock-futex.h
rennergade: these pass down to the macros in futex-internal right? Are we translating twice somehow?

src/glibc/posix/fork.c
ssannkkallpp: Done.

src/glibc/sysdeps/unix/sysv/linux/i386/clone3.c
```diff
@@ -3,14 +3,32 @@
...
+
+  // Reinitialize address translation for child processes
+  if (pid == 0) {
+    __lind_base = 0ULL;
+    __lind_cageid = 0ULL;
+    __lind_init_addr_translation ();
+  }
+
+  // Execute child function if in child process
+  if (pid == 0 && func != NULL) {
```
qianxichen233: can we merge these two `if`?

src/rawposix/src/fs_calls.rs
```diff
@@ -3565,6 +3565,11 @@ pub fn ioctl_syscall(
             "ioctl_syscall"
         );
     }
+    
+    // Only support FIONBIO and FIOASYNC. Return error for unsupported requests.
+    if req_arg != FIONBIO as u64 && req_arg != FIOASYNC as u64 {
+        return syscall_error(Errno::EINVAL, "ioctl", "Unsupported ioctl request");
```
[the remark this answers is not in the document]
ssannkkallpp: Updated.

PR #477, opened 2025-10-19
Rename lind file system root

### LINKED ISSUE #369: Rename lind file system root
Currently, the lind file system root is set to a temporary path (`src/RawPOSIX/tmp`) that I originally introduced. Since the old RawPOSIX has now been replaced by the new rawposix, continuing to use this path results in two rawposix folders, which is confusing.

I propose renaming it to `src/lind_fs_root` and updating the definition in [lind_platform_const.rs](https://github.com/Lind-Project/lind-wasm/blob/main/src/sysdefs/src/constants/lind_platform_const.rs#L12C30-L12C67) accordingly. I’d like to hear everyone’s thoughts on this change (@rennergade @qianxichen233 @yzhang71).

celinehoang177:
This PR simplifies the Lind file system root path by moving it from `src/RawPOSIX/tmp` to `src/tmp` (Fixes #369)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+LIND_ROOT ?= src/tmp
 
 .PHONY: build 
 build: sysroot wasmtime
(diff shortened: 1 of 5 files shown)


rennergade:
@celinehoang177 you have some conflicts here with recent merges you need to resolve

rennergade:
@celinehoang177 this doesn't seem to compile, can you check that. Also needs to be fixed for the linter.

Yaxuan-w (approved):

Issue #436, opened 2025-10-06
Customize/Optimize development container


rennergade:
As discussed in the meeting today there's some customizations we'd like to add to the dev container for easier out of the box workflow.

I know one of the things is building wasmtime in debug mode instead of release mode.

@Yaxuan-w @qianxichen233 would love you two to comment more things you'd like to see in the environment for this issue and @J7-7-7 will set that up.

rennergade:
> Another feature that is probably be convenient for devcontainer is to have the ability to `make` the project anywhere in the directory. For example, if I am working on `wasmtime` and need to frequently modify and re-compile `wasmtime`, currently I have to manually switch to a different terminal (or `cd` into root directory to compile). It would be convenient to have an utility binary in `PATH` that allows us to do something like `lindtool compile_wasmtime` at any directory, so that the we do not need to frequently switch between terminals

We should be able to do that with the updated container that went up today due to the relative paths PR by just running the commands the Makefile calls. However i tested rebuilding sysroot from somewhere else and there was an error, I think some of the PATHs still need to be fixed. Will put that up as a separate issue.

rennergade:
@J7-7-7 the apps they mentioned in the meeting to add are the ones from https://github.com/WebAssembly/wabt including wasm2wat

qianxichen233:
We also need strace in the devcontainer

PR #491, opened 2025-10-23
customize dev container

J7-7-7:
build wasmtime in debug mode instead of release mode, added wabt and strace. Reference issue #436


rennergade (commented):
I'm a bit confused. Will the devcontainer already be built in the debug state?

PR #514, opened 2025-11-04
Modifying HTML Output when tests pass for cleaner reports

DyanB:
Minor UI fix for cleaner HTML reports.
Successful tests don't print the full output.


rennergade:
Thanks!

PR #518, opened 2025-11-07
Logging: add lind_debug_panic macro

drapl0n:
Adding lind_debug_panic(), a macro wrapping rust's panic!.


rennergade:
This needs to pass the linter (use cargo fmt)

Yaxuan-w (approved):
Thanks for doing this! It's clear as example.

PR #519, opened 2025-11-07
Add tests for ftruncate, getdents, link, lseek, and unlink

AlluSudharshanRao:
link, lseek, and unlink


ChinmayShringi:
Looks good! Not sure if you want if conditions replaced with asserts

PR #522, opened 2025-11-10
readlink syscall fix to return symlink target exactly as is

DyanB:
The Bug:
The syscall inverts the behavior depending on whether the symlink contains an absolute or relative path:

For Absolute path symlink
Symlink created:  /home/lind/lind-wasm/src/RawPOSIX/tmp/testfiles/readlinkfile.txt
Native readlink returns: /home/lind/lind-wasm/src/RawPOSIX/tmp/testfiles/readlinkfile.txt
Lind readlink returns:   testfiles/readlinkfile.txt  (LIND_ROOT was stripped)

For Relative path symlink it prepends LIND_ROOT
Hence No matter which approach is used in the test suite, the behavior is inverted and tests fail.

I've made changes to return symlink as is; so that readlink passes tests in debug mode (tests will pass locally) and release mode (CI).

### DIFF
--- a/src/rawposix/src/fs_calls.rs
+++ b/src/rawposix/src/fs_calls.rs
@@ -1759,39 +1759,17 @@ pub fn readlink_syscall(
         let errno = get_errno();
         return handle_errno(errno, "readlink");
     }
-    // Convert the result from readlink to a Rust string
-    let libcbuf_str = unsafe { CStr::from_ptr(libc_buf.as_ptr() as *const c_char) }
-        .to_str()
-        .unwrap();
-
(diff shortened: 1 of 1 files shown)


Yaxuan-w:
I think the reason of failure is mostly caused by incorrect location of readlink.c. [In the test file](https://github.com/Lind-Project/lind-wasm/blob/main/tests/unit-tests/file_tests/deterministic/readlink.c), the lind-wasm user should never see `LIND_ROOT`, so we'll always get different output for this test. I think the solution here would be put readlink.c into non-deterministic folder, and don't change rawposix's implementation.

Issue #37, opened 2024-11-18
Fix locale


qianxichen233:
When fixing bash, I faced a lot of issues related to locale. Previously, locale is not working and I commented them out since I thought they are not important as locale is mainly used for setting up language specific context. But it turns out locale is used much more often than I thought. For example, isblank will look up locale to determine language specific blank character. I currently have to manually modify each function that uses locale, which is working fine, but it might probably be a better choice to just completely fix locale.

rennergade:
![Screenshot from 2024-11-20 16-03-22](https://github.com/user-attachments/assets/e2fba171-7ab4-4e6b-a868-22861be86adb)

From some brief research locale files seem to just be ASCII text, though they are compiled to that? https://stackoverflow.com/questions/65514890/glibcs-isalpha-function-and-the-en-us-utf-8-locale

I believe all we should have to do is compile the locale files and place them in the appropriate directory in the file system

rennergade:
If what I posted above is true it shouldnt take any actual changes, just fixing the build and setup processes. @robinyuan1002 and @yzhang71 can you try that out.

qianxichen233:
To reproduce the issue, I created a simple program that uses `isblank` function, which uses locale stuff internally to determine if the character is considered as blank under current language. This program is not working as expected and `isblank` will always return false.

I digged into this a little bit. And found something interesting (actually by accident). So `isblank` will try to query `__libc_tsd_CTYPE_B`. And `__libc_tsd_CTYPE_B` is something gets set up under `__ctype_init` function. In crt1.c, we did not call the function to initialize ctype, so `__libc_tsd_CTYPE_B` is always empty previously. This is fixed by just calling `__ctype_init` in _start function in crt1.c.
However, `isblank` is still not working after adding `__ctype_init`. In `__ctype_init`, `__libc_tsd_CTYPE_B` is set to `_nl_current_LC_CTYPE`, which sets from `_nl_global_locale` (from `lc-ctype.c` and `localeinfo.h`) by default. The `_nl_global_locale` also has a default value, sets under `global-locale.c` file. One of the entry in `_nl_global_locale` is set to `_nl_C_LC_CTYPE_class`, which should hold all the information about the default character sets. The `_nl_C_LC_CTYPE_class` is a static array defined in `C-ctype.c`.
[...]

PR #534, opened 2025-11-19
fix locale

qianxichen233:
This is a legacy PR splitted from my old big various fix. Seems like I forgot to raise this PR at that time for some reason, and I just realized that this hasn't been merged into main when I encountered some locale issue

### DIFF
--- a/src/glibc/locale/global-locale.c
+++ b/src/glibc/locale/global-locale.c
@@ -20,6 +20,8 @@
 #include "localeinfo.h"
 
 #define DEFINE_CATEGORY(category, category_name, items, a) \
+// lind-wasm: these struct needs not to be weak symbols to work correctly in wasm for some reason
+// TODO: should go through here again when locale is fully fixed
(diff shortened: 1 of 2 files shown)


src/glibc/locale/global-locale.c
```diff
@@ -20,17 +20,14 @@
 #include "localeinfo.h"
 
 #define DEFINE_CATEGORY(category, category_name, items, a) \
-extern struct __locale_data _nl_C_##category; weak_extern (_nl_C_##category)
+extern struct __locale_data _nl_C_##category; /*weak_extern (_nl_C_##category)*/
```
Yaxuan-w: Is there a reason that we mark this as comment?

qianxichen233: yeah, see discussion at https://github.com/Lind-Project/lind-wasm/issues/37

rennergade:
Closing this for now and just fixing ctype init to support isalpha. Will revisit locale when we need to.

Issue #485, opened 2025-10-20
Unify usage of MAKE_SYSCALL wrappers in glibc


Yaxuan-w:
Currently, different system calls in our glibc modifications use different syscall wrappers. Most syscalls use the standard `MAKE_SYSCALL`, but a few (e.g., futex) use `MAKE_SYSCALL_RAW`.

To simplify future development and code search (e.g., when running `grep -r "MAKE_SYSCALL(SYSCALL_WRITE" .` to locate modified syscalls), we should standardize the usage of syscall wrappers across all files.

Yaxuan-w:
Per discussion on slack, my idea is to add a new flag to the MAKE_SYSCALL macro so we can unify the two different wrappers. To minimize code changes, we could introduce a new macro that automatically detects calls without the flag and adds it implicitly.

Issue #509, opened 2025-10-31
grate: Rename and extend glibc make_syscall API for cage-aware syscall propagation


Yaxuan-w:
Currently, our implementation supports cross-cage/grate calls, but the existing glibc `make_syscall` interface does not include the source cageid in its arguments. As a result, when a call jumps to a third or more cage or grate, it becomes impossible to determine the originating Cage ID. To support future grate functionality and more complex inter-cage routing, the API needs to be refactored.

After discussing with @JustinCappos, the changes will look like:

1. Rename `MAKE_SYSCALL` -> `MAKE_TRANSITION`
  - Represents a traditional syscall **issued by a cage** (i.e., without explicit cageid fields). Calls through this API will attach the caller’s cageid automatically inside 3i before propagation.
2. Add new macro `MAKE_THREEI`
  - Represents syscalls **issued by a grate**. Exposed to users for grate-level syscall invocation. This will allow full control of both source and target cageids, as well as per-argument provenance.

`MAKE_THREEI` in glibc looks like:

```C
MAKE_THREEI(
    syscall_num,
    self_cageid,
    target_cageid,
    arg1, arg1_cageid,
    arg2, arg2_cageid,
    arg3, arg3_cageid,
    arg4, arg4_cageid,
    arg5, arg5_cageid,
    arg6, arg6_cageid
)
```

PR #535, opened 2025-11-22
Refactor and unify glibc 3i/legacy syscall path

Yaxuan-w:
This PR resolves issue #485 and issue #509

It introduces a set of structural cleanups and unifications across the syscall/threei/glibc layer. The major changes include:

### 1. Refactor of the legacy make_syscall path

- Replaced the previous `MAKE_SYSCALL` with a macro-based `MAKE_LEGACY_SYSCALL` wrapper used only inside the glibc legacy syscall implementation.
- Added a `raw_flag` to control whether errno translation should occur.
- Unified futex handling by removing the previous parameter-dependent macro expansion and switching to a single consistent `MAKE_LEGACY_SYSCALL ` call.
- Ensured that `MAKE_LEGACY_SYSCALL ` is not exposed to grate authors to prevent accidental misuse.

### 2. Refactor of `lind_syscall` into `make_threei_syscall`

- Introduced `make_threei_syscall ` as the unified Wasmtime entry for both grate calls and legacy syscalls.
- `MAKE_LEGACY_SYSCALL ` now automatically fills `self_cageid`, `target_cageid`, and all `arg*_cageid` fields based on the current cage for future 3i/grate development.
- Established `make_threei_syscall ` as the canonical interface for all future grate-level interposition.

### 3. Consolidation of grate-facing API

- Refactored `register_handler` and `copy_data_between_cages`.
- Moved them into a single public header (`lind_syscall.h`) with `make_threei` which now serves as the user-facing API for grate authors.

### 4. Updated Wasmtime trampoline signatures

- Adjusted the host-side trampoline to match the new argument structure and the updated syscall ABI.

### 5. Updated grate tests and glibc scripts

- Tests and build scripts have been aligned with the new unified syscall and handler API.

### DIFF
--- a/src/glibc/sysdeps/nptl/lowlevellock-futex.h
+++ b/src/glibc/sysdeps/nptl/lowlevellock-futex.h
@@ -58,15 +58,6 @@
 # define __lll_private_flag(fl, private) \
   (((fl) | FUTEX_PRIVATE_FLAG) ^ (private))
 
-// # define lll_futex_syscall(nargs, futexp, op, ...)                      \
-//   ({                                                                    \
-//     long int __ret = MAKE_LEGACY_SYSCALL##nargs (FUTEX_SYSCALL, "syscall|futex", futexp, op, 	\
-// 				       __VA_ARGS__);                    \
-//     (__glibc_unlikely (INTERNAL_SYSCALL_ERROR_P (__ret))         	\
-//      ? -INTERNAL_SYSCALL_ERRNO (__ret) : 0);                     	\
-//   })
-
-
 /* For most of these macros, the return value is never really used.
   Nevertheless, the protocol is that each one returns a negated errno
   code for failure or zero for success.  (Note that the corresponding


src/glibc/io/unlinkat.c
```diff
@@ -27,6 +27,6 @@
 int
 unlinkat (int fd, const char *name, int flag)
 {
-  return MAKE_SYSCALL3(UNLINKAT_SYSCALL, "syscall|unlinkat", (uint64_t)fd, (uint64_t)name, (uint64_t)flag);
+  return MAKE_TRANDITION(UNLINKAT_SYSCALL, "syscall|unlinkat", (uint64_t)fd, (uint64_t)name, (uint64_t)flag, NOTUSED, NOTUSED, NOTUSED, WRAPPED_SYSCALL);
```
rennergade: I'm guessing TRANDITION is a typo of TRANSITION that got persisted?

Even so I'm unclear what MAKE_TRANSITION really means

rennergade (changes requested):
Will try to review fully tomorrow but had an immediate question

src/glibc/io/unlinkat.c
Yaxuan-w: Oh TRANSITION is a typo when I wrote the issue. I intended to say traditional.. there's still a typo anyway. Will fix it tomorrow

rennergade: I'm not sure what MAKE_TRADITIONAL really implies either.

src/glibc/lind_syscall/lind_syscall.c
```diff
@@ -1,39 +1,89 @@
...
+ *       The wrapper does *not* apply any errno translation.
+ *       The raw return value from the trampoline is returned as-is.
+ *
+ * This distinction is required because some syscalls—especially futex-related
+ * operations (e.g., lll_futex_wake, lll_futex_requeue, etc.) expect the
+ * trampoline to return raw -errno value and must not receive additional errno 
+ * post-processing at this layer. Other syscalls, however, rely on the standard 
+ * POSIX errno translation implemented here.
+ */
+int make_threei (unsigned int callnumber,
```
rennergade: Still nitpicking at names so apologies. Would it make sense to lengthen this to `make_threei_call`

Would make more sense to me since were making a call and not making threei

src/glibc/misc/fsync.c
```diff
@@ -18,12 +18,13 @@
...
 #include <unistd.h>
 #include <syscall-template.h>
+#include <lind_syscall_num.h>
 
 /* Make all changes done to FD actually appear on disk.  */
 int
 fsync (int fd)
 {
-	return MAKE_SYSCALL(162, "syscall|fsync", (uint64_t) fd, NOTUSED, NOTUSED, NOTUSED, NOTUSED, NOTUSED);
+	return MAKE_TRADITION(FSYNC_SYSCALL, "syscall|fsync", (uint64_t) fd, NOTUSED, NOTUSED, NOTUSED, NOTUSED, NOTUSED, WRAPPED_SYSCALL);
```
rennergade: More name nitpicking. I know you're just using code from before here but seems like a good opportunity for us to clarify. I never really understoood the Wrapped vs Raw difference. It seems just to be because of errno handling right? If thats true those names are really non-sensical to me. What if we changed it to a handle_errno flag or something like that?

stupendoussuperpowers:
Haven't had the chance to review the entire PR yet, but adding to the naming conversation - 

Is there a reason to have macros with verbose names like `MAKE_TRADITIONAL` or `MAKE_LEGACY_SYSCALL` instead of just sticking with `MAKE_SYSCALL`? From what I can tell within glibc there is only one kind of MACRO being used (`MAKE_TRADITIONAL`) and we are only exposing one kind of function to grate files (`make_threei`).

Yaxuan-w:
> Is there a reason to have macros with verbose names like MAKE_TRADITIONAL or MAKE_LEGACY_SYSCALL instead of just sticking with MAKE_SYSCALL? From what I can tell within glibc there is only one kind of MACRO being used (MAKE_TRADITIONAL) and we are only exposing one kind of function to grate files (make_threei).

@stupendoussuperpowers 
(1) We want to have a unified trampoline function when calling into wasmtime
(2) We want to attach the cageid in glibc for performance optimization (this can help us bypass data clone in wasmtime)

src/glibc/misc/fsync.c
rennergade: or like TRANSLATE_ERRNO is probably clearer

PR #571, opened 2025-12-14
3i: Refactor vmctx storage and grate call execution path

### LINKED ISSUE #538: 3i: panic caused by multiple grate context registrations in single grate
Thanks to @stupendoussuperpowers  for spotting this bug.

We discovered a panic `[3i|_add_global_grate] attempt to overwrite existing grate entry for grateid` that occurs when calling `register_handler` multiple times inside the same grate to register different handlers. The root cause is that we were [registering the grate context multiple times, rather than only updating the handler table](https://github.com/Lind-Project/lind-wasm/blob/main/src/threei/src/threei.rs#L223).

During the previous refactor of the grate execution flow, I changed the context-registration logic so that the grate context would be sent to 3i every time `register_handler` is invoked. This inadvertently caused repeated context registrations for the same grate.

My current fix is to add a cageid–based check inside 3i so that a grate context is only registered once per cage. 

Since this issue only manifests when the entire lind-wasm is running, it cannot be fully captured by existing 3i unit tests, even though we have tests inside 3i that appear to cover this scenario. As part of the fix PR, we will also need to add a dedicated test case in the [unit-test folder](https://github.com/Lind-Project/lind-wasm/blob/main/src/tests/grate-tests) to ensure this behavior is properly validated going forward.

Yaxuan-w:
In lind-wasm, execution needs to re-enter Wasmtime from outside the current instance or call stack. Two common cases are:

1. Process-like operations (fork / exec / exit): RawPOSIX may need to create, clone, or tear down Wasm instances and then return control back into Wasmtime. The cage performing these operations is not necessarily the same cage or grate that issued the syscall.
2. Grate calls: Grate calls require cross-module execution transfers, where control jumps from one cage into another cage or grate.

Both cases require an explicit way to recover the correct Wasmtime runtime state (store + instance) for an arbitrary `cage_id` pair. Relying on an implicit “current instance” is not sufficient.

## Implementation

This PR moves a low-level runtime-state lookup mechanism based on Wasmtime’s `VMContext` pointer from 3i to wasmtime:

- lind-wasm guarantees that each grate call executes in its own store / instance pair.
- During module instantiation, the `VMContext` pointer is extracted and stored in a global table keyed by `cage_id`.
- When control needs to transfer to another cage or grate, lind-3i retrieves the corresponding VMContext pointer and uses Wasmtime-internal mechanisms to recover the correct store and instance.

The table stores raw pointers rather than typed handles. 

### VMContext storage refactoring and execution semantics

This PR resolves #538. Wasmtime manages a global pool indexed by `cage_id` and will be initialized at the start of lind-wasm execution. Each cage owns a dedicated FIFO queue of `VmCtxWrapper` objects. At instance execution startup, a fixed number of Wasmtime instances (currently 10) are created for each cage and Wasmtime stores `VMContext` pointers per `cage_id `. One instance is fully initialized using `instantiate_with_lind`, which performs lind-specific memory setup. The remaining instances are created using `instantiate_with_lind_thread`, and attach to the same linear memory as the primary instance, serving as backup execution contexts. When a grate call is executed, the `VMContext` pointer is retrieved `VMContext` are preallocated in pools to enable concurrent request handling within a single cage or grate. The `VMContext` will be pushed back to the pool after grate call execution.

### Wasmtime instantiation extensions

To minimize divergence from upstream Wasmtime code:

- `new_started_impl` remains unchanged.
- `new_started_impl_with_lind` is used for cage creation and performs lind-specific memory initialization.
- A new helper, `new_started_impl_with_lind_thread`, is introduced for thread creation. Threads do not require cage-level memory initialization, but they do require returning an InstanceId so the VMContext can be registered.

### 3i runtime-agnostic grate dispatch

This PR clarifies the execution boundary between 3i and runtimes:

- Each cage is associated with a runtime ID.
- Each runtime registers a trampoline function that knows how to execute grate calls in that runtime.
- `_call_grate_func` resolves the runtime and delegates execution entirely to the runtime-provided trampoline.

In the current implementation, the Wasmtime runtime registers `grate_callback_trampoline` (in `wasmtime/run.rs`), which:

- retrieves the `VMContext` for the target grate,
- re-enters the correct Wasmtime runtime context,
- and dispatches to the unified Wasm entry function.

3i itself remains runtime-agnostic and does not execute grate code directly.

## Test

New tests for multiple register_handler calls.
New tests for high-workload concurrency grate requests.

### DIFF
--- a/src/wasmtime/crates/lind-multi-process/src/lib.rs
+++ b/src/wasmtime/crates/lind-multi-process/src/lib.rs
@@ -5,7 +5,7 @@ use cfg_if::cfg_if;
 use anyhow::{anyhow, Result};
 use std::ffi::c_void;
 use std::ptr::NonNull;
-use sysdefs::constants::lind_platform_const::{LIND_ROOT, UNUSED_ARG, UNUSED_ID, UNUSED_NAME};
+use sysdefs::constants::lind_platform_const::{UNUSED_ARG, UNUSED_ID, UNUSED_NAME};
 use threei::{threei::make_syscall, threei_const};
 use wasmtime_lind_3i::{
     get_vmctx, get_vmctx_thread, rm_vmctx, set_vmctx, set_vmctx_thread, VmCtxWrapper,


src/wasmtime/src/commands/run.rs
```diff
@@ -115,14 +101,17 @@ pub extern "C" fn grate_callback_trampoline(
 ) -> i32 {
     // Never unwind across the C boundary.
     let res = catch_unwind(AssertUnwindSafe(|| unsafe {
-        // Validatation check
-        if ctx.is_null() {
-            return threei_const::GRATE_ERR;
-        }
+        let threadid = 1; //always 1. Only fork requires threadid passing, and fork is handled specially
```
rennergade: this seems weird, or at least the comment seems misdirecting?

rennergade:
General comment: the term "runtime" here confused me (especially if we eventually are mixing actual runtimes like wasm, mpk etc). Can we think of a better name here?

src/threei/src/threei.rs
```diff
@@ -260,24 +200,25 @@ fn _call_grate_func(
...
+            runtimeid
+        ),
+    };
 
-    let rc = (entry.fn_ptr)(
-        entry.ctx_ptr,
+    let rc = (trampoline)(
         in_grate_fn_ptr_u64,
-        self_cageid,
+        grateid,
```
stupendoussuperpowers: In the previous model this represented the cageid of the calling cage and was therefore helped the `pass_fptr_to_wt` determine the originating cage. 

Here, the `grateid` is useful only for the `_call_grate_func` and I think we can just skip passing it down to the trampoline and in-turn the dispatcher function, since the grate can always determine it's grateid through `getpid()`.

src/threei/src/threei.rs
```diff
@@ -38,215 +35,158 @@ pub type RawCallFunc = fn(
...
-unsafe impl Sync for GrateFnEntry {}
+/// `set_cage_runtime` associates a cage with a runtime ID.
+///
+/// This function is called from the [wasmtime/lind-3i] when a cage is created or initialized. At that
+/// point, the runtime responsible for executing grate calls for the cage is known and recorded here.
+///
+/// If a runtime is already set:
+/// - same value => no-op
+/// - different value => panic (logic bug / double-init mismatch)
+pub fn set_cage_runtime(cageid: u64, runtime: u64) {
```
stupendoussuperpowers: Instead of runtime being a `u64`, I think it should be an enum.

src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -13,39 +16,65 @@
...
+//!
 //! At module creation time, lind-3i extracts the `VMContext` pointer associated with the newly created instance.
 //! This `VMContext` uniquely identifies the execution state of that `Store` / `Instance`. The pointer is
 //! stored in a global table indexed by `cage_id`. When lind-3i needs to transfer execution to
 //! another cage or grate, it looks up the corresponding `VMContext` pointer using the target `cage_id`.
 //! Using Wasmtime’s internal mechanisms, the `VMContext` pointer can be used to recover the associated
 //! `Store` and `Instance`, allowing execution to resume in the correct runtime context.
 //!
-//! The table intentionally stores raw `VMContext` pointers rather than typed store or instance handles.
+//! 2. **Per-cage thread VMContext tables (`VMCTX_THREADS`)**
```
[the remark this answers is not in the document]
Yaxuan-w: Updated 3i doc

src/threei/src/threei.rs
JustinCappos: In the future, this could be possible.  Don't worry too much about it now

src/threei/src/threei.rs
```diff
@@ -339,9 +279,9 @@ pub static EXITING_TABLE: Lazy<DashSet<u64>> = Lazy::new(|| DashSet::new());
 /// Panics if there is an attempt to overwrite an existing handler with a different destination cage.
 pub fn register_handler(
     in_grate_fn_ptr_u64: u64,
-    targetcage: u64,    // Cage to modify
-    targetcallnum: u64, // Syscall number or match-all indicator. todo: Match-all.
-    entry_ptr_u64: u64,
+    targetcage: u64,     // Cage to modify
+    targetcallnum: u64,  // Syscall number or match-all indicator. todo: Match-all.
+    _runtime_id: u64,    // Currently unused, reserved for future potential use
```
JustinCappos: If this is the memory location for something like MPK, then this makes sense.  

Could you list the address of the cage's handler here if you wanted?

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -450,7 +456,45 @@ impl<
 
                     // new cage created, increment the cage counter
                     lind_manager.increment();
-                    // create the cage in rustposix via rustposix fork
+                    // The main challenge in enabling dynamic syscall interposition between grates and 3i lies in Rust’s
+                    // strict lifetime and ownership system, which makes retrieving the Wasmtime runtime context across
+                    // instance boundaries particularly difficult. To overcome this, the design employs low-level context
+                    // capture by extracting and storing vmctx pointers from Wasmtime’s internal `StoreOpaque` and `InstanceHandler`
```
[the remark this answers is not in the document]
Yaxuan-w: Yes. They're same following same logic

stupendoussuperpowers (approved):

Issue #560, opened 2025-12-08
ci/cd: CI failure in PR #558: `unknown import: lind::lind-longjmp`


Yaxuan-w:
In PR #558, all tests pass when the branch is built and tested on our server. However, the PR’s CI/CD pipeline consistently fails with the following error:

```sh
Error: unknown import: `lind::lind-longjmp` has not been defined
```

The failure appears only in the CI environment, not when running the same branch on our server, and the missing import (`lind::lind-longjmp`) is unrelated to the changes in the PR (since the PR does not modify any trampoline-related code).

Yaxuan-w:
The CI error in PR #558 #562 was caused by the import not being attached to the linker.

The root cause is that the import was not attached to the linker. This issue did not surface locally because all tests passed on our development server, but it consistently failed in CI.

The reason is that: 

- On our server, the [Docker image sets `WASMTIME_PROFILE=debug`](https://github.com/Lind-Project/lind-wasm/blob/main/Docker/Dockerfile.dev#L103) and pre-compiles the debug version of wasmtime.
- In normal development workflows, running `make wasmtime` builds the release version of wasmtime.
- As a result, both debug and release binaries coexist under `src/wasmtime/target/`.
- [`make wasmtime` does not modify `WASMTIME_PROFILE`](https://github.com/Lind-Project/lind-wasm/blob/main/Makefile#L22), so the environment variable remains set to debug.
- In the `lind_run` script, [the wasmtime binary selection is based solely on `WASMTIME_PROFILE`](https://github.com/Lind-Project/lind-wasm/blob/main/scripts/lind_run#L34-L41), which means the debug binary is always executed, even when a release binary exists.

Because of this, local testing on the server always ran the debug version and did not expose the missing linker import. To force the release version locally, we had to run `make distclean`, which is not representative of how CI operates.

In contrast, CI runs the `Dockerfile.e2e` tests in a clean environment, without precompiled debug artifacts. Therefore, the failure in CI is expected and correct.

PR #572, opened 2025-12-15
ci/cd: Fix lind script and docker file

Yaxuan-w:
This PR resolves the WASMTIME_PROFILE–related issue described in #560.

Previously, our development environment implicitly relied on `WASMTIME_PROFILE` to decide which wasmtime binary to execute. In particular:

- Dockerfile.dev pre-set `WASMTIME_PROFILE=debug`.
- During normal development, make wasmtime builds the release version, while the environment variable still pointed to debug.

As a result, both debug and release binaries could coexist, but `lind_run` would always execute the debug binary.

This behavior masked issues locally and caused discrepancies between server-side development and CI, which runs in a clean environment.

This PR makes two related changes to eliminate the ambiguity:

In Dockerfile.dev:

- Removes the preset WASMTIME_PROFILE=debug.
- Avoids implicitly forcing the debug profile during development.

In lind_run script:

- Changes the wasmtime selection logic from:
   - “select based on WASMTIME_PROFILE”
   - to:
   - “select based on which wasmtime binary actually exists”.
- If exactly one of {release, debug} exists, it is selected automatically.
- If both binaries exist, the script now requires explicit user input via `--wasmtime-profile={release|debug}`.

In lind_compile script:

- Update lind_compile to determine wasmtime bin according to either user input or path

### DIFF
--- a/scripts/lind_run
+++ b/scripts/lind_run
@@ -20,7 +20,7 @@ Usage:
   lind_run [OPTIONS] <program.wasm|program.cwasm> [wasmtime args...]
 
 Options:
-  --wasmtime-profile=release|debug
+  -p=release|debug, --wasmtime-profile=release|debug
       Explicitly select which wasmtime build to use.
       Required if both release and debug binaries exist.
 
@@ -73,7 +73,7 @@ while [[ $# -gt 0 ]]; do
       print_help
       exit 0
       ;;
-    --wasmtime-profile=*)
+    -p|--wasmtime-profile=*)
       WASMTIME_PROFILE="${1#*=}"
       shift
       ;;
--- a/scripts/lind_compile
+++ b/scripts/lind_compile
@@ -22,8 +22,8 @@ Usage:
   lind_compile [OPTIONS] <path/to/file.c>
 
 Options:
-  --wasmtime-profile=release|debug
-  --wasmtime-profile release|debug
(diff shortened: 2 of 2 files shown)


scripts/lind_run
```diff
@@ -31,13 +61,78 @@ if [[ -z "${REPO_ROOT}" || ! -d "${REPO_ROOT}" ]]; then
...
+# Track existence/executability explicitly to avoid ambiguous fallbacks.
+rel_ok=0; dbg_ok=0
+[[ -x "${REL_BIN}" ]] && rel_ok=1
+[[ -x "${DBG_BIN}" ]] && dbg_ok=1
+
+# --- Decide profile/binary ---
+# Policy:
+#   - If user explicitly specifies a profile, honor it strictly.
+#   - Otherwise:
+#       * Use the only available profile if exactly one exists.
```
rennergade: I think this isn't a great way to go about this since we can easily be confused that were using one while actually using the other. Id rather there be some sort of warning/error if there's a mismatch and we fix the environments properly.

m-hemmings (approved):

scripts/lind_run
rennergade: having the separate containers have the proper builds as defaults as well as not have dual version builds at once

PR #578, opened 2025-12-17
Improvements to lind_compile

### LINKED ISSUE #573: build: Improvements to lind_compile
There're several incremental improvements to the lind_compile script to make it more flexible and developer-friendly.

## 1. Support running compilation steps independently

Currently, lind_compile always executes the full pipeline: .c  ->  .wasm  ->  (wasm-opt)  ->  .cwasm. It would be useful to add options that allow developers to run individual steps independently, for example:

- Compile only: .c to .wasm
- Optimize only: wasm-opt on an existing .wasm
- Precompile only: wasm to .cwasm

Possible interfaces could include step-selection flags (e.g., --compile-only, --opt-only, --precompile-only) etc..

## 2. Option to print executed compilation commands

Adding an option to print (or echo) the exact command lines used for each compilation step would improve transparency and debuggability.
For example: A flag such as `--print-cmd` or `--verbose`
When enabled, the script would print the full clang, wasm-opt, and wasmtime compile command lines after executing them.

J7-7-7:
Fixes #573

1) Adds --compile-only, --opt-only, --precompile-only modes to scripts/lind_compile so each stage can be run independently.

2) Adds --print-cmd / --verbose to print the underlying commands for easier debugging.

3) Keeps default behavior unchanged when no flags are provided.

### DIFF
--- a/scripts/lind_compile
+++ b/scripts/lind_compile
@@ -37,7 +37,7 @@ usage: $(basename "$0") [OPTIONS] <source>
     --precompile-only   : .wasm -> .cwasm
 
   Common options:
-    --print-cmd, --verbose  print the underlying commands that are executed
+    -v, --print-cmd, --verbose  print the underlying commands that are executed
     -h, --help              show this help
 
   Examples:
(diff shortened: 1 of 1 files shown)


rennergade:
@qianxichen233 can you rereview? if you want more specific short options here please specify them specifically

PR #579, opened 2025-12-19
Init Ctypes

rennergade:
This PR inits ctypes in our crt (wasn't added in initial custom crt impl.)

This allows us to call functions like isalpha that are needed for various apps including python.


rennergade:
@qianxichen233 I updated the test to be inclusive. CI failure is due to unrelated timeout bug.

qianxichen233 (approved):

Issue #581, opened 2025-12-22
`copy_data_between_cages`: broken due to inconsistent address translation


stupendoussuperpowers:
Since we now have address translation as a part of `glibc`, we need to update the threei implementation for `copy_data_between_cages` to be cognizant of these changes. 

### What's broken 

Example scenario: A cage calls the `write_syscall` which is registered to a grate. 

Inside the `write_grate` function we call the following: 
```
copy_data_between_cages(thiscage, arg2cage, arg2, arg2cage, (uint64_t)buffer, thiscage, count, 0);
```

Here, we pass in the `srcaddr` and `destaddr` as arguments. The `srcaddr` (`arg2`) in this case is host-coded since this address has passed through `glibc`, however the `dstaddr` (`buffer`) refers to a local address within the grate's wasm module. 

The `cage::memory::check_addr` function which performs pre copy checks to determine if source and destination address ranges are valid expects uaddr-coded addresses so the checks performed fail for the cage.

### Proposed Solution

We need to make the following changes to fix this:
1. Update `check_addr` to expect host-coded addresses. 
2. In glibc, when calling `copy_data_between_cages` add logic for converting possible uaddr addresses to hostaddr. This would ensure that once we are within threei, any address arguments are consistently host-coded.

rennergade:
@stupendoussuperpowers if your recent PR fixes all of this feel free to close

PR #582, opened 2025-12-22
Fix address translation for `copy_data_between_cages`

stupendoussuperpowers:
This addresses #581. 

Fix works by:

- Adding a `TRANSLATE_UADDR_TO_HOST(uaddr, cageid)` macro. 
- Change `check_addr` to subtract `vmmap.base_address` before passing this to `check_addr_mapping`. 
- Additionally, add implementation for the `sc_convert_uaddr_to_host`. Currently this function doesnt do anything. This is not required for this fix, can be removed if this is not needed in the long run. 

My understanding is:

That within glibc, we only have access to `__lind_base` and `__lind_cageid` for the current/calling cage. i.e. when a grate calls `copy_data_between_cages` it is only capable of converting between `uaddr` and `host` for this specific `__lind_cageid`. The macro specified above assumes that if the cageid matches the `__lind_cageid`, it needs to append the base memory, otherwise it can ignore it. 

If there is a cleaner approach to doing this translation, please let me know.

### DIFF
--- a/src/typemap/src/datatype_conversion.rs
+++ b/src/typemap/src/datatype_conversion.rs
@@ -304,7 +304,7 @@ pub fn sc_convert_uaddr_to_host(uaddr: u64, addr_cageid: u64, cageid: u64) -> u6
         panic!(
             "sc_convert_uaddr_to_host: invalid uaddr {:#x} - expected a cage-relative address",
             uaddr
-       );
+        );
     }
 
     uaddr + base_addr


src/glibc/lind_syscall/addr_translation.h
```diff
@@ -42,10 +42,22 @@ extern "C"
     return __lind_base + (uint64_t) (uintptr_t) p;
   }
 
+  static inline uint64_t
+  __lind_translate_uaddr_to_host (const uint64_t uaddr, const uint64_t cageid)
```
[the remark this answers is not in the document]
rennergade: Oh I see that makes sense. Do we have to handle a NULL condition here?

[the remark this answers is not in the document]
stupendoussuperpowers: Don't think so, I believe it's handled at a later stage for copy_data, but will run some tests

src/cage/src/memory/memory.rs
```diff
@@ -204,9 +204,12 @@ pub fn check_addr(cageid: u64, arg: u64, length: usize, prot: i32) -> Result<boo
     // Get write lock on virtual memory map
     let mut vmmap = cage.vmmap.write();
 
+    let base_addr = vmmap.base_address.unwrap() as u64;
```
Yaxuan-w: Add brief explanation here

src/glibc/lind_syscall/addr_translation.h
```diff
@@ -42,10 +42,22 @@ extern "C"
...
+      return __lind_base + uaddr;
+
+    return uaddr;
+  }
+
 // Convenience macro for call sites
 #define TRANSLATE_GUEST_POINTER_TO_HOST(p)                                    \
   __lind_translate_ptr_to_host ((const void *) (p))
 
+#define TRANSLATE_UADDR_TO_HOST(uaddr, cageid)                                \
```
Yaxuan-w: Add comments about the functionality, use cases, and why we need this

Yaxuan-w (changes requested):
Great job! Could you also add unit test cases for this?

src/typemap/src/datatype_conversion.rs
```diff
@@ -290,6 +290,20 @@ pub fn sc_convert_uaddr_to_host(uaddr: u64, addr_cageid: u64, cageid: u64) -> u6
...
+    // Do not convert on NULL.
+    if uaddr == 0 {
+        return uaddr;
+    }
+
+    let cage = get_cage(addr_cageid).unwrap();
+    let vmmap = cage.vmmap.read();
+    let base_addr = vmmap.base_address.unwrap() as u64;
+
+    // Only convert to host if not already converted.
```
rennergade: We should error here instead of being hacky/failing gracefully, this could lead to problems down the line.

rennergade (approved):
Ok looks good. Approved.

Issue #525, opened 2025-11-16
Standardize executable path location


rennergade:
As of now, lind can execute a cwasm file living anywhere on the hosts filesystem, without regarding its relationship to lind_root. This can cause problems with consistency [as seen here](https://github.com/Lind-Project/lind-wasm/issues/481#issuecomment-3539200225), possible security problems, or just [lack of a general philosophy ](https://github.com/Lind-Project/lind-wasm-apps/pull/18#issuecomment-3530457421).

The simple fixes are just to change the lind-clang script to mv any executable to lind-root, or a relative path from there, and change lind-wasm to only execute things starting from lind-root.

We probably want to consider adding checks at runtime to only accept files with proper paths as well.

PR #585, opened 2025-12-29
Ensure compiled executables are moved to LIND_ROOT

ssannkkallpp:
This PR includes changes for issue https://github.com/Lind-Project/lind-wasm/issues/525

- Modified lind_compile to store all compiled .cwasm files in LIND_ROOT
- Added validation to lind_run to ensure executables are in LIND_ROOT
- Modified wasmtestreport.py to expect executables in LIND_ROOT, updated Makefile to clean up executables


scripts/wasmtestreport.py
rennergade: I think you need to move the testfiles to lind root as well

PR #589, opened 2026-01-01
Remove `sc_convert_uaddr_to_host` usage for address conversions in rawposix and 3i

stupendoussuperpowers:
The `futex_syscall`, `getrandom_syscall`, and `copy_data_between_cages` still use the `sc_convert_uaddr_to_host` function, which is not necessary with the address translations changes. The use of this helper function has been removed from other functions that used them, this PR cleans up the remaining occurrences.

### DIFF
--- a/src/rawposix/src/fs_calls.rs
+++ b/src/rawposix/src/fs_calls.rs
@@ -279,21 +279,21 @@ pub fn futex_syscall(
     futex_op_cageid: u64,
     val_arg: u64,
     val_cageid: u64,
-    val2_arg: u64,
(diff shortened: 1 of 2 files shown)


Yaxuan-w (commented):
Changes make sense to me, but the CI tests failed.

Yaxuan-w (approved):
Looks good now! I'd suggest to change the title to explicitly mention that we're removing in both rawposix and 3i. Maybe sth like "Remove usage of sc_convert_uaddr_to_host in rawposix and 3i"?

PR #594, opened 2026-01-05
Fix getsockname syscall: convert guest pointers to host pointers

ChinmayShringi:
- Use sc_convert_buf() to properly convert guest pointers (arg3, addr_arg) to host pointers
- Fix addrlen parameter handling by reading from converted pointer
- Use temporary sockaddr_storage buffer for kernel call
- Copy results back using copy_out_sockaddr() with converted pointers
- Fix C test: change int addrlen to socklen_t addrlen
- Fix C test: update main() signature and printf to show correct output

Fixes EFAULT error when calling getsockname on Unix domain sockets.

### DIFF
--- a/src/rawposix/src/net_calls.rs
+++ b/src/rawposix/src/net_calls.rs
@@ -1333,12 +1333,12 @@ pub fn getsockname_syscall(
     arg6_cageid: u64,
 ) -> i32 {
     let fd = convert_fd_to_host(fd_arg, fd_cageid, cageid);
-    let addr = addr_arg as *mut u8;
+    let user_addr = addr_arg as *mut SockAddr;
+    let lenp = arg3 as *mut socklen_t;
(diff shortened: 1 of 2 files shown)


rennergade:
I think there are two separate issues here.

First it does not make sense to convert the sockaddr before calling getsockname. In the test the argument we pass in server_addr2 is blank so there is nothing to convert at that point. getsockname is what retrieves the sockaddr from the kernel. The conversion from host sockaddr to user sockaddr should happen after the syscall not before it. Right now this is happening in the wrong order.

Second the sockaddr we are getting back from the kernel looks wrong. From the test output we are getting an absolute path even though the input was relative. The absolute path is also incorrect because it is missing the final slash. This suggests we are probably constructing or appending the wrong sockaddr during the bind syscall.

PR #600, opened 2026-01-08
Add support for `sched_yield` syscall in rawposix

### LINKED ISSUE #596: Add sched_yield
Both https://github.com/Lind-Project/lind-wasm/issues/595 and https://github.com/Lind-Project/lind-wasm/pull/591 mention needing sched_yield().

Both seem to solve it by adding an alias to the i686 file, though this implementation only returns 0.

We should figure out if we should actually be using/linking the sched_yield out of nptl since thats the origin of most threading mechanisms. 

Whichever is the right one to link we should add 3i/rawposix paths so this will actually switch threads when called.

stupendoussuperpowers:
Closes #596 

Adds support for sched_yield_syscall in rawposix and the required changes in glibc and 3i.

### DIFF
--- a/src/rawposix/src/sys_calls.rs
+++ b/src/rawposix/src/sys_calls.rs
@@ -931,17 +931,17 @@ pub fn sigprocmask_syscall(
 pub fn sched_yield_syscall(
     cageid: u64,
     arg1: u64,
-    arg1cage: u64,
+    arg1_cageid: u64,
     arg2: u64,
-    arg2cage: u64,
+    arg2_cageid: u64,
     arg3: u64,
-    arg3cage: u64,
+    arg3_cageid: u64,
     arg4: u64,
-    arg4cage: u64,
+    arg4_cageid: u64,
     arg5: u64,
-    arg5cage: u64,
+    arg5_cageid: u64,
     arg6: u64,
-    arg6cage: u64,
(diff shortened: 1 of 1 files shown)


src/rawposix/src/sys_calls.rs
```diff
@@ -921,6 +921,45 @@ pub fn sigprocmask_syscall(
...
+        && sc_unusedarg(arg5, arg5_cageid)
+        && sc_unusedarg(arg6, arg6_cageid))
+    {
+        panic!(
+            "{}: unused arguments contain unexpected values -- security violation",
+            "sched_yield_syscall"
+        );
+    }
+
+    (unsafe { sched_yield() }) as i32
```
Yaxuan-w: need to use `libc::` if we're calling kernel

stupendoussuperpowers:
Based on further testing this implementation does what we'd expect `sched_yield` to do. 

If we add a busy-wait in the inner loop for a test like this https://github.com/Lind-Project/lind-wasm/issues/596#issuecomment-3720015117, we do see the effects of `sched_yield`.

The `pthread_self` recorded inside the thread builder in `pthread_create_call` is same as the one recorded inside `rawposix::sched_yield_syscall`, which also aligns with this implementation.

PR #608, opened 2026-01-16
Fix waitpid and get_cage api to handle pid=-1 (wait for any child)

### LINKED ISSUE #543: Tests: Panic raised by waitpid syscall when invalid pid is passed as argument
Function Used:

```
void test_waitpid_non_child() {
    printf("\n[TEST 9] waitpid on non-child pid\n");
    pid_t pid = fork();
    if (pid == 0) _exit(1);

    pid_t fake = pid + 10000;
    int status = 0;
    pid_t r = waitpid(fake, &status, WNOHANG);
    printf("[PARENT] waitpid non-child returned=%d errno=%d (%s)\n", (int)r, errno, strerror(errno));

    waitpid(pid, &status, 0);
}
```
Response:

```
[TEST 9] waitpid on non-child pid

thread 'main' (174811) panicked at crates/cage/src/cage.rs:100:9:
Cage ID is outside of valid range
```

rishabhBudhouliya:
Fixes #543 and https://github.com/Lind-Project/lind-wasm-apps/issues/31

Also added relevant unit tests to cover the newly introduced code.

# Purpose
This PR introduces two distinct changes in [waitpid](https://github.com/Lind-Project/lind-wasm/blob/2461272db0cec6fee916a70db13db7213341b524/src/rawposix/src/sys_calls.rs#L247) and [cage.rs
](https://github.com/Lind-Project/lind-wasm/blob/2461272db0cec6fee916a70db13db7213341b524/src/cage/src/cage.rs#L125)

Specifically, 

1. https://github.com/Lind-Project/lind-wasm-apps/issues/31 exposed us with incorrect handling of signed integers as `cageid_arg` (which was being sent by Bash, as pid = -1). The fix involves using 
```
let cage_id_to_wait = sc_convert_sysarg_to_i32(cageid_arg, cageid_arg_cageid, cageid);
```
to cast the u64 unsigned data into a signed i32 data. 

2. But the real issue underneath the behavior that we saw with the bash issue was #543. While checking for a valid cage id, we panic in case the cage id is out of bound. This behavior crashes the host code which was agreed to be an undesirable behavior. For this problem, the fix I've applied is:
```
pub fn get_cage(cageid: u64) -> Option<Arc<Cage>> {
    if cageid >= MAX_CAGEID as u64 {
        return None;
    }
```
This returns None to the caller and the caller is expected to deal with it appropriately. Turns out our waitpid is dealing with it by returning an appropriate error.

A point of discussion can be that I've still left the panic behavior untouched here: 
https://github.com/Lind-Project/lind-wasm/blob/2461272db0cec6fee916a70db13db7213341b524/src/cage/src/cage.rs#L98-L102

The reason to do so was
1) keep the blast radius minimal
2) apart from `get_cage()`, only two other internal apis use `check_cage_id`: `add_cage()` and `remove_cage()` which both are internal apis and can afford to crash if there is a problem with the input cage id. But `get_cage()` api can be used by an external user (host) where crashing doesn't seem like an acceptable behavior. 

# Related Issues
#543 and https://github.com/Lind-Project/lind-wasm-apps/issues/31


rishabhBudhouliya:
Thanks for that suggestion @rennergade . I'll also be more diligent about checking the e2e tests in lind server before pushing them here. 

Oh I thought PRs were only allowed from forked repositories, good to know!

Yaxuan-w (commented):
Besides nicks comments about tests, overall logic makes sense to me

rishabhBudhouliya:
@Yaxuan-w and @rennergade thanks for your review. I'm grateful for the prompt responses I got on this PR. 
I've modified the original description of the PR and would like an overall review of one point that I think might be worth surfacing since this PR brings in two changes, not one: 

> A point of discussion can be that I've still left the panic behavior untouched here:
> 
> [lind-wasm/src/cage/src/cage.rs](https://github.com/Lind-Project/lind-wasm/blob/2461272db0cec6fee916a70db13db7213341b524/src/cage/src/cage.rs#L98-L102)
> 
> The reason to do so was
> 
> keep the blast radius minimal
> apart from get_cage(), only two other internal apis use check_cage_id: add_cage() and remove_cage() which both are internal apis and can afford to crash if there is a problem with the input cage id. But get_cage() api can be used by an external user (host) where crashing doesn't seem like an acceptable behavior

Yaxuan-w (approved):

rennergade (changes requested):
Actually if get_cage returns None now won't it panic on unwrap?

That could still cause a user submitted panic right? If an invalid pid is passed to wait I think that would just return an errno right?

rennergade:
> @rennergade that's true, this can still cause a panic at unwrap. However, this was done purposefully to differentiate between current cage id and user submitted cage id. For example,
> 
> https://github.com/Lind-Project/lind-wasm/blob/2461272db0cec6fee916a70db13db7213341b524/src/rawposix/src/sys_calls.rs#L287
> 
> can potentially cause a panic, but I assumed it should because current cage id is invalid and that must be a programmatic error rather than a user-input related error. Please correct me if I'm wrong with my assumptions here.

its still from the input of a user (ie the waitpid arg). So I think in this case instead of unwrapping we should return ECHILD if its none: https://linux.die.net/man/2/waitpid

Yaxuan-w:
> its still from the input of a user (ie the waitpid arg). So I think in this case instead of unwrapping we should return ECHILD if its none

I recalled our previous discussion was panic asap for our implementation, and switch to soft panic afterwards. Probably we should use same methodology here?

rennergade:
> > its still from the input of a user (ie the waitpid arg). So I think in this case instead of unwrapping we should return ECHILD if its none
> 
> I recalled our previous discussion was panic asap for our implementation, and switch to soft panic afterwards. Probably we should use same methodology here?

We want to panic if we're doing something unconventional (like how we don't support certain ioctl cmds or support mapping pages executable). Here we're not doing anything weird, if a user supplies an invalid pid to waitpid that returns an errno in Linux so we should do the same.

rennergade:
> Note, it's fine to go either way here. If we do as Alice suggests and find it annoying that normal code does this and stops, we could switch later to returning the error code.

Sounds good. Either way we should debug_panic! or errno here and not have a generic panic from unwrap.

rishabhBudhouliya:
Based on my understanding of the discussion here, I've committed a change that:
1. get_cage() in cage.rs returns a None as it isn't aware of Linux specific Errno subjective to the caller syscall (separation of concerns)
2. replaces all `get_cage().unwrap()` calls with match based error handling.

rennergade:
> Based on my understanding of the discussion here, I've committed a change that:
> 
>     1. get_cage() in cage.rs returns a None as it isn't aware of Linux specific Errno subjective to the caller syscall (separation of concerns)
> 
>     2. replaces all `get_cage().unwrap()` calls with match based error handling.

Sorry about this. I want to apologize because I think there is some confusion here that I added to. The way the actual Linux waitpid() works is that it takes a pid as its first argument. We call pids cageids so whenever this was patched in that first argument was labeled `cageid_arg`, but that is also somewhat confusing because all syscalls in rawposix receive their own "caller" cageid as the first argument.

If `cageid` (the first arg for all rawposix syscalls) is None then that actually should panic. The caller not existing while trying to call a syscall means something is very corrupt and we should abort.

On the other hand, waitpid's argument `cageid_arg` which is the cageid were waiting on is user provided and can be None, this is where we should appropriately errno or debug_panic.

It makes sense to revert the changes you did to get_cageid in the other syscalls (as well on the caller cageid here). We just need to appropriately error if the child cageids are invalid. It may make sense to name cageid_arg to something like child_cageid_arg here to avoid confusion?

rishabhBudhouliya:
@rennergade Thanks for catching the confusion. Here are the changes I've made as I've tried to understand how behavior of error handling depends on the nature of cage id (child cage id or caller's cage id). 

- `get_cage()` returns None for out-of-range IDs                                                                                                                                                                                                                                                                                     - `waitpid` handles user-provided cage ID gracefully (returns ECHILD)
- Caller cage IDs use .unwrap()
- Renamed `cageid_arg` to `child_cageid_arg`

rennergade:
> I think we should clearly separate the syscall argument conversion/interpretation phase from the argument usage phase...

This is a really good point and I agree, its worth opening up a separate issue on this

Issue #611, opened 2026-01-19
Remove LIND_ROOT style namespacing from RawPOSIX


rennergade:
We should remove the current trusted LIND_ROOT namespacing logic from src/rawposix in this repo (lind-wasm/src/rawposix) because the existing approach relies on hand-rolled path rewriting and containment checks in userspace, which has proven fragile and hard to get fully correct. This code expands our trusted computing base and complicates testing and CI, and it still doesn’t match native Linux filesystem semantics in edge cases.

A good place to start is looking at helper functions in path_conversion.rs and network_helpers.rs which should be removed then we can remove their invocation from individual syscalls.

rennergade:
After we do this it should be simple to invoke chroot on the runtime. Either as a separate PR or as part of [lind-boot](#510 )

Yaxuan-w:
Post [the comment about roadmap](https://github.com/Lind-Project/lind-wasm/pull/626#issuecomment-3785412376) here for future reference:

> Its just basically:
> 
> remove LIND_ROOT and path functions
> add chroot
> fix test paths

PR #626, opened 2026-01-22
refactor: remove LIND_ROOT namespace from RawPOSIX

rishabhBudhouliya:
# Purpose
The idea is to remove `LIND_ROOT` which is a handrolled filesystem namespacing isolation mechanism from RawPOSIX.

LIND_ROOT was a userspace path rewriting mechanism that: 
- prefixed all paths with '/home/lind/lind-wasm/src/tmp' before kernel calls
- was stripped from kernel results before returning to the user

Action taken:
- remove path prefixing and stripping across multiple syscalls involved with the fs
- paths do not contain LIND_ROOT manipulation now

# Issues
#611 

Currently, no additional test coverage has been added to reflect the path related changes.

### DIFF
--- a/src/rawposix/src/sys_calls.rs
+++ b/src/rawposix/src/sys_calls.rs
@@ -1052,6 +1052,13 @@ pub fn setitimer_syscall(
 /// detailed error messages will be printed if set
 pub fn rawposix_start(verbosity: isize) {
     let _ = VERBOSE.set(verbosity); //assigned to suppress unused result warning
+    unsafe {
+        let tmp_path = CString::new("/tmp").unwrap();
+        libc::mkdir(tmp_path.as_ptr(), 0o755);
+        let ret = libc::chroot(tmp_path.as_ptr());
+        let root = CString::new("/").unwrap();
+        libc::chdir(root.as_ptr());
+    }
     cagetable_init();
 
     fdtables::register_close_handlers(FDKIND_KERNEL, fdtables::NULL_FUNC, kernel_close);


rennergade:
Thank you for putting this up so quickly @rishabhBudhouliya 

Before you change around the tests let @Yaxuan-w and I review the code changes here and then let's talk about chrooting tomorrow since that will change things

src/typemap/src/path_conversion.rs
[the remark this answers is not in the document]
rennergade: @Yaxuan-w if those get removed does it make sense to deprecate this file and move the rest of the functions to datatype_conversion?

src/rawposix/src/fs_calls.rs
```diff
@@ -1885,62 +1839,26 @@ pub fn readlinkat_syscall(
...
-    let libcret = if virtual_fd == libc::AT_FDCWD {
-        // Case 1: AT_FDCWD - path is already converted by sc_convert_path_to_host
-        unsafe {
-            libc::readlink(
-                path.as_ptr(),
-                libc_buf.as_mut_ptr() as *mut c_char,
-                libc_buflen,
-            )
-        }
+    let ret = if virtual_fd == libc::AT_FDCWD {
```
rennergade: do we need this logic now if were working from chroot? I'm not sure.

src/typemap/src/path_conversion.rs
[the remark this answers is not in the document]
Yaxuan-w: But we need to keep functions of u64 -> str

rennergade:
> Thanks for doing this! Do we have sth like the roadmap of doing this? I assume this is the first step of adding chroot, right?

Its just basically:

1. remove LIND_ROOT and path functions
2.  add chroot
3. fix test paths

src/rawposix/src/fs_calls.rs
rishabhBudhouliya: I'm sorry, I don't follow how this logic is related to chroot or fixed path namespacing?
I read about FDCWD and it seems to be a sentinel value that signifies the CWD file descriptor. If we were to pass that in the else block, wouldn't that cause a failure because our virtual file descriptor table wouldn't have an entry for it?

rennergade: Ok yeah I think you're correct here

Issue #73, opened 2024-12-30
Fix wasmtime epoch_interruption


rennergade:
Our current idea for handling signals involves using the epoch_interruption feature of wasmtime to asynchronously interrupt code. As of the last time I checked in with @qianxichen233 this feature wasn't working anymore, possibly due to asyncify. Let's investigate the causes here and see if we can get interruption to work.

qianxichen233:
Previously, the issue is about when epoch_interruption is enabled, new wasm instances created by fork/thread will be interrupted immediately once they start the execution (but the initial wasm instance will not). I highly suspect that this was just due to some incorrect handling of epoch initialization for new wasm instances, and I just confirmed this and fixed it - I only need to add one line of code when initializating new wasm instances for fork/thread and it could work.
Based on this, I also start experimenting signals with epoch_interruption. I did a minimal prototype for signal implementation, and I am able to run a process that stuck in a dead loop, but are able to jump to a custom user function that are able to do things like printf when host intended to trigger it.
However, `fork` is still not able to run within the custom user function (or signal handler). When I am trying to do fork within custom user function, I am getting `unreachable instruction executed error`, which is something I am expecting because I believe fork within signal handler shouldn't work correctly with current approach.

PR #645, opened 2026-01-28
register flock syscall (#73) in dispatcher table

Bhanuu01:
Added flock syscall (syscall 73) in dispatcher table.


Yaxuan-w (changes requested):
You need to add entry for (73, flock_syscall) pair in syscall table too

PR #647, opened 2026-01-28
Patching secure mode (`--features secure`)

stupendoussuperpowers:
Baseline fixes:
- Add: Missing feature in src/wasmtime, plumb it with typemap/secure
- Fix: Invalid ordering of cfg(fast) and cfg(secure)
- Fix: Broken compilation of typemap due to invalid imports, return types, duplicate declarations, and non-updated util function usage

Specific fix for `sc_unusedargs`
- Fix: sc_unusedargs that is used throughout rawposix should not expect `*_cageid` is empty, since these should always be populated.


stupendoussuperpowers:
As an addendum - We should prolly think about adding a CI check to make sure we can at least compile with `--features secure`. Most of these issues were majorly due to missed changes during refactors and no one testing with secure mode.

Issue #593, opened 2026-01-04
add epoll_create1 syscall


qianxichen233:
epoll_create1 takes an additional flag argument than epoll_create, which is only used for CLOEXEC. This should be straightforward and easy to implement

PR #661, opened 2026-01-28
Add epoll_create1 syscall

vidyalakshmir:
Add epoll_create1 syscall
[Issue #593 ](https://github.com/Lind-Project/lind-wasm/issues/593)

- Made changes in glibc, 3i and rawposix. 
- Added test case to test if the flags passed as argument to epoll_create1 are validated and handled correctly.

### DIFF
--- a/src/rawposix/src/net_calls.rs
+++ b/src/rawposix/src/net_calls.rs
@@ -8,7 +8,7 @@ use std::collections::{HashMap, HashSet};
 use std::time::Instant;
 use std::{mem, ptr};
 use sysdefs::constants::err_const::{get_errno, handle_errno, syscall_error, Errno};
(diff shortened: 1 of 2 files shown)


src/rawposix/src/net_calls.rs
```diff
@@ -586,6 +586,87 @@ pub fn epoll_create_syscall(
...
+///     - flags_arg: Bitmask of flags (only `EPOLL_CLOEXEC` is recognized).
+///     - flags_cageid: cage ID for flags_arg validation.
+///     - arg2-arg6: unused arguments with their respective cage IDs.
+///
+/// ## Returns:
+///     - positive value: file descriptor for the new epoll instance
+///     - negative value: error occurred (errno=EINVAL if invalid flags are passed)
+
+
+
```
Yaxuan-w: Unnecessary blank lines

vidyalakshmir: Updated it, but still getting format error.

rennergade: you should just be able to run cargo fmt on your code

PR #684, opened 2026-02-04
lind-boot: register RawPOSIX syscall handlers in 3i during rawposix startup

Yaxuan-w:
During startup, rawposix_start now explicitly installs the RawPOSIX syscall function pointers into the 3i dispatch table, ensuring that subsequent syscalls from cages are correctly routed to the RawPOSIX implementation.

### DIFF
--- a/src/rawposix/src/init.rs
+++ b/src/rawposix/src/init.rs
@@ -10,7 +10,7 @@ use std::path::PathBuf;
 use std::sync::atomic::{AtomicI32, AtomicU64, Ordering::*};
 use std::sync::Arc;
 use sysdefs::constants::{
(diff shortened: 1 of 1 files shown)


src/rawposix/src/syscall_table.rs
```diff
@@ -73,14 +72,14 @@ pub const SYSCALL_TABLE: &[(u64, RawCallFunc)] = &[
...
     (54, setsockopt_syscall),
     (55, getsockopt_syscall),
+    (56, fork_syscall),
     (57, fork_syscall),
     (59, exec_syscall),
     (60, exit_syscall),
     (61, waitpid_syscall),
     (62, kill_syscall),
     (67, shmdt_syscall),
     (72, fcntl_syscall),
```
stupendoussuperpowers: Adding to Nick's comment about randomly missing syscalls

PR #691, opened 2026-02-04
Add support for flags to be passed to wasmtestreport.py and lind_comp…

tashabits:
## ISSUE

The current build workflow relies on pre-constructed default compiler flag sets, making it difficult to explicitly control or experiment with compiler behavior. This limits reproducibility, debugging, and per-test customization. The objective of this work is to treat compiler flags as explicit, user-provided inputs rather than implicit defaults embedded in wrapper logic.

---

### 1. `lind_compile.sh`

- Added support for passing arbitrary clang arguments directly to the compiler, including arguments provided after the source file and those separated using `--`.
- Introduced a `--no-default-clang-flags` option to disable optional default clang flags while preserving mandatory configuration and sysroot validation.
- Improved debug output to report parsed arguments and to distinguish between default, user-supplied, and effective clang flags.

### 2. `wasmtestreport.py`
 
- Uses `--compile-flags` to pass extra flags to both lind and native builds.
- Parses compile flags that begin with `-` correctly
- Supports a `--dir-flags` JSON file that maps directory prefixes to extra lind/native flags.
- Automatically appends `-lm` for tests under the configured math directory (`MATH_TEST_DIR`, defaults to `math`).

---

### `lind_compile.sh`

**Default compilation (with default flags enabled):**
```bash
./scripts/lind_compile.sh path/to/program.c
```

**Passing additional clang arguments:**
```bash
./scripts/lind_compile.sh path/to/program.c -O0 -g -DDEBUG
```

**Using `--` to explicitly delimit clang arguments:**
```bash
./scripts/lind_compile.sh path/to/program.c -- -O0 -g -DDEBUG
```

**Disabling optional default clang flags:**
```bash
./scripts/lind_compile.sh --no-default-clang-flags path/to/program.c -- -O0 -g
```

---

#### Passing compile flags via CLI

**Apply flags to both lind and native builds:**
```bash
python3 wasmtestreport.py --compile-flags "-O0 -g -DDEBUG"
```

### DIFF
--- a/scripts/wasmtestreport.py
+++ b/scripts/wasmtestreport.py
@@ -14,17 +14,10 @@
 #   "./wasmtestreport.py --pre-test-only" to copy the testfiles to lind fs root(does not run tests)
 #   "./wasmtestreport.py --clean-testfiles" to delete the testfiles from lind fs root(does not run tests)
 #   NOTE: without the last two testfiles arguments, we will always copy the test cases and then run the tests
-#   Per-directory compile flags can be specified by adding a "compile_flags.json" file next
-#   to tests (or in a parent test directory) with {"compile-flags": [...]}.
-#   Example compile_flags.json:
-#   {
-#     "compile-flags": ["-pthread", "-lpthread", "-O2", "-g"]
-#   }
 import json
 import os
 import subprocess
 from pathlib import Path
-from typing import Dict
 import argparse
 import shutil
 import logging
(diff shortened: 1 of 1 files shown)


m-hemmings:
> This looks good.
> 
> One thing which was not clear to me is how are the flags specific to test directory passed in wasmtestreport.py
> 
> Also one question about lind_compile.sh, does lind_compile flags always precede the program name and clang flags are always provided after the program name
> 
> scripts/lind_compile.sh hello.c --compile-only hello.c -O0 -g -DDEBUG

Your example is correct:
- lind_compile flags before program name
- clang flags after program name

vidyalakshmir (approved):

PR #693, opened 2026-02-05
lind-boot: exec

Yaxuan-w:
Add execve implementation in lind-boot

**NOTICE:**

- Need to be merged with other lind-boot PRs
- Cannot pass ci/cd individually


src/rawposix/src/sys_calls.rs
```diff
@@ -168,7 +168,24 @@ pub fn exec_syscall(
     *threadid_guard = 0;
     drop(threadid_guard);
 
-    0
+    threei::make_syscall(
```
rennergade: I'm a little unclear about how this all works, we use threei to get to the rawposix exec and then threei again to get to the wasmtime part? I know we've discussed this but its slipping my mind a bit. Is there a doc? Maybe worth commenting here too

PR #694, opened 2026-02-05
lind-boot: Update lindctx struct

Yaxuan-w:
Update lindctx struct and initialization logic to fit lind-boot.

Need to be merged with other lind-boot together.


rennergade (approved):
Just simple fixes this is fine.

Issue #643, opened 2026-01-27
lind-boot: slow startup time due to wasm-only loading path


Yaxuan-w:
In `lind-boot`, we currently cannot reuse the existing wasm -> cwasm compilation pipeline.

The cwasm artifacts generated from the original toolchain are not compatible with the current lind-boot execution environment, so we are forced to load and instantiate the raw Wasm binary directly.

This issue is mainly a TODO to track future work.

PR #695, opened 2026-02-05
lind-boot: use .wasm files for wasmtestreport.py by default

### LINKED ISSUE #649: lind-boot: update testing script
After the lind-boot changes are merged, the existing testing scripts will need to be updated to reflect the new execution and CLI structure.

stupendoussuperpowers:
Closes #649 

The test runner now uses `.wasm` files for the new `lind-boot` based `lind_run` script. 

It also adds a new flag, `--allow-pre-compiled` for future proofing, that would compile and run these tests based with `.cwasm` files like is the current behaviour.


Yaxuan-w (dismissed):

rennergade:
> 1. lind-boot doesnt currently support `.cwasm` files: [lind-boot: slow startup time due to wasm-only loading path #643](https://github.com/Lind-Project/lind-wasm/issues/643)
> 
>     2. After chroot changes, we need to place binaries in LINDFS_ROOT: [lind-boot with chroot changes current user-provided binary path #703](https://github.com/Lind-Project/lind-wasm/issues/703)

the chroot stuff makes sense. The questions in that issue were never answered though?

PR #698, opened 2026-02-05
lind-boot: fork

### LINKED ISSUE #507: 3i: Handle handler behavior during fork()
By default, the child process should inherit the parent’s handler table after a `fork()`. This means that fork() should implicitly trigger a call to `cp_handler_table` at an appropriate point in the process creation flow. After the fork, grate/cage developers can still modify the inherited handler table as needed. For example, by using `register_handler` to add or remove specific handlers.

The current plan is to invoke `3i::cp_handler_table` inside rawposix, since we're doing the resource management of fork() there. This also ensures that handler tables are properly duplicated before the child begins execution.

Given the upcoming `lind-boot` implementation #510, circular imports shouldn't be expected to be a blocker for this approach.

Yaxuan-w:
Add fork implementation in lind-boot
Resolves #507 

**NOTICE:**

- Need to be merged with other lind-boot PRs
- Cannot pass ci/cd individually

### DIFF
--- a/src/rawposix/src/sys_calls.rs
+++ b/src/rawposix/src/sys_calls.rs
@@ -1,11 +1,10 @@
 //! System syscalls implementation
 //!
 //! This module contains all system calls that are being emulated/faked in Lind.
-use crate::fs_calls::kernel_close;
 use cage::memory::vmmap::{VmmapOps, *};
 use cage::signal::signal::{convert_signal_mask, lind_send_signal, signal_check_trigger};
 use cage::timer::IntervalTimer;
-use cage::{add_cage, cagetable_clear, cagetable_init, get_cage, remove_cage, Cage, Zombie};
+use cage::{add_cage, get_cage, remove_cage, Cage, Zombie};
 use dashmap::DashMap;
 use fdtables;
 use libc::sched_yield;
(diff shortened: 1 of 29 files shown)


src/rawposix/src/sys_calls.rs
```diff
@@ -68,36 +95,105 @@ pub fn fork_syscall(
...
+            UNUSED_ARG,
+            UNUSED_ID,
+            UNUSED_ARG,
+            UNUSED_ID,
+        );
+    }
 
-    add_cage(child_arg, cageobj);
-    0
+    // Delegate execution back to Wasmtime via 3i.
```
rennergade: comment about what the grateid thing is here would be helpful too

src/rawposix/src/sys_calls.rs
```diff
@@ -68,36 +95,105 @@ pub fn fork_syscall(
...
+    //
+    // Wasmtime will:
```
JustinCappos: This is true for now, but this comment would be stale if we add MPK, etc.  I don't think the code here will need to change.  So, maybe update the comment to make it more generic?

qianxichen233 (approved):

Issue #673, opened 2026-02-02
Syscall: (creat) Not Implemented in RawPOSIX


rishabhBudhouliya:
### Problem 

The [creat()](https://linux.die.net/man/2/creat) system call is not implemented in RawPOSIX. It currently returns a fake success (valid file descriptor) without actually creating the file on disk.

https://github.com/Lind-Project/lind-wasm/blob/e788eaf74d5d4dee9c0039fa0f024d4ce4871ca7/tests/unit-tests/file_tests/deterministic/creat_access.c#L15

### Context
This was discovered while fixing the failing test: https://github.com/Lind-Project/lind-wasm/blob/e788eaf74d5d4dee9c0039fa0f024d4ce4871ca7/tests/unit-tests/file_tests/deterministic/creat_access.c#L15

rishabhBudhouliya:
As mentioned in the fix PR, this was happening due to: 
https://github.com/Lind-Project/lind-wasm/blob/e717f4a853406607af83a7ce7977383850d22ab8/src/glibc/sysdeps/unix/sysv/linux/creat.c#L30-L34

I spent a lot of time looking at the default implementation for creat.c but glibc uses above mentioned implementation of creat.c based on the chosen host. The creat.c is designed to call open but it was calling SYSCALL_CANCEL instead due to a hardcoded decision branch.

PR #704, opened 2026-02-06
fix: creat_access.c test failure

rishabhBudhouliya:
## Purpose
Remove the `SYSCALL_CANCEL` branch from `src/glibc/sysdeps/unix/sysv/linux/creat.c` to always use `__open()`:

**Initial Cause To Investigate**
Test `tests/unit-tests/file_tests/deterministic/creat_access.c` fails because `creat()` returns fd=0 without creating the file, causing subsequent `access()` to fail with ENOENT.

## Root Cause
https://github.com/Lind-Project/lind-wasm/blob/e717f4a853406607af83a7ce7977383850d22ab8/src/glibc/sysdeps/unix/sysv/linux/creat.c#L30-L34

`src/glibc/sysdeps/unix/sysv/linux/creat.c` takes the `#ifdef __NR_creat` branch because:
- Build target is `--host=i686-linux-gnu`
- `__NR_creat` is defined as 8 in `sysdeps/unix/sysv/linux/i386/arch-syscall.h:42`

## Issue
#673


qianxichen233:
Do we have creat syscall in rawposix? If so, I think use creat syscall is better, otherwise current fix looks good to me

PR #705, opened 2026-02-06
fix: mmaptest.c test failure

rishabhBudhouliya:
## Purpose
[mmaptest.c](https://github.com/Lind-Project/lind-wasm/blob/main/tests/unit-tests/memory_tests/deterministic/mmaptest.c) was failing due to

```
mmap failed as expected, errno: 22 (Invalid argument)
[DEBUG] [wasm run STDERR]
/tmp/wasmtest_artifacts_u3hlj_xv/memory_tests/deterministic/mmaptest.c:30: int main(void): Assertion `mmappedData != MAP_FAILED' failed.
Error: failed to run main module `/tmp/wasmtest_artifacts_u3hlj_xv/memory_tests/deterministic/mmaptest.cwasm`

Caused by:
    1: error while executing at wasm backtrace:
           0: 0x8fe8d - mmaptest.wasm!abort
           1: 0x2f817 - mmaptest.wasm!__assert_fail_base
           2: 0x2f9ec - mmaptest.wasm!__assert_fail
           3: 0x1282 - mmaptest.wasm!__original_main
           4:  0xd5f - mmaptest.wasm!_start
           5: 0x9cb5e - mmaptest.wasm!_start.command_export
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information
    2: wasm trap: wasm `unreachable` instruction executed
[ERROR] FAILURE: Lind Wasm Segmentation Failure
```

Focussing on the specific log line: `errno: 22 (Invalid argument)`, with a simple process of elimination, I was able to arrive at the conclusion that the flag `MAP_POPULATE` is an unsupported flag in our mmap implementation at lind-wasm. 

To confirm, I checked our mmap implementation:
https://github.com/Lind-Project/lind-wasm/blob/e717f4a853406607af83a7ce7977383850d22ab8/src/rawposix/src/fs_calls.rs#L701-L705

We explicitly reject a flag outside our list of allowed flags for mmap.

# Discussion
@qianxichen233 @rennergade 
 There can be a discussion on whether the flag `MAP_POPULATE` needs to be supported by our mmap implementation. 
Screenshot 2026-02-06 at 12 57 18 PM

I've personally used mmap for a database project and based on performance benchmarks, I am positive that this flag can help in reducing page cache misses which is quite important for a database (especially a high performing one). I suspect this doesn't come under our scope as of this moment but nevertheless a concern to bring up.

### DIFF
--- a/tests/unit-tests/memory_tests/deterministic/mmaptest.c
+++ b/tests/unit-tests/memory_tests/deterministic/mmaptest.c
@@ -19,7 +19,7 @@ int main(void) {
 	int fd = open(FILENAME, O_RDONLY, 0);
 	assert(fd != -1);
 	//Execute mmap
-	void* mmappedData = mmap(NULL, filesize, PROT_READ, MAP_PRIVATE, MAP_POPULATE, fd, 0);
+	void* mmappedData = mmap(NULL, filesize, PROT_READ, MAP_PRIVATE | MAP_POPULATE, fd, 0);
 	assert(mmappedData != MAP_FAILED);
 	//Write the mmapped data to stdout (= FD #1)
 	write(1, mmappedData, filesize);


rennergade:
We should probably support MAP_POPULATE. There's a lot of mmap flags and some of them are probably dangerous for us. I think for now besides adding MAP_POPULATE to allow we should lind_debug_panic instead of returning an errno if we receive a flag were not allowing, will make it easier to debug if we encounter anything else odd.

rennergade:
> @rennergade I checked the implementation of lind_debug_panic() and it panics. Wouldn't this crash the host code for passing an unsupported flag while using mmap that the user might not be aware of?

Right, yeah. At this point we're just at a research and development stage. These will eventually be removed if we go towards production but this makes it easier for us to figure out problems as we build this out.

src/rawposix/src/fs_calls.rs
```diff
@@ -701,7 +701,7 @@ pub fn mmap_syscall(
     let allowed_flags =
         MAP_FIXED as i32 | MAP_SHARED as i32 | MAP_PRIVATE as i32 | MAP_ANONYMOUS as i32;
     if flags & !allowed_flags != 0 {
-        return syscall_error(Errno::EINVAL, "mmap", "Unsupported mmap flags");
+        lind_debug_panic("Unsupported mmap flag detected! Only MAP_FIXED, MAP_SHARED, MAP_PRIVATE and MAP_ANONYMOUS allowed");
```
rennergade: I think it's worth adding map populate to allowed and adding it back to the test as well

PR #708, opened 2026-02-07
fix for malloc_printerr

### LINKED ISSUE #664: fix failure of  malloc_printerr()
`malloc_printerr()` is responsible for printing diagnostic error messages during the dynamic memory management process. This issue  first raised in #347, reports that no error message is displayed when a glibc crash occurs.

To reproduce the issue, I executed a test in Lind that double-frees a buffer allocated by `malloc()`, thereby triggering `malloc_printerr()`. As a result, execution terminates with a WASM unreachable instruction, which is the intended failure behavior, and a backtrace is produced. However, no error message is printed prior to termination:
```
Error: failed to run main module `malloc_trigger.cwasm`

Caused by:
    1: error while executing at wasm backtrace:
           0: 0x28d80 - malloc_trigger.wasm!abort
           1: 0x5002e - malloc_trigger.wasm!__libc_message_impl
           2: 0x1d598 - malloc_trigger.wasm!malloc_printerr
           3: 0x1df0e - malloc_trigger.wasm!_int_free
           4: 0x1da64 - malloc_trigger.wasm!__libc_free
           5:  0xf14 - malloc_trigger.wasm!__original_main
           6:  0xd7f - malloc_trigger.wasm!_start
           7: 0x9b282 - malloc_trigger.wasm!_start.command_export
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information
    2: wasm trap: wasm `unreachable` instruction executed
``` 

Internally, `malloc_printerr()` calls `__libc_message()`, which in turn calls `__libc_message_impl()`. The actual error message emission is performed via the `WRITEV_FOR_FATAL` macro, which uses the `writev()` syscall to print the message.

 `__libc_message_impl()` never reaches a successful invocation of `WRITEV_FOR_FATAL`. Examining the macro definition shows the following:
```
#ifndef WRITEV_FOR_FATAL
# define WRITEV_FOR_FATAL	writev_for_fatal
static bool
writev_for_fatal (int fd, const struct iovec *iov, size_t niov, size_t total)
{
  return TEMP_FAILURE_RETRY (__writev (fd, iov, niov)) == total;
}
#endif
```

`WRITEV_FOR_FATAL`  is also defined in the Linux specific implementation located at `src/glibc/sysdeps/unix/sysv/linux/libc_fatal.c`, where `WRITEV_FOR_FATAL` is defined before including the generic POSIX implementation:

```
#define WRITEV_FOR_FATAL	writev_for_fatal

#include <sysdeps/posix/libc_fatal.c>
```

drapl0n:
This PR resolves #664: failure of `malloc_printerr()`

### DIFF
--- a/src/glibc/sysdeps/unix/sysv/linux/libc_fatal.c
+++ b/src/glibc/sysdeps/unix/sysv/linux/libc_fatal.c
@@ -24,6 +24,8 @@
 static bool
 writev_for_fatal (int fd, const struct iovec *iov, size_t niov, size_t total)
 {
+  // replaced INTERNAL_SYSCALL_CALL with writev()
+  // to resolve malloc_printerr() failure
   ssize_t cnt;
   do
     cnt = writev(fd, iov, niov);


rennergade:
This seems fine for a quick workaround that gives us helpful debug info. We should put up an issue about SYSCALL_INTERNAL though in case thats causing other problems. Should we be disabling it? If so we should probably make it debug panic

vidyalakshmir:
@drapl0n Does this change fix the issue - does it print the diagnostic message in case of a crash. Also do we have a test case for this? How do we verify this?

vidyalakshmir:
> > Does this change fix the issue - does it print the diagnostic message in case of a crash. Also do we have a test case for this? How do we verify this?
> 
> Yes, it does print the error message. I just wrote a test that double frees a buffer:
> 
> ```
> #include <stdlib.h>
> #include <stdio.h>
> 
> int main(void)
> {
>     char *buf = malloc(16);
>     if (!buf) {
>         perror("malloc");
>         return 1;
>     }
> 
>     free(buf);
>     free(buf);
> 
>     return 0;
> }
> ```
> 
>  > There are multiple glibc functions that invokes INTERNAL_SYSCALL_CALL. It would be good to start an issue to look into this and decide if this is the right way to handle all those cases.
> 
> That's the reason I was unsure about submitting a quick fix for this.

Great! This is good for now. Will start another issue to look at INTERNAL_SYSCALL_CALL.

vidyalakshmir (approved):
Looks good!

vidyalakshmir (approved):

Yaxuan-w (approved):
Lgtm! Could you also mention in comments this is a quick fix by us too?

PR #711, opened 2026-02-07
Reorganize docs and add clamping/stacking paragraphs

rennergade:
Changes

Added a clear description of grate stacking, including the possible outcomes of interception (handle-and-pass, handle-only, no interception, and blocking).

Added documentation for grate clamping, explaining how ancestor grates interpose at handler registration time to enforce routing structure on descendants.

Separated concerns between 3i (mechanism) and grates (policy/services) by moving grate-specific behavior out of the 3i overview.

### DIFF
--- a/docs/internal/grates.md
+++ b/docs/internal/grates.md
@@ -31,7 +31,7 @@ In addition to inheritance across fork, an ancestor grate may modify the system 
 
 3i allows system call arguments to specify which cage owns a referenced buffer. This enables grates to safely inspect, modify, or forward memory arguments without unnecessary copying.
 
-For example, if Cage A calls `write` and passes a pointer to a buffer, a grate can explicitly reference that buffer as belonging to A. This allows the grate to examine or adjust the data before forwarding the call, without incorrectly accessing its own memory space.
+For example, if Cage A calls write and passes a pointer to a buffer, a grate can explicitly reference that buffer as belonging to A. This allows the grate to examine or adjust the data before forwarding the call. Without this, the pointer would be interpreted relative to the grate's own address space, which is incorrect — cages have separate address spaces, and A's pointer is only valid in A's context.
 
 
 ## Acting on behalf of other cages


docs/index.md
```diff
@@ -4,74 +4,55 @@ id: Overview
...
+RawPOSIX is a small, trusted runtime component that provides POSIX-compatible system call implementations on behalf of cages. It runs entirely within the unprivileged Lind process and is responsible for interacting with the host kernel.
 
 ## Components
 
-### Wasmtime (our caging technology)
-[Wasmtime](https://wasmtime.dev) is a fast and secure runtime for WebAssembly designed by the [Bytecode Alliance](bytecodealliance.org). Lind-wasm uses wasmtime as a runtime with added support for multi-processing via Asyncify.
+### Wasmtime
 
-### lind-glibc
+Lind-Wasm uses Wasmtime as the WebAssembly runtime responsible for executing cages. Wasmtime provides fast execution, memory isolation, and a well-defined embedding API.
```
Yaxuan-w: Worth including the org web for Wasmtime and bytecode as citation

docs/internal/3i.md
```diff
@@ -1,269 +1,100 @@
...
 
-3i (Three-I) provides a general means for [cages](doc/index.md#cage) (and [grates](doc/index.md#grates)) to make system calls and intercept system calls via a programmable system call table.  The goal is to enable complex functionality without modifying the [microvisor](doc/index.md#core-concepts) or increasing the microvisor's trusted computing base. 
+## Motivation
 
-### 1.1 Motivation
-
-In a traditional Linux environment, extending or intercepting system calls (e.g., adding a new filesystem, tracing, or filtering calls) requires kernel modifications or mechanisms such as ptrace, which incur heavy overhead and depend on kernel-level mediation.
+In a traditional Linux environment, extending or intercepting system calls (for example, adding a new filesystem, tracing, or filtering calls) requires kernel modifications or mechanisms such as `ptrace`, which incur heavy overhead and depend on kernel-level mediation.
```
JustinCappos: mention eBPF

docs/internal/grates.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Composition
+
+Grates are composable. A grate may itself have another grate beneath it that provides additional functionality. This mirrors the Unix philosophy of building complex behavior from small, composable components.
+
+In practice, grates are composed using two patterns: stacking and clamping.
+
+## Stacking
+
+Stacking is the most common form of grate composition. Grates are arranged in a linear chain, and system calls flow through them sequentially.
```
JustinCappos:
```suggestion
Stacking is the most common form of grate composition. Grates are arranged in a linear chain, and system calls flow through them sequentially.  This is analogous to how output flows through a Unix pipeline from one process's stdio to another's stdout.  Note that in a Unix pipeline, the output is often transformed in the process, including creating new emissions to stderr or other files.  Similarly, a grate may pass system calls below (e.g., strace), may transform them (e.g., a file encryption grate), omit them (e.g., seccomp), perform different calls (e.g., a network filesystem), etc. in whatever means it has been designed to do.
```

docs/internal/grates.md
```diff
@@ -0,0 +1,73 @@
...
+When a system call reaches the ancestor grate, it issues a new 3i call. The call is then routed according to the ancestor’s policy, which may direct it to:
+- a service grate
+- another grate further down the stack
+- or eventually to RawPOSIX via normal 3i routing
+
+Service grates do not receive system calls directly from applications; they are invoked only when explicitly chosen by the ancestor grate.
+
+## Acting on behalf of other cages
+
+A grate may perform system calls on behalf of another cage so that the system behaves as though the originating cage made the call. This is required for correct semantics in operations such as process creation and memory management.
```
JustinCappos: far too succinct

docs/internal/3i.md
```diff
@@ -1,269 +1,100 @@
...
 
-3i is designed as a **runtime-agnostic** interposition layer that can operate on top of a wide range of isolation backends (e.g., software sandboxes, hardware-assisted memory protection, etc.) to **execute arbitrary code**. Its core goal is to provide a uniform mechanism for inter-cage call routing -- including system call interception, syscall customization, and cross-cage RPC -- without being tied to any specific runtime.
+To achieve this, 3i exposes an abstraction that allows runtimes to register arbitrary entry and exit hooks, implemented as plain C-ABI function pointers. These hooks allow each backend to integrate its own cage-management requirements (such as switching execution contexts, updating thread-local state, or preparing runtime metadata) into the call path.
```
stupendoussuperpowers: Entry/exit "hooks" might be interpreted to be something similar to eBPF where you just have notification-like event handlers that can modify data. In 3i if we attach to a syscall we need to handle it end to end which is different

docs/internal/grates.md
```diff
@@ -0,0 +1,73 @@
...
+
+Grates allow policy and system services to be implemented outside the trusted runtime, without kernel modifications or special privileges.
+
+## Why grates exist
+
+3i enables cages to intercept system calls from other cages. Because writing such interception logic is lightweight and common, Lind gives these cages the special name “grates.”
+
+This design allows functionality such as logging, filtering, virtualization, and service implementation to be expressed as ordinary user-space programs.
+
+## Composition
```
stupendoussuperpowers: Composability maybe?

docs/internal/grates.md
```diff
@@ -0,0 +1,73 @@
...
+
+Grates are composable. A grate may itself have another grate beneath it that provides additional functionality. This mirrors the Unix philosophy of building complex behavior from small, composable components.
+
+In practice, grates are composed using two patterns: stacking and clamping.
+
+## Stacking
+
+Stacking is the most common form of grate composition. Grates are arranged in a linear chain, and system calls flow through them sequentially.  This is analogous to how output flows through a Unix pipeline from one process's stdio to another's stdout.  Note that in a Unix pipeline, the output is often transformed in the process, including creating new emissions to stderr or other files.  Similarly, a grate may pass system calls below (e.g., strace), may transform them (e.g., a file encryption grate), omit them (e.g., seccomp), perform different calls (e.g., a network filesystem), etc. in whatever means it has been designed to do.
```
stupendoussuperpowers: ~stdio~ stdin? And also probably from stdout to stdin instead of the other way around. 

We can probably also make these brief:

"may pass" -> log/observe
"omit" -> block
"perform different calls" -> modify

docs/internal/grates.md
```diff
@@ -0,0 +1,73 @@
...
+   This is common for virtualization or user-space service implementations.
+
+3. **Not intercept**  
+   The grate does not register a handler for the system call. The call bypasses the grate and is routed by 3i to the next handler.
+
+4. **Block**  
+   The grate intercepts the system call and denies it, returning an error without forwarding it further.  
+   This is analogous to policy enforcement mechanisms such as seccomp.
+
+Each grate makes its decision independently. The overall behavior emerges from their composition.
```
stupendoussuperpowers: "makes decisions" seems a bit off. Maybe something like "Each grate runs/is designed independently, but they can be stacked together to achieve the desired behaviour"

docs/internal/grates.md
stupendoussuperpowers: Agreed. Can probably state a simple example here for e.g. the chroot grate where it modifies the input and then makes FS calls on behalf of the child.

docs/index.md
```diff
@@ -4,74 +4,55 @@ id: Overview
...
 ### 3i Implementation
-The intra-process interposable interface (3i) enables secure and efficient cage communication with speed similar to a function call. It provides POSIX interfaces between cages with interposition capabilities, enabling fine-grained security and access control by supporting the construction of grates.
-
-## Frequent Q&A
-
-### How hard is it to use Lind-Wasm?
 
-Lind-Wasm aims to minimize application changes.  
-Most applications can run without source-level modifications and only need to be recompiled using `clang` with Lind-Wasm–related feature flags.
+The 3i implementation provides the core system call routing and interposition mechanism used by both application cages and grates.
```
stupendoussuperpowers: Tiniest nit: do we not want to specify the etymology of lind or the what 3i stands for?

docs/internal/grates.md
```diff
@@ -0,0 +1,63 @@
...
+
+By allowing a grate to specify the target cage for a system call, 3i preserves POSIX semantics while still enabling interposition.
+
+## Cross-cage buffers
+
+3i allows system call arguments to specify which cage owns a referenced buffer. This enables grates to safely inspect, modify, or forward memory arguments without unnecessary copying.
+
+## Inheritance properties
+
+- A child inherits its parent’s system call handlers on fork.
```
JustinCappos: The rationale / impact needs to be explained

docs/internal/grates.md
```diff
@@ -0,0 +1,63 @@
...
+- An ancestor may modify the system call tables of its descendants and perform calls on their behalf.
```
JustinCappos: Why / how this is implemented, etc. needs to be explained

docs/internal/wasmtime.md
```diff
@@ -9,3 +9,144 @@ Wasmtime can run WebAssembly modules that follow the WASI standard, providing a
...
+#### 1. Process-like Operations (`fork`, `exec`, `exit`)
+
+Operations such as fork, exec, and exit require Wasmtime instances to be created, cloned, or destroyed. However, the logic that performs the semantic handling of these operations may execute in a different cage or grate than the one that originally issued the system call.
+
+After RawPOSIX completes the semantic work, control must return to Wasmtime, not necessarily to the instance that initiated the call.
+As a result, lind-3i cannot rely on an implicit “current” runtime state. Instead, it must explicitly retrieve the correct execution context:
+
+For `fork`, `exec`, and `exit`, these operations conceptually create, replace, or terminate the execution state of a process. After RawPOSIX completes the semantic handling, lind-wasm must resume execution in a Wasmtime instance associated with an *arbitrary* `cage_id`, which may differ from the calling cage. The appropriate runtime context is therefore retrieved by directly looking up the execution context associated with the target `cage_id`.
+
+#### 2. Thread-like Operations
```
JustinCappos: Why is this "thread-like"?

docs/internal/wasmtime.md
```diff
@@ -9,3 +9,144 @@ Wasmtime can run WebAssembly modules that follow the WASI standard, providing a
...
+
+Lind-Wasm maintains two global, per-cage pools of `VMContext` pointers:
+
+#### 1. General execution context lookup table
+
+The global `VMCTX_QUEUES` structure primarily manages execution contexts for the *main thread* (`tid = 1`) of each cage and indexed by `cage_id`. Each cage owns a FIFO queue that stores the execution contexts currently available to it. The total number of cages is fixed at startup (`MAX_CAGEID`).
+
+Importantly, table slots are never removed from the global pool. Instead, the *slot contents (the queue)* of each slot may be inserted or removed over time. When a cage terminates, its corresponding queue slot remains present, but its slot content is cleared and set to `None`.
+
+This design ensures that each table index always directly corresponds to a `cage_id`, eliminating the need for dynamic index management or lookup structures. As a result, `cage_id` can be used as a stable, constant-time index into the pool, reducing lookup overhead and avoiding additional search or indirection costs.
```
JustinCappos: Is this the only reason?  If we perform one more lookup, does that really matter so much?  Why is this used often?

docs/internal/wasmtime.md
```diff
@@ -9,3 +9,144 @@ Wasmtime can run WebAssembly modules that follow the WASI standard, providing a
...
+
+To support thread-related operations, lind-wasm maintains a separate, thread-specific execution context table. This table is used *only* for non-main threads (`tid != 1`) and exists to support thread-related syscalls and thread `exit`. Each `(cage_id, tid)` maps to at most one `VMContext`. No pooling is performed. This table is not consulted for normal execution or grate calls.
+
+```rust
+static VMCTX_THREADS: OnceLock<Vec<Mutex<HashMap<u64, VmCtxWrapper>>>>;
+```
+
+---
+
+### Execution Flow
```
JustinCappos: Maybe needs some high level context.

docs/internal/clamping.md
```diff
@@ -0,0 +1,151 @@
+# Clamping
+
+Clamping is a composition mechanism that allows a grate to selectively route syscalls to other grates based on some condition. Rather than all calls flowing through every grate in the stack unconditionally (as with stacking), a clamping grate evaluates a routing rule and only sends matching calls through the clamped grates. Non-matching calls bypass them entirely.
+
+The routing condition is up to the clamping grate. It could be based on path prefix, syscall type, file descriptor properties, cage identity, or anything else the clamping grate can inspect at dispatch time. The mechanism is the same regardless of the condition.
```
JustinCappos: Does it need to be at dispatch time?

docs/internal/clamping.md
```diff
@@ -0,0 +1,151 @@
...
+
+A useful way to read these is as if/endif blocks. The command line reads bottom-to-top as a stack, with the application on top. Note that the if condition's arguments are specified before the `%{` (on the command line), not after the `%}` as you might expect from traditional if/then syntax:
+
+`namespace-grate --prefix /tmp %{ imfs-grate %} python` reads as:
+
+```
+python
+if --prefix /tmp
+    imfs-grate
+endif
```
JustinCappos: Perhaps show you'd write it as this and then translate it over to what you have listed.

python
%} 
imfs-grate 
namespace-grate --prefix /tmp %{

docs/internal/clamping.md
```diff
@@ -0,0 +1,151 @@
...
+
+The `%}` is literally part of the command line that gets passed down through execs. When a cage eventually tries to exec `%}`, the clamping grate (which is interposing on exec) intercepts it, strips the `%}`, and rewrites the exec to whatever comes after. The cage that was about to exec `%}` now execs the remaining command line instead.
+
+Everything before the `%}` exec is inside the clamp. Everything after is above the clamp. Once the clamping grate sees the `%}`, it sets a flag to stop intercepting register_handler calls from subsequent descendants. Grates outside the delimiters register normally.
+
+If a clamped grate dynamically registers new handlers later, the clamping grate still intercepts those, since its register_handler interposition remains active for descendants of cages inside the clamp.
```
JustinCappos:
```suggestion
There is also a rare corner case that happens for grates that fork and exec new grates.  If a clamped grate dynamically registers new handlers later, the clamping grate still intercepts those, since its register_handler interposition remains active for descendants of cages inside the clamp.
```

docs/internal/grates.md
```diff
@@ -0,0 +1,111 @@
...
+```
+python
+if --prefix /tmp
+    imfs-grate
+endif
+```
+
+Clamping is made possible by interposing on 3i operations such as `register_handler` and `exec`. When a clamped grate attempts to register a handler for a system call, the clamping grate intercepts that registration, installs itself as the handler, and sets up a forwarding path to the clamped grate under an internal system call number. This ensures that the clamping grate remains in the routing path and can evaluate its condition before dispatching. Clamps can be nested, placed in series, or combined with unconditional stacking.
+
+The full mechanism, including command-line syntax, exec and register_handler interposition, fd table management, and worked examples, is described in [Clamping](clamping.md).
```
JustinCappos: I wonder if something like a tee grate (clamp?) and clamps that do if/else and case statements should also be included there.  Maybe the tee grate is a separate type because it isn't conditional?  I don't know.  It still wraps stacks in the same way as a clamp.

docs/internal/clamping.md
```diff
@@ -0,0 +1,151 @@
...
+
+## How it works
+
+A clamping grate interposes on `register_handler`, `exec`, `fork`, and `exit`.
+
+### exec interposition
+
+Both clamping and non-clamping grates follow the same basic pattern: consume arguments, fork a child, and exec the rest of the command line as the child's arguments. Grates don't parse or understand anything after their own arguments. They just pass it along. The only difference is that a clamping grate also interposes on register_handler, exec, fork, and exit to set up and manage its routing table.
+
+The `%}` is literally part of the command line that gets passed down through execs. When a cage eventually tries to exec `%}`, the clamping grate (which is interposing on exec) intercepts it, strips the `%}`, and rewrites the exec to whatever comes after. The cage that was about to exec `%}` now execs the remaining command line instead.
```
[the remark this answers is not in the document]
rennergade: I probably need to explain that part better.

rennergade:
@JustinCappos @stupendoussuperpowers @Yaxuan-w 

I tried to address the majority of your comments on the last commit. I omitted some things that would take more reorganization for a later PR. (breaking off a stacking section or possibly teeing, splitting out the execution section from wasmtime).

Sanchit had a question of the namespace by syscall being odd, @JustinCappos not sure if we want to change that here. Otherwise id like to merge this.

docs/internal/clamping.md
```diff
@@ -0,0 +1,151 @@
...
+Everything before the `%}` exec is inside the clamp. Everything after is above the clamp. Once the clamping grate sees the `%}`, it sets a flag to stop intercepting register_handler calls from subsequent descendants. Grates outside the delimiters register normally.
+
+There is also a rare corner case that happens for grates that fork and exec new grates.  If a clamped grate dynamically registers new handlers later, the clamping grate still intercepts those, since its register_handler interposition remains active for descendants of cages inside the clamp.
+
+### fork and exit interposition
+
+The clamping grate interposes on fork to track new cage IDs as they are created. When a cage forks, the clamping grate needs to know whether the new cage is a descendant of a clamped grate (inside the clamp) or a descendant of a cage above the clamp. This determines whether the clamping grate should intercept that cage's register_handler calls and apply routing to its syscalls.
+
+The clamping grate also interposes on exit to clean up state associated with a cage when it terminates.
+
```
vidyalakshmir: Duplicated section

PR #721, opened 2026-02-11
rawposix: allow ioctl(TIOCGWINSZ) passthrough

### LINKED ISSUE #662: ioctl 21523: TIOCGWINSZ
With coreutils now compiling, `ls` fails with lind debug panic on ioctl with code 21523 which is TIOCGWINSZ.

We never saw this in lind-nacl and I believe it just failed silently and isnt necessary for ls to run but panics in our case because of how were handling ioctl with lind debug panic.

I'm not sure if TIOCGWINSZ is something we want to implement? I believe there are some problems with it. My guess is that @JustinCappos thinks we should let it pass through and do blocking with grates which makes sense. In general it probably makes sense for us to shift to a full ioctl grate.

Bhanuu01:
Allow TIOCGWINSZ (21523) in RawPOSIX ioctl handling to prevent lind_debug_panic.  
Normalized requests are forwarded to the host kernel.  

Files changed:
- src/rawposix/src/fs_calls.rs
- src/sysdefs/src/constants/fs_const.rs

Fixes #662


rennergade (approved):
Thanks. approved!

PR #726, opened 2026-02-11
added deterministic test for getsockname

Bhanuu01:
added test for getsockname


Yaxuan-w (commented):
What's the difference between this and [the old one](https://github.com/Lind-Project/lind-wasm/blob/main/tests/unit-tests/networking_tests/deterministic/uds-getsockname.c)?

Issue #733, opened 2026-02-12
bug: munmap_syscall double address translation makes munmap a no-op


rishabhBudhouliya:
# Problem
munmap_syscall in RawPOSIX applies `PROT_NONE` at the wrong host address due to double address translation. 

# Root Cause
glibc first translates WASM pointer to host address
https://github.com/Lind-Project/lind-wasm/blob/97d5e2b9e71db90b7edee3a65cc0fe8035360822/src/glibc/sysdeps/unix/sysv/linux/munmap.c#L11-L13

Then we truncate the 64-bit host address and translate it again here: 
https://github.com/Lind-Project/lind-wasm/blob/97d5e2b9e71db90b7edee3a65cc0fe8035360822/src/rawposix/src/fs_calls.rs#L938

This bug was exposed while trying to fix https://github.com/Lind-Project/lind-wasm/blob/main/tests/unit-tests/memory_tests/fail/mmap-negative2.c where post munmap invocation, we do not encounter segmentation fault or any undefined behavior when we try to access already freed up memory address here: 
https://github.com/Lind-Project/lind-wasm/blob/97d5e2b9e71db90b7edee3a65cc0fe8035360822/tests/unit-tests/memory_tests/fail/mmap-negative2.c#L26-L31

PR #734, opened 2026-02-12
fix: munmap_syscall double address translation

rishabhBudhouliya:
# Purpose
 munmap_syscall  in RawPOSIX applies `PROT_NONE`  at the wrong host address due to a double address translation. This makes memory accessible after unmapping as seen after running the test [mmap-negative2.c](https://github.com/Lind-Project/lind-wasm/blob/main/tests/unit-tests/memory_tests/fail/mmap-negative2.c)
 
 # RCA
 The RCA was written in the issue: https://github.com/Lind-Project/lind-wasm/issues/733

# Test coverage
mmap-negative2.c (a previously skipped test) should work now, consequently proving munmap works.

### DIFF
--- a/skip_test_cases.txt
+++ b/skip_test_cases.txt
@@ -1,5 +1,4 @@
 file_tests/deterministic/popen.c
-memory_tests/fail/mmap-negative2.c
 networking_tests/deterministic/uds-nb-select.c
 networking_tests/deterministic/uds-serverclient.c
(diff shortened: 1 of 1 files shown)


rennergade:
> Should we move this test out of `fail/` folder as well?

so the fail tests are tests that are supposed to intentionally fail. It seems like this test revealed a bug though. 

@rishabhBudhouliya you need to take it off the skip list too

PR #758, opened 2026-02-18
Add cwasm support to lind-boot

### LINKED ISSUE #643: lind-boot: slow startup time due to wasm-only loading path

rennergade:
Closes #643 

  - Add --precompile flag to lind-boot for AOT-compiling .wasm → .cwasm
  - Add --allow-precompile flag to lind-boot for running .cwasm artifacts
  - Restore --precompile-only mode to lind_compile
  - Full lind_compile pipeline now produces .cwasm again: .c → .wasm → wasm-opt
   → .cwasm
  - lind_run passes --allow-precompile by default

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ sync-sysroot:
 test: prepare-lind-root
 	# NOTE: `grep` workaround required for lack of meaningful exit code in wasmtestreport.py
 	LIND_WASM_BASE=. LINDFS_ROOT=$(LINDFS_ROOT) \
-	./scripts/wasmtestreport.py && \
+	./scripts/wasmtestreport.py --allow-pre-compiled && \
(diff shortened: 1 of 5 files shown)


rennergade:
Putting this up as a draft. This probably will need some help discussion and help on reformatting to fit the tests again and configuring our scripts correctly.

Yaxuan-w (changes requested):
Awesome! The general implementation makes sense to me. Left a few suggestions.

PR #765, opened 2026-02-18
730 tests add dedicated python test runner for grate tests and create unified python test entry point

tashabits:
## ISSUE
- Grate testing needed to be included in the test running process, and a unified test runner created
## CHANGES
- Added a unified harness runner at `scripts/test_runner.py` that:
  - discovers harness modules under `scripts/harnesses/`,
  - runs selected harnesses through a common `run_harness(...)` interface,
  - writes per-harness artifacts to `reports/`, and
  - generates a combined `reports/report.html` aggregating all harness outputs.
- Added `--export-report` support to `scripts/test_runner.py` so CI can copy the combined report to `/report.html` for downstream PR/comment workflows.
- Added/updated harness modules in `scripts/harnesses/`:
  - `wasmtestreport.py` (standalone harness copy aligned to updated behavior),
  - `gratetestreport.py` (grate harness implementation with timeout support, strict success-on-exit-code-0 semantics, and detailed failure capture),
- Included grate harness behavior to align operationally with wasm flow:
  - prefers `scripts/lind_run` (with environment override support),
  - resolves generated `.wasm` outputs robustly,
  - emits explicit per-test `SUCCESS` / `FAILURE: ...` style logging,
  - uses `grates.json` as the report artifact name.
- Updated repo integration points to use the unified runner and expose new artifacts:
  - `Makefile` test target now uses `scripts/test_runner.py` and exports combined HTML,
  - `Docker/Dockerfile.e2e` updated for unified-runner script usage/artifacts,
  - `.github/workflows/e2e.yml` updated to upload/report unified artifacts (including combined HTML and harness JSON outputs).

## NOTES
- Timeout/slowness symptoms were observed before the unified runner migration and appear to be broader runtime behavior rather than caused solely by the harness orchestration layer.
- The combined report path is now `reports/report.html` (with optional export to `/report.html`), while per-harness files remain separately available for debugging.
- Grate harness success criteria is intentionally strict: only process exit code `0` is success; all non-zero exits are reported as failures with captured output.

### DIFF
--- a/skip_test_cases.txt
+++ b/skip_test_cases.txt
@@ -2,7 +2,7 @@ file_tests/deterministic/popen.c
 process_tests/deterministic/exit_failure.c
 networking_tests/deterministic/ipv6_basic.c
 networking_tests/deterministic/epoll_edge_triggered.c
-grate-tests/interpose-mmap.c
-grate-tests/interpose-mmap_grate.c
-grate-tests/simple-tests/cpdata.c
-grate-tests/simple-tests/cpdata_grate.c
+interpose-mmap.c
+interpose-mmap_grate.c
+simple-tests/cpdata.c
+simple-tests/cpdata_grate.c


rennergade:
I'm confused since the output says failure but the runner reports success?

rennergade:
@Yaxuan-w I thought about this a bit. Does using asserts in the tests instead of exit failure fix these problems?

Yaxuan-w:
> Cool if we change all of these to use asserts we should be able to integrate this right? Feels like that will be a huge help

Yes

Yaxuan-w (approved):

rennergade (approved):

Issue #774, opened 2026-02-19
Inconsistent argument semantics in register_handler and nondeterministic handler selection for duplicate mappings


Yaxuan-w:
There're two issues in the current register_handler implementation:

**(1) Argument semantic mismatch between register_handler and make_syscall**

The argument semantics in the `register_handler` signature are inconsistent with those used in `make_syscall`.  This mismatch did not surface earlier because all existing test cases happened to use `self_cage == 1`. Since the translated and untranslated values coincidentally aligned under that assumption, the issue remained hidden. 

For the fix, `make_syscall` should perform argument translation to adapt to the expectations of `register_handler`.

**(2) Nondeterministic handler selection when duplicate mappings exist**

The current design allows multiple mappings for the same `(cageid, syscall_num)` pair. When a regular glibc syscall is invoked, the `target_cageid` typically equals `self_cageid`. In this case, the current `register_handler` lookup logic returns the first matching entry found in the handler table. If multiple entries exist for the same `(cageid, syscall_num)` pair, the selected handler depends on how the `handler_table` is internally ordered. 

For the fix, 3i should refactor and add lookup logic to allow search by set `target_id`

Yaxuan-w:
Nondeterministic handler selection when duplicate mappings exist:

As discussion with @stupendoussuperpowers, beside adding lookup logic to allow search by (src_cageid, sysno, target_cageid)

1. fill the target_cageid to RAWPOSIX_CAGEID in MAKE_LEGACY_SYSCALL
2. the refactored handler_table will only have at most two entries per <cageid, sysno> (1 for grate call -- added by register_handler, 1 for rawposix -- always exists). If make_syscall in 3i receives a call with target_cageid=RAWPOSIX_CAGEID, it'll check the length of the entries. That means, if the length==1, it's a normal call and doesn't being handled by grates. Otherwise it would be handled by grates.

Yaxuan-w:
The current behavior is as follows:

In all cases except explicit removal, registration follows a create-or-overwrite semantic. The (srccage, targetcallnum) containers are created if they do not already exist, and the handler is inserted into the innermost map, replacing any previous handler registered for the same handlefunccage. If a rawposix handler already exists under the same (srccage, syscall) pair, it is removed before the new handler is inserted. This enforces the invariant that rawposix behaves strictly as a fallback dispatch target and must not shadow a more specific interposed grate handler. This invariant is enforced at registration time.

[...]

PR #780, opened 2026-02-20
Fix register_handler semantics in make_syscall

Yaxuan-w:
Fix the argument semantic mismatch between register_handler and make_syscall mentioned in #774


stupendoussuperpowers (approved):
LGTM.

PR #783, opened 2026-02-20
fix: instead of panic, return an empty CString

rishabhBudhouliya:
# Purpose
While testing an lmbench test: disk (that measures zone bandwidths and seek times), `path_conversion.rs` was crashing the entire runtime causing the test to fail. 

https://github.com/Lind-Project/lind-wasm/blob/fe5a9aac86829ea76adf86e691b67f0c848c3cb0/src/typemap/src/path_conversion.rs#L123-L125

# RCA
~~Lmbench's disk.c does a lseek until it fails and gets an EOF with -1 as a file descriptor. It's probing in chunks until read() fails. Due to our path_converstion panic behavior, we cause the entire runtime to crash and hence the test fails. Instead of the crash, the syscall should return an error.~~

https://github.com/Lind-Project/lind-wasm-apps/blob/fa2e16cd89b3bf894269787d38c425a742fe298b/lmbench/src/disk.c#L240

**Correction**
I had incorrectly assumed that the lseek + read probing loop was the cause of this error. In fact, `disk.c` was run without proper arguments, causing `av[1]` to be NULL. The call `open(NULL, 0)` passed a null pointer to
 `sc_convert_path_to_host`, which panicked instead of returning an error. Strange turn of events but we needed this fix not for the reason which I mentioned earlier. 

The true cause of the bug was path_conversion utility not reporting errors correctly but the trigger of the bug was not passing the right arguments to the [disk.c ](https://github.com/Lind-Project/lind-wasm-apps/blob/main/lmbench/src/disk.c) lmbench test which triggered a NULL argument invocation of open that caused the panic.

# Test Coverage
 Added a test: tests/unit-tests/file_tests/deterministic/path_conversion_safety

### DIFF
--- a/src/typemap/src/path_conversion.rs
+++ b/src/typemap/src/path_conversion.rs
@@ -116,7 +116,7 @@ pub fn sc_convert_path_to_host(
     path_arg: u64,
     path_arg_cageid: u64,
     cageid: u64,
-) -> Result<CString, i32> {
+) -> Result<CString, Errno> {
     #[cfg(feature = "secure")]
     {
         if !validate_cageid(path_arg_cageid, cageid) {
@@ -126,20 +126,20 @@ pub fn sc_convert_path_to_host(
 
     let path = match get_cstr(path_arg) {
         Ok(path) => path,
-        Err(e) => return Err(Errno::EFAULT as i32),
+        Err(_) => return Err(Errno::EFAULT),
     };
     // We will create a new variable in host process to handle the path value
     let relpath = normpath(convpath(path), path_arg_cageid);
     let relative_path = match relpath.to_str() {
         Some(s) => s,
-        None => return Err(Errno::EINVAL as i32),
+        None => return Err(Errno::EINVAL),
     };
 
     // Check if exceeds the max path
     #[cfg(feature = "secure")]
     {
         if relative_path.len() >= PATH_MAX {
-            return Err(Errno::ENAMETOOLONG as i32);
+            return Err(Errno::ENAMETOOLONG);
         }
     }
 
@@ -148,6 +148,6 @@ pub fn sc_convert_path_to_host(
     let full_path = relative_path.to_string();
     match CString::new(full_path) {
         Ok(c_path) => Ok(c_path),
-        Err(_) => return Err(Errno::EINVAL as i32),
+        Err(_) => return Err(Errno::EINVAL),
     }
 }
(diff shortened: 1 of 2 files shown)


rennergade:
Hmm  spitballing here...

Right now, if this fails, we just return CString::new("").unwrap(), which means the caller quietly passes "" to libc.

That’s misleading. On Linux, open("") returns ENOENT, but a NULL path should be EFAULT, and a too-long path should be ENAMETOOLONG. By swapping in an empty string, we lose that distinction. The caller can’t tell “conversion failed” from “the guest actually passed an empty string.”

The right fix is to return something like Result<CString, i32> so we can propagate the correct errno. That’s a bigger change, but it keeps the semantics honest. At the very least, we should signal failure somehow instead of silently papering over it.

src/typemap/src/path_conversion.rs
```diff
@@ -111,36 +112,42 @@ pub fn get_cstr<'a>(arg: u64) -> Result<&'a str, i32> {
...
+        Some(s) => s,
+        None => return Err(Errno::EINVA as i32),
+    };
 
     // Check if exceeds the max path
     #[cfg(feature = "secure")]
     {
         if relative_path.len() >= PATH_MAX {
-            panic!("Path exceeds PATH_MAX (4096)");
+            return Err(libc::ENAMETOOLONG);
```
rennergade: still using libc

rennergade (commented):
some things still left, and needs to pass linter

src/typemap/src/path_conversion.rs
```diff
@@ -111,36 +111,42 @@ pub fn get_cstr<'a>(arg: u64) -> Result<&'a str, i32> {
...
         if !validate_cageid(path_arg_cageid, cageid) {
             panic!("Invalid Cage ID");
         }
     }
-    let cage = get_cage(path_arg_cageid).unwrap();
 
     let path = match get_cstr(path_arg) {
         Ok(path) => path,
-        Err(e) => panic!("{:?}", e),
+        Err(e) => return Err(libc::EFAULT),
```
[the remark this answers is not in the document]
rishabhBudhouliya: thanks for catching that, I looked at the sysdef error constants and found all the relevant error numbers.

src/typemap/src/path_conversion.rs
rennergade: nit but can you remove pub use libc::*; from this file, i dont think we need it anymore

Yaxuan-w (approved):

rennergade (approved):

PR #789, opened 2026-02-21
fix clone3 cageid overwrite bug

rennergade:
clone3's child return path unconditionally zeroed __lind_cageid and __lind_base in shared linear memory before re-querying the host. 

For threads (CLONE_VM), this raced with parent/sibling  threads reading __lind_cageid on every syscall, causing 3i panics: "handler table for cage 0 not found". Guard the reset with !(flags & CLONE_VM) so threads skip the zeroing. Fork still  reinitializes as before.


stupendoussuperpowers (approved):
LGTM.

qianxichen233 (approved):

PR #790, opened 2026-02-21
Fix cxa exit

rennergade:
glibc's __cxa_atexit registers functions with signature void (*)(void *) (per the Itanium C++ ABI), but internally casts and stores them as void (*)(void *, int), calling them with an  extra status argument. 

In native C, the extra argument is harmlessly ignored due to calling convention flexibility. In WASM, indirect calls through the function table are strictly type-checked — the signature mismatch causes a trap: "indirect call type mismatch".
                                                                                                                                                                                              
This caused any program that registered __cxa_atexit handlers (most non-trivial C/C++ programs) to trap inside _run_exit_handlers during exit(). The trap was silently swallowed by  invoke_func (let _ = func.call(...)), making it appear as a hang — load_main_module returned without the exit syscall ever firing, so lind_manager.wait() blocked forever.                  
                                                                                                    
Note: atexit() was already fixed for this same issue (@qianxichen233 ), but __cxa_atexit was missed.     

This fixes deadlocks in git and curl

### DIFF
--- a/src/glibc/stdlib/cxa_atexit.c
+++ b/src/glibc/stdlib/cxa_atexit.c
@@ -50,7 +50,7 @@ __internal_atexit (void (*func) (void *), void *arg, void *d,
     }
 
   PTR_MANGLE (func);
-  new->func.cxa.fn = (void (*) (void *, int)) func;
+  new->func.cxa.fn = func;
   new->func.cxa.arg = arg;
   new->func.cxa.dso_handle = d;
   new->flavor = ef_cxa;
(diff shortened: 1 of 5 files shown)


stupendoussuperpowers (approved):
Changes make sense to me, bit skeptical about the test. 

I feel like we should have this test as a C++ file (eventually when we have that bandwidth), given that cxa_exit is prolly not supposed to be called manually like this.

rennergade:
> Changes make sense to me, bit skeptical about the test.
> 
> I feel like we should have this test as a C++ file (eventually when we have that bandwidth), given that cxa_exit is prolly not supposed to be called manually like this.

Yeah agreed, this is the best were going to do for now in terms of a unit tests though.

PR #791, opened 2026-02-21
Dynamic Loading

qianxichen233:
dynamic loading implementation

code cleanup hasn't fully finished yet, but lind-boot and glibc integration is done. This version is compatible with main and can run dynamic loaded programs

I will split this PR into two PRs later, one for dynamic loading implementation and one for related CI changes

### DIFF
--- a/scripts/append_tls_relocs_export.sh
+++ b/scripts/append_tls_relocs_export.sh
@@ -36,7 +36,7 @@ wat_in="$tmpdir/in.wat"
 wat_out="$tmpdir/out.wat"
 
 # 1) Convert wasm -> wat
-/home/lind/wabt-1.0.37/bin/wasm2wat --enable-all "$IN_WASM" -o "$wat_in"
+wasm2wat --enable-all "$IN_WASM" -o "$wat_in"
 
 # 2) If export already present, keep as-is
 if grep -Fqx "$INSERT_LINE" "$wat_in"; then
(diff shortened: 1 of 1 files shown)


src/wasmtime/crates/lind-dylink/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    U: Clone + Send + 'static + std::marker::Sync,
+>(
+    mut caller: &mut Caller<'_, T>,
+    file: i32,
+    mode: i32,
+    loader: DynamicLoader<T>
+) -> i32
+{
+    let base = get_memory_base(&mut caller);
+    let path = typemap::get_cstr(base + file as u64).unwrap();
```
rennergade: invalid pointer panics, same in dlsym

src/glibc/dlfcn/dlopen.c
```diff
@@ -23,6 +23,11 @@
 #include <ldsodefs.h>
 #include <shlib-compat.h>
 
+int __lind_dlopen(char* filename, int mode) __attribute__((
```
rennergade: supposed to be void*? same with dlsym

rennergade (changes requested):
Left a few bug comments.

I'm trying to ignore the problems with threading/multi-processing for now as well as the fact that this isn't portable to support static modules anymore.

PR #793, opened 2026-02-21
WIP : Glibc fixes to fix build errors and undefined symbol errors

vidyalakshmir:
This PR does additional glibc fixes:

1. It adds a placeholder for libdl.a which is required for apps like python

2. Adds unwind_def object file for libc.a

3. Updated s_frexpl.c with latest glibc code to fix the build error. Essentially control flow always matched #else of the preprocessing directive which was causing build error for the file. 

4. Compiler flags -mno-sse, -mno-mmx and -mfpmath-387 were used to build files within elf/ for shared builds and this was compilation errors as these flags are not supported for wasm. Hence, removed these flags and all files within elf/ are successfully compiled and object files are created.

5. Undefined symbol error were observed for 'pow' and other functions while running tests with glibc shared build. WebAssembly does not support ELF-style symbol versioning (.symver). 
When compiling glibc shared objects (.os), the Clang Wasm backend silently drops public symbols (like `pow`) that are wrapped in these versioning macros. Furthermore, since Wasmtime does not use the 
standard Linux `ld.so` dynamic loader, glibc's internal PLT-bypassing and hidden visibility mechanisms are incompatible. This commit updates `include/libc-symbols.h` to:
- Strip ELF versioning in `SHARED` builds by redefining `symbol_version` and `default_symbol_version` to fallback to standard C `strong_alias` and `weak_alias` macros.
- Disable the `visibility("hidden")` attribute block for shared builds (acting as a hard guarantee alongside `CFLAGS="-DNO_HIDDEN"`).

These changes force the glibc shared build to emit standard, globally visible symbols, mirroring the behavior of static (.o) builds and allowing the Clang Wasm backend to correctly link the library.

### DIFF
--- a/src/glibc/sysdeps/i386/Makefile
+++ b/src/glibc/sysdeps/i386/Makefile
@@ -48,7 +48,8 @@ endif
 # the first 3 mm/xmm/ymm/zmm registers are used to pass vector parameters
 # which must be preserved.
 # With SSE disabled, ensure -fpmath is not set to use sse either.
-#rtld-CFLAGS += -mno-sse -mno-mmx -mfpmath=387
+# Flags -mno-sse -mno-mmx -mfpmath=387 has been removed as they
+# are not supported by wasm
 ifeq ($(subdir),elf)
 CFLAGS-.os += $(if $(filter $(@F),$(patsubst %,%.os,$(all-rtld-routines))),\
 		   $(rtld-CFLAGS))


src/glibc/sysdeps/i386/Makefile
```diff
@@ -48,7 +48,7 @@ endif
 # the first 3 mm/xmm/ymm/zmm registers are used to pass vector parameters
 # which must be preserved.
 # With SSE disabled, ensure -fpmath is not set to use sse either.
-rtld-CFLAGS += -mno-sse -mno-mmx -mfpmath=387
+#rtld-CFLAGS += -mno-sse -mno-mmx -mfpmath=387
```
rennergade: would prefer to remove rather than comment out? leave a comment that it was removed.

rennergade (approved):
Left a minor comment. Also the title of the PR would be better to be more descriptive.

Other than that looks good, approved.

qianxichen233 (approved):

PR #799, opened 2026-02-22
Fix printf in multi-threading

### LINKED ISSUE #129: Fix printf in multi-threading

rennergade:
Closes #129 

Fix multi-threaded stdio deadlock caused by `SINGLE_THREAD_P` being incorrectly defined as `(1)` in `sysdeps/generic/sysdep-cancel.h`.
                                                                                                                                                                                            
When two threads contend on a stdio lock (e.g. via `flockfile`/`funlockfile` or `printf`), the losing thread calls __lll_lock_wait_private` which sets the futex value to 2 and sleeps. The winning thread's `_IO_lock_unlock` (in `stdio-lock.h`) is supposed to call `lll_unlock` → `atomic_exchange_release` → `futex_wake` to wake the waiter. However, `_IO_lock_unlock` has a    
fast path guarded by `SINGLE_THREAD_P`: when true, it does a plain `lock = 0` store with no `futex_wake`, leaving the waiting thread stuck forever.
                                                                                                                                                                                            
The only `sysdep-cancel.h` in the include path was the generic fallback (`sysdeps/generic/sysdep-cancel.h`) which defines `SINGLE_THREAD_P (1)`. There was no Linux/nptl-specific override. This made the single-threaded fast path unconditional — the `lll_unlock` branch was dead code.

Note: a previous fix patched `SINGLE_THREAD_P` to `(0)` in `single-thread.h`, but that file is never included by `stdio-lock.h`'s include chain (`stdio-lock.h` → `lowlevellock.h` → `lowlevellock-futex.h` → `sysdep-cancel.h`), so it had no effect on stdio locking.

### DIFF
--- a/src/glibc/sysdeps/generic/sysdep-cancel.h
+++ b/src/glibc/sysdeps/generic/sysdep-cancel.h
@@ -1,7 +1,7 @@
 #include <sysdep.h>
 
 /* No multi-thread handling enabled.  */
-#define SINGLE_THREAD_P (1)
+#define SINGLE_THREAD_P (0)
 #define RTLD_SINGLE_THREAD_P (1)
 #define LIBC_CANCEL_ASYNC()	0 /* Just a dummy value.  */
 #define LIBC_CANCEL_RESET(val)	((void)(val)) /* Nothing, but evaluate it.  */


JustinCappos:
Did you mean to close this PR?  Seems like a well understood problem.

rennergade:
yeah re-opened here on a clean branch: https://github.com/Lind-Project/lind-wasm/pull/800

PR #802, opened 2026-02-22
Correct store fault mechanism to report once

### LINKED ISSUE #784: bug: Forked cage store accumulates duplicate memory instances
All network lmbench benchmarks clients crash with `assert!(fault.is_none())` in [store.rs](https://github.com/Lind-Project/lind-wasm/blob/main/src/wasmtime/crates/wasmtime/src/runtime/store.rs) when the benchmp harness calls `fork()`. 

Root Cause
The child cage's wasmtime Store contains multiple memory instances that all report the same wasm_accessible() memory range.

Reproduction
Any lmbench benchmark that uses `benchmp()` will crash:
`lind_run /lmbench/wasm32-wasi/lat_tcp.opt.wasm 127.0.0.1`

with an error trace:
```
thread 'lind-fork-2' panicked at src/wasmtime/crates/wasmtime/src/runtime/store.rs:2297:
  assertion failed: fault.is_none()
```

rennergade:
- Fixed assertion panic in store.rs:2297 where multiple instances sharing the same linear memory (backup VMContext pool instances) caused assert!(fault.is_none()) to fire during WASM fault handling                                                                                                                                                                                   
- The signal handler now returns the first matching fault instead of panicking on duplicate shared-memory matches  

Closes #784

### DIFF
--- a/src/wasmtime/crates/wasmtime/src/runtime/store.rs
+++ b/src/wasmtime/crates/wasmtime/src/runtime/store.rs
@@ -2294,7 +2294,13 @@ impl StoreOpaque {
         let mut fault = None;
         for instance in self.instances.iter() {
             if let Some(f) = instance.handle.wasm_fault(addr) {
-                assert!(fault.is_none());
+                // Multiple instances may share the same linear memory (e.g.,
+                // backup VMContext pool instances import the same shared memory).
+                // Return the first match — all shared instances report identical
+                // fault info so subsequent matches can be skipped.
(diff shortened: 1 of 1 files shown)


Yaxuan-w:
Is it possible to have a test case for the previous panic?

rennergade:
> Is it possible to have a test case for the previous panic?

Not with our current setup because we can't compare to native.

Yaxuan-w (approved):

PR #805, opened 2026-02-22
Fix signal tests: sigpipe, sa_mask, and sigprocmask

rennergade:
This fixes three signal tests that are failing on the runner.

For sigpipe, we weren't sending the signal internally on a failed write. Implemented this behavior in rawposix.

For sa_mask and sigprocmask the tests were using sigpending and raise which we both dont implement, changed the behavior to not use those.

### DIFF
--- a/src/rawposix/src/fs_calls.rs
+++ b/src/rawposix/src/fs_calls.rs
@@ -16,7 +16,7 @@ use sysdefs::constants::fs_const::{
 };
 
 use sysdefs::constants::lind_platform_const::{FDKIND_KERNEL, MAXFD, UNUSED_ARG, UNUSED_ID};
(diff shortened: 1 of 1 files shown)


src/rawposix/src/fs_calls.rs
```diff
@@ -336,6 +337,10 @@ pub extern "C" fn write_syscall(
 
     if ret < 0 {
         let errno = get_errno();
+        // Linux delivers SIGPIPE before returning EPIPE on broken pipe writes
+        if errno == libc::EPIPE as i32 {
```
Yaxuan-w: Should we imports from sysdefs here?

Yaxuan-w (approved):

PR #815, opened 2026-02-24
Add support for scheduler calls

### LINKED ISSUE #523: Add support for scheduler calls
Apparently we need to support some scheduling calls for lmbench to build:

```
		#include <errno.h>
		#include <sched.h>
		int sched_get_priority_max(int policy) { (void)policy; errno = ENOTSUP; return -1; }
		int sched_setscheduler(pid_t pid, int policy, const struct sched_param *param) {
		  (void)pid; (void)policy; (void)param; errno = ENOTSUP; return -1;
		}
```

The current build script [here](https://github.com/Lind-Project/lind-wasm-apps/pull/18) just makes blank stubs. It probably is better that we don't have to add stubs to any app build, but I'm guessing we also just want to fake any scheduler calls for now.

Curious on thoughts from @JustinCappos and @qianxichen233

celinehoang177:
Add support for scheduler calls

Changes:
- `sched_setscheduler`, `sched_getscheduler`, `sched_get_priority_max`, `sched_get_priority_min`, `sched_getparam`, `sched_setparam` now call `lind_debug_panic()` before returning
- Fixed missing `weak_alias`/`libc_hidden_def` in `sched_getp.c` and `sched_setp.c` so their symbols are properly exported

Closes #523

### DIFF
--- a/src/glibc/sysdeps/unix/sysv/linux/i386/i686/sched_getp.c
+++ b/src/glibc/sysdeps/unix/sysv/linux/i386/i686/sched_getp.c
@@ -1,11 +1,15 @@
 #include <unistd.h>
 #include <sys/types.h>
 #include <sched.h>
+#include <lind_debug.h>
 
 int
 __sched_getparam (pid_t pid, struct sched_param *param)
 {
+  lind_debug_panic("sched_getparam called but not supported!");
   return -1;
 }
+libc_hidden_def (__sched_getparam)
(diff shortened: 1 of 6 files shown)


vidyalakshmir:
@Yaxuan-w I did not find any scheduler tests in our test suite.

Yaxuan-w:
Could we add some for them?

PR #817, opened 2026-02-24
Fix vmmap off by one and tls on fork bugs

### LINKED ISSUE #813: bug: fork_memory() ignores parent's current memory size
## RCA
After @rennergade 's [fix](https://github.com/Lind-Project/lind-wasm/pull/802) to unblock lmbench network tests, a deeper memory bug was exposed by the lmbench network tests: 
`lind_run /lmbench/wasm32-wasi/lat_tcp.opt.wasm 127.0.0.1`

The assertion panic was masking this underlying bug. When the test ran, the output was: 
```
lind@9d66bdc4863c:~/lind-wasm$ ./scripts/lind_run /lmbench/wasm32-wasi/lat_tcp.opt.wasm 127.0.0.1
Error: memory fault at wasm address 0x10000b3 in linear memory of size 0x100000000

Caused by:
    wasm trap: out of bounds memory access
```

This prompted me to investigate WASM module memory import declaration through
```
lind@9d66bdc4863c:~/lind-wasm$ wasm-objdump -x lindfs/lmbench/wasm32-wasi/lat_tcp.opt.wasm 2>&1 | grep -i "mem\|pages\|import.*memory\|data.*size"
 - func[4] sig=20 <__imported_lind_get_memory_base> <- lind.lind-get-memory-base
 - memory[0] pages: initial=4 max=1024 shared <- env.memory
```

I was under the suspicion that address `0x10000b3` > 256KB which might cause the out of bound memory access error. 

It seems like we have a bug here: 
https://github.com/Lind-Project/lind-wasm/blob/f54e07f49b3ae17cdf144999b5c3b4a8c5db57fb/src/wasmtime/crates/lind-multi-process/src/lib.rs#L1228-L1246

The size parameter is passed in as `parent_addr_len` but is completely ignored. This function creates a new SharedMemory using the module's original MemoryType which has initial 4 pages (each of 64KB size)

## Steps to reproduce
Assuming you have a docker environment with securesystemslab/lind-wasm-dev

1. cd ~/lind-wasm
2. git checkout store-fault
3. make lind-boot
4. make lmbench in lind-wasm-apps and copy it to lindfs
5. Run the lmbench test: ./scripts/lind_run /lmbench/wasm32-wasi/lat_tcp.opt.wasm 127.0.0.1
6. Should see the error: 
```
  Error: memory fault at wasm address 0x10000b3 in linear memory of size 0x100000000

  Caused by:
      wasm trap: out of bounds memory access
```

rennergade:
Closes #813 

- Fix three causes of memory faults after fork(): skip PROT_NONE regions in fork_vmmap, fix off-by-1 in vmmap overlap calculation, and track mprotect permission changes in vmmap so fork copies use correct protections                                                       
- Fix __tls_base WASM global not being restored for forked children — the child's __wasm_init_memory skips initialization (shared memory CAS flag already set), leaving __tls_base=0 and corrupting all TLS accesses (e.g. __ctype_b used by strtoul/isspace)                 
- Add fork_large_memory and fork_tls_ctype regression tests

### DIFF
--- a/scripts/cargo-lind_compile
+++ b/scripts/cargo-lind_compile
@@ -29,6 +29,7 @@ export CARGO_TARGET_WASM32_WASIP1_RUSTFLAGS="\
 -C link-arg=-Wl,--max-memory=67108864 \
 -C link-arg=-Wl,--export=__stack_pointer \
 -C link-arg=-Wl,--export=__stack_low \
+-C link-arg=-Wl,--export=__tls_base \
 -C link-arg=-Wl,--export=pass_fptr_to_wt"
 
 # Set wasm-opt and wasmtime paths (allow override via environment variables)


rennergade:
@rishabhBudhouliya can you verify this fixes the lmbench bug?

src/wasmtime/crates/wasmtime/src/runtime/instance.rs
```diff
@@ -362,10 +370,44 @@ impl Instance {
...
+        // The module's start function (__wasm_init_memory) uses an atomic flag in
+        // shared linear memory to ensure data-segment initialization runs exactly
+        // once. For the first instance it sets global[1] (__tls_base) to the
+        // address where .tdata was placed. For forked children the flag is already
+        // set (copied from parent), so __wasm_init_memory skips initialization and
+        // __tls_base stays at 0 (the module default). With __tls_base=0 every TLS
+        // variable access (e.g. __ctype_b used by strtoul/isspace) hits the wrong
+        // address, reading garbage pointers and causing spurious memory faults.
+        //
+        // global[1] is __tls_base by LLVM/wasm-ld convention for wasm32 shared-
```
qianxichen233: is there a reference link for this

rennergade: 1. LLVM commit implementing TLS: https://github.com/llvm/llvm-project/commit/42bba4b852b1a63db4043798bba7d9fcea61cbaf — the original implementation of __tls_base in wasm-ld                
2. WebAssembly tool-conventions Linking.md: https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md — documents the TLS segment and __tls_base convention

qianxichen233 (approved):

PR #828, opened 2026-02-25
Deprecate wasip1

### LINKED ISSUE #682: discussion: lind-wasm uses WASI-p1 context for argv/environ/stdio initialization
Although lind-wasm does not use wasi-libc for syscall handling (we rely on glibc / RawPOSIX instead), the runtime currently still attaches a WASI preview1 context (`wasi_common::WasiCtx`) and registers preview1 imports (`args_get`, `args_sizes_get`, `environ_get`, `environ_sizes_get`) during module instantiation.

Our wasm binaries are linked against a customized glibc, whose [startup path (`crt1.c`) still follows the WASI preview1 ABI for process metadata initialization](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/lind_syscall/crt1/crt1.c). During `_start`, glibc uses: (1) `__wasi_args_sizes_get` / `__wasi_args_get` to initialize `argc` / `argv`; (2) `__wasi_environ_sizes_get` / `__wasi_environ_get` to initialize the global `environ`. We also [populate stdin/stdout/stderr to WasiCtx](https://github.com/Lind-Project/lind-wasm/blob/f6351d0a17c59b5df6bf09470543dafda5444dea/src/wasmtime/src/commands/run.rs#L1206).

These functions are thin wrappers around imported `wasi_snapshot_preview1::{args_*, environ_*}` symbols. As a result, even without wasi-libc syscalls, the guest still expects WASI preview1 imports to exist for argument and environment setup. Without attaching a `WasiCtx` and populating it with argv/env data, glibc startup results in missing or NULL arguments (e.g., `argv[1] == NULL`), or instantiation failure due to unresolved imports.

I'm not sure whether we want to keep this dependency long-term.

rennergade:
Closes #682  
                                                                          
- Replaces the entire wasi-common crate (which registered ~60 unused WASI preview1 functions) with a minimal LindEnviron module (~130 LOC) that implements only the 4 functions our glibc crt1.c actually imports during _start()                                                                                                                                                     
- Renames the WASM import module from wasi_snapshot_preview1 to lind in both glibc and lind-boot, removing the last WASI naming dependency from the argv/environ path
- Fixes a bug where --env FOO=BAR was parsed by clap but never passed to the WASI context, so guest environment variables were always empty
- Removes 3 unnecessary dependencies from lind-boot: wasi-common, wiggle, cap-std

Our customized glibc's _start() calls 4 imported functions for argv/environ initialization:

_start()
→ __lind_args_sizes_get()      // get argc + total argv buffer size
→ __lind_args_get()            // write argv pointers + string data
→ __lind_environ_sizes_get()   // get env count + total buffer size
→ __lind_environ_get()         // write environ pointers + "KEY=VALUE\0" strings

Previously, these were imported under the wasi_snapshot_preview1 module name and served by the full wasi-common + wiggle stack, which registered ~60 other WASI functions (fd_read, fd_write, path_open, etc.) that are never called — all real syscalls go through glibc → lind-common → 3i → RawPOSIX.

### DIFF
--- a/src/wasmtime/crates/lind-common/src/lib.rs
+++ b/src/wasmtime/crates/lind-common/src/lib.rs
@@ -76,22 +76,14 @@ unsafe fn write_bytes(base: *mut u8, offset: usize, src: &[u8]) {
     }
 }
 
-/// Register all Lind host functions under the `"lind"` linker module.
-///
-/// This includes both the core syscall/debug/signal functions and the 4
-/// argv/environ functions that glibc's `_start()` calls to initialize
-/// `argc`, `argv`, and `environ` before entering `main()`.
-///
-/// The `get_environ` closure extracts the `LindEnviron` from the store data
-/// so that the argv/environ functions can read args and env vars.
-pub fn add_to_linker<
+/// Register the `make-syscall` host function: the unified syscall entry point
+/// from guest glibc into 3i.
+fn add_syscall_to_linker<
     T: LindHost<T, U> + Clone + Send + 'static + std::marker::Sync,
     U: Clone + Send + 'static + std::marker::Sync,
 >(
     linker: &mut wasmtime::Linker<T>,
-    get_environ: impl Fn(&T) -> &LindEnviron + Send + Sync + Copy + 'static,
 ) -> anyhow::Result<()> {
-    // attach make_syscall to wasmtime
     linker.func_wrap(
         "lind",
         "make-syscall",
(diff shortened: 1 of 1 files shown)


src/lind-boot/src/lind_wasmtime/environ.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+/// Write `src` bytes at `base + offset`.
+unsafe fn write_bytes(base: *mut u8, offset: usize, src: &[u8]) {
+    unsafe {
+        std::ptr::copy_nonoverlapping(src.as_ptr(), base.add(offset), src.len());
+    }
+}
+
+/// Register the 4 WASI preview1 functions under `wasi_snapshot_preview1`.
```
Yaxuan-w: We're not registering for wasi

Yaxuan-w (changes requested):

Yaxuan-w:
Could you merge main to see if the test can pass?

src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -191,5 +250,96 @@ pub fn add_to_linker<
             },
         )?;
     }
+
```
qianxichen233: The `add_to_linker` function here starts getting a little bit messy. Can we group these Lind host functions into seperate functions for clarity?

Issue #718, opened 2026-02-10
Performance Tracker


stupendoussuperpowers:
## Cage Round-Trip Times

Timers were set in the cage for these. e.g. - 

```c
ll start = get_time();
for (int i=0;i<1000000;i++) {
    function_call();
}
ll end = get_time(); // Averaged out.
```

Basic Tests:
| Test | Linux (ns) | Lind (ns) | Test File | 
|------|------------|---------|---|
| Syscall | 77.889 | 62.667 | [close.c](https://github.com/stupendoussuperpowers/lind-wasm/blob/benchmarks/benchmarks/close.c) |
| Regular Function Call | 691.6 | 2208.5 | [sfi-penalty.c](https://github.com/stupendoussuperpowers/lind-wasm/blob/benchmarks/benchmarks/sfi-penalty.c) | 
| `geteuid()` - RawPOSIX | 90.318 | 122.318 | geteuid.c |
| `geteuid()` - Grate | 90.318 | 358 | geteuid_grate.c |
-----

Test | Estimated |
|----|----|
`wasmtime` <-> WASM Round trip | 30-40ns

This time was estimated from calling a no-op function from wasmtime.

----

File System Tests:

| Test | Linux (ns) | Lind (ns) | IMFS Grate (ns) | IMFS on Lind (ns) |
|----|-------|------|------|-------|
| [Open and Close](https://github.com/stupendoussuperpowers/lind-wasm/blob/benchmarks/benchmarks/open-and-close.c) | 1236 | 3678 | 1696 | 365 |
| [Read [10 bytes]](https://github.com/stupendoussuperpowers/lind-wasm/blob/benchmarks/benchmarks/combined-read.c) | 230 | 381 | 722 | 16 |
| Read [4096 bytes] | 238 | 406 | 798 | 60 |
| [Write [10 bytes]](https://github.com/stupendoussuperpowers/lind-wasm/blob/benchmarks/benchmarks/combined-write.c) | 324 | 507 | 717 | 21 |
| Write [4096 bytes] | 339 | 529 | 789 | 77 |
-------

## Breakdowns 

How to read these tables:

Consider the snippet: 

```python
def A():
    B()
    C()

def C():
    D()
```

When benchmarking these functions, they'll be represented like so:

```
> A     | XXns
> > B   | XXns
> > C   | XXns
> > > D | XXns
```

These were tested by having timers at the start and end of these lines, which were then averaged out. e.g. 

```rust
let start = Instant::now();
[...]

PR #830, opened 2026-02-25
`tests/benchmarks`: writing and running cage-rtt benchmarks

stupendoussuperpowers:
The previous data measured and reported in #718 was done for the most part with a hacky and temporary script that needed manual intervention at various points. 

This PR adds all the tests used in the Performance Tracker (+ IPC tests) with a runner.

Changes:
- `tests/benchmark` folder which contains source code for benchmarking tests
- `benchrunner.py` to compile, run, and aggregate the results to either print to the terminal or store as a `json`
- `README.md` detailing how these tests are run, and how to add new tests


scripts/benchrunner.py
rennergade: - collect_tests globs * without filtering by .c suffix — will try to compile README.md, imfs.h, etc. and os._exit(1)                                                                        
- Bare except: catches KeyboardInterrupt/SystemExit — should be except Exception:

stupendoussuperpowers:
As discussed with @rennergade, instead of having grates duplicates in this benchmarks folder, I'll update the runner to just fetch them from the example grates repo like a quasi-package manager of sorts.

stupendoussuperpowers:
Updated this to use the example grates repo to source grates. Added a PR to example grates repo as well to make sure it's set up to be used like that. 

Also update the test runner script for some better error handling (e.g. skip if compilation/run fails, print debug logs, etc.)

scripts/benchrunner.py
rennergade: this always reports failure?

scripts/benchrunner.py
```diff
@@ -0,0 +1,336 @@
...
+
+            res[test][param][platform] = avg
+            res[test][param]["loops"] = loops
+    except Exception:
+        print("Invalid output from test: ", output.decode("utf-8"))
+
+
+def run_lind(wasm_paths, res, platform):
+    """Run lind-boot with one or more wasm paths."""
+    cmd = ["sudo", "lind-boot"] + wasm_paths
```
rennergade: probably worth calling lind_run?

rennergade (changes requested):
added some more comments.

probably want a .gitignore

Should think about factoring out the imfs code

rennergade (dismissed):

rennergade (dismissed):

stupendoussuperpowers:
> I tried to run the runner locally and it works well. I see there's a grates/ folder under benchmarks but that doesn't show on this PR.. I'm not sure what's happening here..

All grates are fetched from the lind-wasm-example-grates repo using a git shallow clone, the grates/ folder within the benchmarks folder is equivalent to `node_modules` with our package registry being the example grates repo. 

As previously discussed, this was done so that we don't need to manually make changes to a grate we might want to benchmark in two separate places i.e. once in the example-grates repo and then once in this benchmarks folder. It therefore should not be committed.

JustinCappos (commented):
Need to board a flight.  Will try to look more later

scripts/benchrunner.py
```diff
@@ -0,0 +1,411 @@
...
+import csv
+import logging
+import re
+import subprocess
+from pathlib import Path
+
+
+def repo_root() -> Path:
+    """Return repo root (scripts/..)."""
+    return Path(__file__).resolve().parent.parent
```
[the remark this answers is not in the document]
stupendoussuperpowers: Will change the doc string it's a bit confusing. 

It's supposed to mean => `benchrunner.py` is in the `scripts/` folder (`Path(__file__).parent`), the root of the lind-wasm repo is `scripts/..`

tests/benchmarks/fs_read.c
```diff
@@ -0,0 +1,51 @@
...
+	emit_result("Read", count, avg_time, loops);
+
+	close(fd);
+	free(buf);
+}
+
+int main(int argc, char *argv[]) {
+	int sizes[4] = {1, KiB(1), KiB(4), KiB(10)}; // MiB(1), MiB(10)};
+
+	char wchar = 'A';
```
JustinCappos: ??

tests/benchmarks/ipc_uds.c
```diff
@@ -0,0 +1,140 @@
...
+	free(buf);
+
+	char param[30] = "";
+	sprintf(param, "(STREAM, %d)", msg_size);
+	emit_result_string("Unix Domain Socket (RTT)", param,
+			   (end - start) / loops, loops);
+}
+
+int main() {
+	int sizes[] = {1, KiB(4), KiB(16), KiB(32)};
```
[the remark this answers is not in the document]
stupendoussuperpowers: Moved these arrays to the bench.h/.c file clubbed independently for fs_ and ipc_ tests. along with a const for the sizes.

scripts/benchrunner.py
```diff
@@ -0,0 +1,411 @@
...
+        )
+    # Pull latest changes.
+    run_cmd(["git", "-C", str(GRATES_REPO_DIR), "checkout", "main"])
+
+
+def resolve_grate_dir(grate_name: str) -> Path:
+    """Find a grate directory, preferring the external repo."""
+    ensure_grates_repo()
+    add_sparse_path(f"examples/{grate_name}")
+    repo_path = GRATES_EXAMPLES_DIR / grate_name
```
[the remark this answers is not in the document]
stupendoussuperpowers: repo_path will be `grates/imfs-grate` or `grates/geteuid-grate` etc, depending on which grate is being fetched. The only constant is the base path encoded in GRATES_EXAMPLES_DIR

stupendoussuperpowers:
Updated the `benchrunner.py` to be consistent with the new repo layout for `lind-wasm-example-grates`. 

Gave the existing tests a run to ensure they run fine (both the grate and non grate related tests).

Yaxuan-w (approved):
The failure test should be related to https://github.com/Lind-Project/lind-wasm/issues/1155, which is not related to this PR's changes, so approved

rennergade (approved):
Had approved before and changes look good.

PR #835, opened 2026-02-25
Adapt 3i glibc APIs to new make_syscall logic and make register_handler interposable

### LINKED ISSUE #821: 3i: fix copy_data_between_cages and copy_handler_table to adapt with new make_syscall lookup logic
Since I've refactored the make_syscall logic in PR #798 and set [copy_data_between_cages and copy_handler_table to use normal make_syscall routine on separate PRs](https://github.com/Lind-Project/lind-wasm/blob/ebb5abe0d9f51b168a1826802d4c93aad5acacca/src/glibc/lind_syscall/lind_syscall.c#L150), current main branch copy_data_between_cages and copy_handler_table will panic on `Handler not found` in 3i. This is caused by not adapting arguments semantics with new make_syscall search logic.

Fix would be adjust glibc arguments to adapt with new implementation.

Yaxuan-w:
Resolves #821 

1. Make copy_handler_table and copy_data_between_cages's glibc API adapt with new make_syscall lookup logic
2. Make register_handler inter posable
3. Add test

### DIFF
--- a/src/threei/src/handler_table/hashmap_impl.rs
+++ b/src/threei/src/handler_table/hashmap_impl.rs
@@ -257,7 +257,8 @@ pub fn register_handler_impl(
     // remove it to ensure it does not shadow the new handler.
     if target_map.contains_key(&lind_platform_const::RAWPOSIX_CAGEID) {
         target_map.remove(&lind_platform_const::RAWPOSIX_CAGEID);
-    } else if target_map.contains_key(&lind_platform_const::THREEI_CAGEID) {
+    }
+    if target_map.contains_key(&lind_platform_const::THREEI_CAGEID) {
         target_map.remove(&lind_platform_const::THREEI_CAGEID);
     }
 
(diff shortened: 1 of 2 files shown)


src/threei/src/handler_table/hashmap_impl.rs
```diff
@@ -250,10 +253,12 @@ pub fn register_handler_impl(
     let call_map = table.entry(srccage).or_insert_with(HashMap::new);
     let target_map = call_map.entry(targetcallnum).or_insert_with(HashMap::new);
 
-    // If a RAWPOSIX fallback handler exists for this (srccage, targetcallnum),
+    // If a RAWPOSIX/3i fallback handler exists for this (srccage, targetcallnum),
     // remove it to ensure it does not shadow the new handler.
     if target_map.contains_key(&lind_platform_const::RAWPOSIX_CAGEID) {
         target_map.remove(&lind_platform_const::RAWPOSIX_CAGEID);
+    } else if target_map.contains_key(&lind_platform_const::THREEI_CAGEID) {
```
rennergade: I think this is a bug?

This else if means if both a RAWPOSIX and THREEI handler exist for the same syscall, only RAWPOSIX gets removed. THREEI would still shadow the new grate handler in the _get_handler lookup (THREEI is checked before the grate fallback). The dashmap_impl.rs version correctly removes both unconditionally.

PR #846, opened 2026-02-27
`lind-perf`: baseline changes. add lind-perf crates, include basic feature plumbing.

stupendoussuperpowers:
`lind-perf` [base PR]

**Before merge:** PRs #847 and #848 have this PR as their base, so merge those into `lind-perf-init` before merging this branch to main.

--- 

- `lind-perf` implementation.
- Initial plumbing for lind-boot. (`--features lind_perf` + `--perf` flag)
- Initial benchmarking sites within `lind-boot`.

--- 

Changes:

`lind-perf` contains the crate responsible for timing and reporting micro-benchmarks. This crate only performs timing related operations when built with the `enabled` feature, otherwise every operation is a no-op. The dependency thus gets included regardless avoiding the excessive need for `cfg` flags in downstream crates.

The `enabled` feature is wired to `lind-boot`'s `lind_perf` feature.

The logic and design decisions behind `lind-perf` are explained in more detail in the `README.md` and int the comments.

| File | Changes |
|----|----|
|`lind-perf/disabled.rs` | contains the No-Op counters. |
|`lind-perf/timers.rs` | contains the `CLOCK_MONOTONIC_RAW` and the `RDTSC` timers. (Only invoked when `enabled`) |
|`lind-perf/enabled/` | contains the actual timing Counter and helpers for printing results to console. |

The following changes were made to ensure these features are visible to `lind-boot`:

| File | Changes |
| ---- | -------|
| `lind-boot/src/perf.rs` | accumulates, enables, and reports on counter defined within lind-boot and its dependencies. |
| `lind-boot/src/cli.rs` | adds the `--perf` flag to enable performance runs. |
| `lind-boot/src/main.rs` | update the execution logic to enable `--perf` runs. |
| `lind-boot/src/lind_wasmtime/trampoline.rs` | adding `lind-perf` based benchmark timers on a few hot paths. | 
| `Dockerfile.e2e` | ensure `lind-perf` is included in the build. | 

---

Example output from a performance run: 

```
--------------------------------------------LIND-BOOT--------------------------------------------
name                                                              calls        total          avg
-------------------------------------------------------------------------------------------------
lind_boot::grate_callback_trampoline                            1000000    295.412ms    295.000ns
lind_boot::typed_func_call                                      1000000     49.907ms     49.000ns


---------------------------------------------THREE-I---------------------------------------------
name                                                              calls        total          avg
-------------------------------------------------------------------------------------------------
threei::_call_grate_func                                        1000000    308.366ms    308.000ns
threei::make_syscall                                            1000000    362.457ms    362.000ns
```


src/lind-perf/src/enabled/report.rs
```diff
@@ -0,0 +1,86 @@
...
+    const NAME_W: usize = 60;
+    const CALLS_W: usize = 10;
+    const NUM_W: usize = 12;
+
+    let pad = "-";
+    let total = 97 - header.len();
+    let left = total / 2;
+    let right = total - left;
+
+    println!("\n{}{}{}", pad.repeat(left), header, pad.repeat(right),);
```
rennergade: youre mixing println and eprintln in this file, is that on purpose?

src/lind-perf/src/timers.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[inline(always)]
+pub fn read_end(_kind: TimerKind) -> u64 {
+    #[cfg(feature = "enabled")]
+    match _kind {
+        TimerKind::Rdtsc => rdtsc_end(),
+        TimerKind::Clock => clock_now(),
+    }
+
+    #[cfg(not(feature = "enabled"))]
```
rennergade: the intent here is confusing is this working correctly?

rennergade (commented):
a few questions, looks really good generally. very impressed!

src/lind-perf/src/timers.rs
stupendoussuperpowers: `lind-perf` if not `enabled` should always be a no-op so as to not add overhead to a regular build. I've consolidated all the `cfg`s for that within the public APIs itself. 

When built without the flag it'll just perform a no-op (and not actually go start/read a timer) like it did for the current CI runs. 

When built with the flag, it will actually do the work of figuring out which timer to read and then read them. 

This allows for crates using lind-perf to be agnostic of these implementation details and avoid them having to use `cfg`s or other ways to determine whether lind-perf is enabled or not.

PR #852, opened 2026-02-28
Snapshot and rollback cage's memory and threads in case of a failed execv on fork.

### LINKED ISSUE #754: failed exec causes panic on exit from child
## Issue

When running a fork-exec program, the child cage panics on exit if the exec fails with `ENOENT`

## How to reproduce

Run the [forkexecv.c](https://github.com/Lind-Project/lind-wasm/blob/22d59ed8b1d670cb1a05a72b2a44515dcdc33904/tests/unit-tests/process_tests/deterministic/forkexecv.c) without compiling or placing the dependent child binary.

## Rootcause

In `rawposix::exec_syscall`, we clear the child cage's epoch handler map: 

https://github.com/Lind-Project/lind-wasm/blob/22d59ed8b1d670cb1a05a72b2a44515dcdc33904/src/rawposix/src/sys_calls.rs#L265-L267

This `epoch_handler` is repopulated in [load_main_module](https://github.com/Lind-Project/lind-wasm/blob/22d59ed8b1d670cb1a05a72b2a44515dcdc33904/src/lind-boot/src/lind_wasmtime/execute.rs#L423) upon exec, but this code only runs if the child binary exists and can be loaded into wasmtime. In a `ENOENT` case, this function is never run and therefore the child cage's `epoch_handler` remains empty.

On exit, when calling `lind_thread_exit`, it's unable to clean up the exiting thread and panics since it cannot find this threadid in the `epoch_handler` map. 

https://github.com/Lind-Project/lind-wasm/blob/22d59ed8b1d670cb1a05a72b2a44515dcdc33904/src/cage/src/signal/signal.rs#L333-L335

## Possible fixes

In `lind_thread_exit`, skip the cleanup in cases of exec. 

Don't clear the `epoch_handler` table early.

stupendoussuperpowers:
Closes #754.

Need comments on whether this is the correct way to approach this fix. Posting this PR because this issue is blocking further work on `grate-rs` so want to get the ball rolling. 

Before calling exec_syscall, we store a Snapshot of the cage's thread/memory/signalhandlers before clearing them. In case the exec fails we restore this snapshot so that the cage can properly exit. 

Also added a test for fork-execv on a file that doesnt exist.


rennergade:
Hmm.

This feels a bit heavyweight IMO. We could check if the file exists before we clear the cage (even though this is technically a TOCTTOU) to error early without doing this. If the binary isn't there when we actually do our runtime exec at that point we should probably just harsh cage exit.

This is just my suggestion though we should probably come to a consensus.

PR #855, opened 2026-03-02
Replace hardcoded numbers in codebase with named constants (Rust + C)

### LINKED ISSUE #757: replace hardcoded numbers in codebase with named constants (Rust + C)
There are still several hardcoded numeric values across both the Rust and C sides of the codebase. These should be replaced with named constants to improve readability and maintainability.

### (1) Add missing platform-level constants
We should introduce constants (e.g., INIT_CAGE, etc.) in lind_platform_const.rs, and these will replace hardcoded numeric values currently used in rawposix.

### (2) Replace existing hardcoded numbers
We should replace Rust-side hardcoded numbers with already-defined constants, instead of duplicating numeric values inline in rawposix/3i/etc.

rawposix init file https://github.com/Lind-Project/lind-wasm/blob/main/src/rawposix/src/init.rs
constants file https://github.com/Lind-Project/lind-wasm/blob/main/src/sysdefs/src/constants/lind_platform_const.rs

### (1) Relocate constant definitions to avoid dependency issues
Currently, constants such as `NOTUSED` / `TRANSLATE_ERRNO_ON` / etc. are defined in [src/glibc/sysdeps/unix/syscall-template.h](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/sysdeps/unix/syscall-template.h). However, due to dependency constraints, they cannot be used inside [src/glibc/lind_syscall/lind_syscall.c](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/lind_syscall/lind_syscall.c).

We can move C-side constant definitions into (1) a dedicated header file and (2) place that file under src/glibc/lind_syscall/.

### Replace remaining magic numbers in lind_syscall.c
In lind_syscall.c, there are still hardcoded values such as 0 (in currently unused paths) /  -256 / conditions like translate_errno == 0. These should be replaced with named constants to clarify intent.

celinehoang177:
Closes #757

### Rust side
**New constants in lind_platform_const.rs:**
- INIT_CAGEID - cage ID for the bootstrap cage
- MAIN_THREADID - thread ID for the main thread

**Replaced hardcoded values in init.rs:**
- All hardcoded 1 used as the init cage ID → INIT_CAGEID
- Hardcoded 1 used as main thread ID in rawposix_shutdown() → MAIN_THREADID

### C side
- Created lind_constants.h - a header under lind_syscall/ containing NOTUSED, TRANSLATE_ERRNO_ON, TRANSLATE_ERRNO_OFF, and MAX_ERRNO. This resolves the dependency issue where lind_syscall.c could not include constants from syscall-template.h.
- Updated syscall-template.h to include lind_constants.h instead of defining those constants inline.

**Replaced magic numbers in lind_syscall.c:**
- translate_errno == 0 → translate_errno == TRANSLATE_ERRNO_OFF
- ret > -256 → ret > -MAX_ERRNO
- 0 /* translate_errno=0 ... */

### DIFF
--- a/src/glibc/lind_syscall/lind_syscall.c
+++ b/src/glibc/lind_syscall/lind_syscall.c
@@ -124,15 +124,15 @@ int register_handler (int64_t targetcage,
 {
     return make_threei_call(
         REGISTER_HANDLER_SYSCALL, 
-        0, // callname is not used in the trampoline, set to 0
+        NOTUSED, // callname is not used in the trampoline
         targetcage, // pass targetcage as self_cageid
         targetcage, // pass targetcage as target_cageid. Self_cageid and target_cageid are the same to adapt with regular make_syscall lookup logic in 3i
         targetcage, 
         targetcallnum, 
-        0, // runtime_id currently not used, set to 0
+        NOTUSED, // runtime_id currently not used
         this_grate_id, // handlefunccage is the grate id of the handler function, which is the same as this_grate_id
         in_grate_fn_ptr_u64,
-        0, 0, 0, 0, 0, 0, 0,
+        NOTUSED, NOTUSED, NOTUSED, NOTUSED, NOTUSED, NOTUSED, NOTUSED,
         TRANSLATE_ERRNO_OFF /* do not translate errno: return the raw result */
     );
 }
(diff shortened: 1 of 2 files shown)


celinehoang177:
I fixed it @vidyalakshmir

PR #856, opened 2026-03-02
Fix exit code mismatch between wasmtime and lind-boot exit

### LINKED ISSUE #771: program exits with non-zero status reports "exit normal"
As mentioned in #770, when a program exits with `EXIT_FAILURE` (non-zero exit status), lind-wasm still reports that the program "exited normally".

This is misleading because a non-zero exit status indicates failure, even though the process terminated normally via `exit()`, and causing #765 reports success in grate runner.

Yaxuan-w:
Resolves #771 

The previously error in #771 was caused by how lind-boot handled the return value from Wasmtime in main.

At the time, lind-boot did not interpret the exit code returned by the Wasmtime runtime. Instead, it only checked whether the call returned `Err` to determine failure. If the result was `Ok(_)`, the host process would always exit with status 0, regardless of the actual exit code produced by the Wasm program.

However, in lind-wasm, when a cage exits with a non-zero status, the exit code is propagated as `Ok(Val::I32(code))` rather than as an `Err`. As a result, non-zero exit codes from the Wasm side were being treated as successful execution by lind-boot, leading to the mismatch.

I have updated lind-boot to properly extract and propagate the exit code returned by lind-wasm. The exit code carried in `Val::I32` is now mapped directly to the host process exit status, ensuring that non-zero exits from the Wasm runtime are correctly reflected at the host level.

### DIFF
--- a/src/lind-boot/src/lind_wasmtime/execute.rs
+++ b/src/lind-boot/src/lind_wasmtime/execute.rs
@@ -46,7 +46,7 @@ use wasmtime_wasi_threads::WasiThreadsCtx;
 /// entrypoint. On successful completion it waits for all cages to exit before
 /// shutting down RawPOSIX, ensuring runtime-wide cleanup happens only after the
 /// last process terminates.
-pub fn execute_wasmtime(lindboot_cli: CliOptions) -> anyhow::Result<Vec<Val>> {
+pub fn execute_wasmtime(lindboot_cli: CliOptions) -> anyhow::Result<i32> {
     // -- Initialize Lind + RawPOSIX + 3i runtime --
     // Initialize the Lind cage counter
     let lind_manager = Arc::new(LindCageManager::new(0));


src/lind-boot/src/main.rs
```diff
@@ -71,9 +71,25 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    //
+    // At this layer, we intentionally do NOT interpret Wasm return
+    // conventions or runtime-specific details. All exit semantics
+    // (e.g., proc_exit, return values, traps) are already normalized
+    // inside `execute_wasmtime` into a single `i32` exit code.
+    //
+    // This design keeps the runtime backend pluggable (e.g., Wasmtime,
+    // MPK-based runtime, SGX-enclosed runtime) while preserving a
+    // consistent host process contract: exit(code) on success,
+    // exit(1) on unrecoverable runtime error.
```
qianxichen233: I don't think this comment makes sense. We are not distinguishing runtime error with cage exit code by using exit code of 1. A cage could also exit with exit code of 1

PR #885, opened 2026-03-05
Fix addr_translation init order

### LINKED ISSUE #883: move `__lind_init_addr_translation` before `__wasm_call_ctors`
I remember previously we would sometimes see some weird `clock_gettime` syscall invoke before `__lind_init_addr_translation` happens, this would panic threei since the cageid left uninitialized and it is 0. I am encountering the problem again when I was trying to run python again today with latest branch and container.

`thread 'main' (51410) panicked at /home/lind/lind-wasm/src/wasmtime/crates/threei/src/handler_table/hashmap_impl.rs:91:9:
No handler table for self_cageid=0 (syscall_num=228, dest_grateid=0)`

I looked into the issue. It was because `__wasm_call_ctors` functions. This function is a wasm constructor function injected by wasm-ld. This function is invoked even before crt1. It is normally an empty function and won't cause any problem. But sometimes there are content injected into this function. For example, python would inject a `_mi_process_init` function into `__wasm_call_ctors`. This function seems to be an initialization function from `mimalloc`. I havn't looked into how `__wasm_call_ctors` function works, but mimalloc is able to inject such a function into `__wasm_call_ctors` before the address initialization in crt1, and the clock_gettime syscall in `_mi_process_init` could panic the system.

We need to find a way to move `__lind_init_addr_translation` even earlier.

rennergade:
Closes #883 

                                                                                        
Summary                                                                                                                                                                              
                                                                                                                                                                                     
- lind-boot now explicitly calls __lind_init_addr_translation before _initialize (which runs __wasm_call_ctors), guaranteeing __lind_base and __lind_cageid are set before any WASM constructors execute
- Exported __lind_init_addr_translation from the linker via --export in lind_compile
- Removed the now-redundant call from _start() in crt1
- Added ctor_syscall_test.c which calls clock_gettime from a constructor to verify the fix

Context

__wasm_call_ctors runs all .init_array constructors before _start. Libraries like mimalloc inject constructors (e.g. _mi_process_init) that make syscalls during initialization. Previously __lind_init_addr_translation was called at the top of _start, so these constructors ran with __lind_cageid = 0, panicking 3i.

Moving initialization to the host side (before _initialize) is more robust than using constructor priorities, since it avoids any ambiguity about ordering between same-priority constructors.

### DIFF
--- a/src/glibc/lind_syscall/addr_translation.c
+++ b/src/glibc/lind_syscall/addr_translation.c
@@ -8,7 +8,13 @@ uint64_t __lind_base = 0ULL;
 // Cached cage id (pid), initialized on first call
 uint64_t __lind_cageid = 0ULL;
 
-void
+/* Run as a high-priority .init_array constructor so that syscalls from
+   user/library constructors (e.g. mimalloc's _mi_process_init) already
+   have __lind_base and __lind_cageid available.  Priority 101 runs
+   after memory.init (passive data segments) but before default-priority
+   constructors.  A host-side call before _start doesn't work because
+   memory.init resets the globals back to zero.  See issue #883.  */
+void __attribute__ ((constructor (101)))
 __lind_init_addr_translation (void)
 {
   if (__lind_base != 0ULL && __lind_cageid != 0ULL)
--- a/tests/unit-tests/process_tests/deterministic/ctor_syscall_test.c
+++ b/tests/unit-tests/process_tests/deterministic/ctor_syscall_test.c
@@ -1,0 +1,28 @@
+/* Test that syscalls work from constructors (__attribute__((constructor))).
+   This validates that __lind_init_addr_translation runs before user
+   constructors via .init_array priority ordering.  Without the fix,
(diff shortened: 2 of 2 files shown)


rennergade:
> I think this trick would work for 99.99% of time, but I am not sure what would happen if application also defined any priority-101 constructor which usees syscall.
> 
> If we can enforce this at wasm level/host level, that would be the best. For example, explicitly invoke `__lind_init_addr_translation` before invoking `_start` in host

I think the only way to do that is by adding an additional export and running it before libc start. Ill try that out.

rennergade:
> > I think this trick would work for 99.99% of time, but I am not sure what would happen if application also defined any priority-101 constructor which usees syscall.
> > If we can enforce this at wasm level/host level, that would be the best. For example, explicitly invoke `__lind_init_addr_translation` before invoking `_start` in host
> 
> I think the only way to do that is by adding an additional export and running it before libc start. Ill try that out.

OK that actually doesn't work. I don't think your concern is one that is ever going to matter though.

PR #887, opened 2026-03-06
Introduce temporary syscall number definitions under sysdefs/ to replace hardcoded values in Rust

### LINKED ISSUE #756: Introduce temporary syscall number definitions under sysdefs/ to replace hardcoded values in Rust
We have been discussing a unified solution for sharing constant definitions between C and Rust in #365, but there is currently no final agreement on the approach.

In the meantime, the Rust codebase urgently needs to replace hardcoded syscall numbers with named constants for maintainability and clarity.

As an interim solution, I propose to (1) add a dedicated syscall number definition file under sysdefs/ and (2) update the Rust side to eliminate hardcoded numeric values.

The change will involve wasmtime/lind-utils, rawposix/syscall_table.rs, and rawposix/init.rs

This does not preclude us from later introducing a fully unified constant-sharing mechanism across C and Rust. If/when we reach agreement on that, this structure can be refactored accordingly.

celinehoang177:
Closes #756. This PR introduces a unified syscall number definition file in the sysdefs library and updates the Rust codebase to eliminate hardcoded numeric values.

Changes
- Created syscall_const.rs with Linux x86_64 syscall constants.
- Updated syscall_table.rs to use named constants instead of raw integers.
- Updated lib.rs and other wasmtime modules to import syscall numbers from sysdefs.
- Removed lind_syscall_numbers.rs and reorganized its usage.

### DIFF
--- a/src/rawposix/src/syscall_table.rs
+++ b/src/rawposix/src/syscall_table.rs
@@ -149,6 +149,7 @@ pub const SYSCALL_TABLE: &[(u64, RawCallFunc)] = &[
         syscall_const::SYNC_FILE_RANGE_SYSCALL as u64,
         sync_file_range_syscall,
     ),
+    (syscall_const::ACCEPT4_SYSCALL as u64, accept4_syscall),
     (
         syscall_const::EPOLL_CREATE1_SYSCALL as u64,
         epoll_create1_syscall,
--- a/src/sysdefs/src/constants/syscall_const.rs
+++ b/src/sysdefs/src/constants/syscall_const.rs
@@ -96,6 +96,7 @@ pub const OPENAT_SYSCALL: i32 = 257;
 pub const UNLINKAT_SYSCALL: i32 = 263;
 pub const READLINKAT_SYSCALL: i32 = 267;
 pub const SYNC_FILE_RANGE_SYSCALL: i32 = 277;
+pub const ACCEPT4_SYSCALL: i32 = 288;
 pub const PRLIMIT64_SYSCALL: i32 = 302;
 pub const EPOLL_CREATE1_SYSCALL: i32 = 291;
 pub const DUP3_SYSCALL: i32 = 292;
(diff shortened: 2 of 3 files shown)


src/sysdefs/src/constants/syscall_const.rs
```diff
@@ -0,0 +1,99 @@
+//! Syscall number constants for the Lind platform.
```
JustinCappos: Was this manually copied over or is there a script, etc.?  Is the same information in glibc?  How does this + the kernel sync?  Is it so slow moving that this doesn't matter?

src/rawposix/src/syscall_table.rs
```diff
@@ -29,97 +28,127 @@ use super::sys_calls::{
...
-    (232, epoll_wait_syscall),
-    (233, epoll_ctl_syscall),
-    (263, unlinkat_syscall),
-    (267, readlinkat_syscall),
-    (277, sync_file_range_syscall),
-    (291, epoll_create1_syscall),
-    (292, dup3_syscall),
-    (293, pipe2_syscall),
-    (318, getrandom_syscall),
+    (syscall_const::READ_SYSCALL as u64, read_syscall),
```
JustinCappos: Maybe this makes sense.  Maybe it should be autogenerated.  I'm not sure.

Yaxuan-w (changes requested):
Also need to update exit/exec/fork in [rawposix/src/sys_call.rs](https://github.com/Lind-Project/lind-wasm/blob/33b4c200d3d3917fc9b1cdd30081cec6ecc65566/src/rawposix/src/sys_calls.rs#L186)

src/sysdefs/src/constants/syscall_const.rs
celinehoang177: These were manually copied from the Linux x86_64 syscall table. The same numbers are already in glibc's lind_syscall_num.h, so this file is basically the Rust equivalent of that. Since the Linux syscall table is append-only (existing numbers never get reassigned), sync drift isn't really a concern in practice. But yeah, I think auto-generating both the C header and this Rust file from a single source would definitely be cleaner and is a good follow-up to do!

PR #900, opened 2026-03-08
refactor handler resolution logic to remove usage of operational_cageid in rawposix

### LINKED ISSUE #896: determining operational cageid for syscalls.
Moving the [discussion](https://github.com/Lind-Project/lind-wasm-example-grates/pull/14#issuecomment-4017066045) from the grates repo to here. 

Currently, when a syscall execution hits rawposix, the way it determines the original calling cage changes based on individual syscalls, and the process of extracting this can lead to breakages in the semantic understanding of what the `make_threei_call` arguments represent. In most cases, this operational cageid is extracted from an arbitrary `(argX, argXcageid)` tuple.

Walking through this setup:

- We run `chroot-grate.wasm --chroot-dir /tmp program.wasm` 
- When `program.wasm` issues `open("/path", ...)`, it is intercepted by the `chroot-grate` which will change the path to `/tmp/path`. 
- This new path buffer (`/tmp/path`) now exists in the grate's memory, so we will now call `make_threei_call(..., new_path, grateid,...)`. 
- The current way `open_syscall` in rawposix handles things, the `operational_cageid` is determined by the `path_arg_cageid`, so in this scenario it would call `open` on behalf of the chroot-grate, and not `program.wasm`, and an FD will be opened for the grate.
https://github.com/Lind-Project/lind-wasm/blob/678448c365c2b4ed601efb4085cfc4939fa3428a/src/rawposix/src/fs_calls.rs#L100-L107

In the implementation that exists right now, the only safe way to modify `open`'s path buffer is to overwrite the existing argument, and this will only be safe in cases where the new_path length is equal to or less than the original path. 

We have the following possible fixes for this:

1. Use unused args to represent the operational cageid: For e.g. `(arg6, arg6cage)` when unused. Cons: This is a little hacky, and might lead to problems for syscalls that utilize all 6 arguments. It also breaks the contract of where `(argX, argXcage)` are only supposed to be used for address translations without any hidden meaning.

2. Change the semantics of how `(selfcageid, targetcageid)` work. Cons: This is a really complicated change.

3. For `make_threei_call` replace the unused `syscall_name` parameter to represent the operational cageid instead, and change rawposix syscall signatures to add an added parameter called `operational_cageid`. We would also need to change the signature of `pass_fptr_to_wt` so that grates know which cage it received this call from.

Yaxuan-w:
Resolves #896 

Update 3i handler resolution to lookup only according to (source_cageid, syscallnum)

## Changes in this PR 

1. Change 3i lookup logic to let target_cageid always represent "the cage that handles operation"
2. Remove rawposix operational cageid usage. Now the `cageid` arg in rawposix represent the cage that handles operation
3. Update grate tests to reflect this logic


rennergade (approved):
I don't have any concerns here this looks good to go.

PR #902, opened 2026-03-08
Restore wasip linker for rust

### LINKED ISSUE #901: wasip1 deprecation breaks rust cages.
After the #828 changes that removed wasip1 interfaces from sysroot, cages written in Rust encounter run-time linker issues. 

```bash
lind@0ceb86e49a08:~/lind-wasm/rgrate$ sudo lind-boot rgrate.wasm
Error: unknown import: `wasi_snapshot_preview1::args_sizes_get` has not been defined
```

`wasm-objdump`:

```bash
lind@0ceb86e49a08:~/lind-wasm/rgrate$ wasm-objdump -x ~/lind-wasm/lindfs/rgrate.wasm | grep args_sizes_get
 - func[0] sig=1 <wasi::lib_generated::wasi_snapshot_preview1::args_sizes_get> <- wasi_snapshot_preview1.args_sizes_get
 - func[1456] sig=4 <wasi::lib_generated::args_sizes_get>
 - func[1456] size=559 <wasi::lib_generated::args_sizes_get>
 - func[0] <wasi::lib_generated::wasi_snapshot_preview1::args_sizes_get>
 - func[1456] <wasi::lib_generated::args_sizes_get>
```

This is primarily because when compiling Rust to Lind, we use the [`wasm32-wasip1` target](https://github.com/Lind-Project/lind-wasm/blob/f77760a63b1668a4526f0c60c997c307c1e5cb67/scripts/cargo-lind_compile#L8), and [build `std` for this target](https://github.com/Lind-Project/lind-wasm/blob/f77760a63b1668a4526f0c60c997c307c1e5cb67/scripts/cargo-lind_compile#L40), which generates imports that are dependent on `wasip1`. 

Besides just helpers to fetch environment variables, wasip1 provides functions that we do not have lind wrappers for, e.g. :

```bash
lind@0ceb86e49a08:~/lind-wasm-example-grates/examples/chroot-grate-rs$ wasm-objdump -x ~/lind-wasm/lindfs/chroot-grate-rs.wasm | grep random_get
 - func[0] sig=1 <wasi::lib_generated::wasi_snapshot_preview1::random_get> <- wasi_snapshot_preview1.random_get
 - func[2121] sig=3 <wasi::lib_generated::random_get>
 - func[2121] size=390 <wasi::lib_generated::random_get>
 - func[0] <wasi::lib_generated::wasi_snapshot_preview1::random_get>
 - func[2121] <wasi::lib_generated::random_get>
```

We need further discussion on how to handle this, and whether we need to deprecate wasip1. 

cargo doesn't support fully stdlib that are not wasi-based. From the [Rust Book](https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-unknown-unknown.html#:~:text=The%20wasm32%2Dunknown%2Dunknown%20target%20has,wasm32%2Dwasip1%20or%20wasm32%2Dwasip2%20targets):

> The wasm32-unknown-unknown target has support for the Rust standard library but many parts of the standard library do not work and return errors. For example println! does nothing, std::fs always return errors, and std::thread::spawn will panic. There is no means by which this can be overridden. For a WebAssembly target that more fully supports the standard library see the [wasm32-wasip1](https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-wasip1.html) or [wasm32-wasip2](https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-wasip2.html) targets.

rennergade:
Add WASI snapshot preview1 compatibility for Rust programs                                                                           
                                                                                                                                     
After the wasip1 deprecation (#828), Rust programs compiled with wasm32-wasip1 fail at runtime because Rust's std imports            
args_sizes_get, args_get, and random_get from the wasi_snapshot_preview1 module, which is no longer registered in the linker.
                                                                                                                                     
This registers the 5 host functions that Rust std needs (args_sizes_get, args_get, environ_sizes_get, environ_get, random_get) under the "wasi_snapshot_preview1" module in addition to "lind". Also adds random_get under "lind" for completeness. C programs are unaffected — they continue to import from "lind" via glibc.

Closes #901

### DIFF
--- a/src/wasmtime/crates/lind-common/src/lib.rs
+++ b/src/wasmtime/crates/lind-common/src/lib.rs
@@ -218,20 +218,6 @@ fn add_runtime_to_linker<
         },
     )?;
 
-    linker.func_wrap(
-        "lind",
-        "random_get",
-        move |caller: Caller<'_, T>, buf: i32, buf_len: i32| -> i32 {
-            let base = get_memory_base(&caller) as *mut u8;
-            let slice = unsafe {
-                std::slice::from_raw_parts_mut(base.add(buf as usize), buf_len as usize)
-            };
-            let mut file = std::fs::File::open("/dev/urandom").unwrap();
-            std::io::Read::read_exact(&mut file, slice).unwrap();
-            0 // __WASI_ERRNO_SUCCESS
-        },
-    )?;
-
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


qianxichen233:
> > This registers the 5 host functions that Rust std needs (args_sizes_get, args_get, environ_sizes_get, environ_get, random_get)
> 
> Are we 100% sure that these are the only 5 functions that std needs? Or this is based on experimentation.
> 
> Also tagging @qianxichen233 here to ask: Is it possible for `wasm-opt` to replace something like `wasi_snapshot_preview1::args_sizes_get` in the `.wasm` binary to `lind::args_sizes_get`? I remember a similar change being implemented in #631. This might help reduce the redundant linker functions.

I don't think original `wasm-opt` has such flag. It should be possible to implement as our own custom pass though, as we are already maintaining our own `wasm-opt` anyway

Yaxuan-w:
I remembered in our previous implementation, we don't need to manually implement linker func for those wasip1 api..?

stupendoussuperpowers (approved):

PR #904, opened 2026-03-09
Add signal_callback flag to script

Yaxuan-w:
Export signal_callback when compiling cwasm/wasm binaries to enable signal features


rennergade:
I'm getting this now too

PR #910, opened 2026-03-09
Implement structured exit status handling and correct POSIX wait status encoding

Yaxuan-w:
Previously, cage termination did not consistently distinguish between normal exits (exit(code)) and signal-based termination. Additionally, exit codes were not explicitly encoded according to the POSIX wait status format. This could lead to incorrect behavior when applications use standard glibc macros such as: WIFEXITED, WEXITSTATUS, etc.

This PR:

1. Fix signal-triggered termination to handle exit code
2. Introduce a structured exit status representation
3. Encode wait status to match POSIX / glibc expectations

### DIFF
--- a/tests/unit-tests/signal_tests/deterministic/kill.c
+++ b/tests/unit-tests/signal_tests/deterministic/kill.c
@@ -13,7 +13,9 @@ int main(void) {
 
     if (pid == 0) {
         // child: wait for parent to kill us
-        sleep(10);
+        while (1) {
+            
+        }
         _exit(123);
     }
 


Yaxuan-w:
DNS failure is not related to this PR

rennergade:
Why is the kill test skipped here?

src/rawposix/src/sys_calls.rs
```diff
@@ -346,6 +347,9 @@ pub extern "C" fn exit_syscall(
     // in the cage (0 = no, 1 = yes).
     let mut is_last_thread = 0;
 
+    // Only the low 8 bits of the exit code are observable via wait()
+    let exit_st = ExitStatus::Exited(status & 0xff);
```
rennergade: The exit code is masked to 8 bits too early:
  let exit_st = ExitStatus::Exited(status & 0xff);

encode_wait_status already does code & 0xff. Double-masking isn't harmful but is confusing — better to store the raw code in ExitStatus::Exited(status) and let encode_wait_status handle truncation.

PR #913, opened 2026-03-10
Encode cageids to enable address translations: `(arg, cageid)` tuple independence from calling cages.

stupendoussuperpowers:
Building on the discussion here: https://github.com/Lind-Project/lind-wasm/issues/896#issuecomment-4020037624

We now allow:

1. Grates/cages to indicate to threei that the argument they are passing needs translation. 
2. Internally translate arguments on call to make_threei_call.
3. Ensure support for existing ways to use make_threei_call both in glibc and cages are still supported. 

How this works:

1. The MSB of cageid for each argument in make_threei_call, when set to 1, translate the address in the argument to be host-coded. 
2. Example can be seen in the tests attached, when a grate wants to translate an address, it's indicated by modifying the cageid MSB.
3. make_threei_call changes the cageid back to a a regular cageid so that existing behaviour is not broken.

--- 

Tests added try to pass a grate-local address to the make_threei_call invocation made on behalf of child, and ensure that copy_data_between_cages hasn't regressed:

```bash
lind@9a3542c81efe:~/lind-wasm$ sudo lind-boot diff-cage-args_grate.cwasm diff-cage-args.cwasm
[Grate|interpose-exec] Handling function ptr: 2 from cage: 1
[Grate|interpose-exec] In exec_grate 1 handler for cage: 1
[rawposix|open] path="/tmp/redirected.txt" oflag=0 mode=0
Hello world. FD=3
[Grate|interpose-exec] Handling function ptr: 3 from cage: 1
Goodbye world! ret=4321 buf=helloworld
[Grate|interpose-exec] PASS
```


stupendoussuperpowers:
> option 3 is the best for security since we should be doing those checks
anyways.  Is there any other reason it would be a problem to do them in 3?

We could run _checks_ at 3 through the various typemap helpers rawposix already uses but are currently no-ops. This should not be that big of a hurdle but might need some new negative unit tests.

But should the scope of these also be to try and convert the address to be host-coded? In which case, address translations are unnecessary at any stage.

> One downside which is perhaps why we did it like this is that in theory we
could have a 64 bit grate, etc. and doing the translation once at call time
avoids us needing to do it again and again for different grates.

Yeah I think this point was brought up in the original issue as well.

Yaxuan-w:
If I understand the previous discussion correctly, option 3 does not necessarily conflict with the current glibc-based translation approach. It can support both fixed and non-fixed addresses. In practice, it just adds an additional option where grate developers do not need to perform address translation in userspace before invoking the syscall.

Regarding performance, one possible approach could be to add a conditional compile flag. If a user prioritizes performance, we could disable the RawPOSIX-side checks and assume that all addresses have already been fixed before entering RawPOSIX.

From my perspective, based on what I understand about this PR, the work is important for supporting the chroot grate. For the current stage, the lowest-cost and fastest path to getting this working might be to perform the address translation in glibc so that it matches the existing RawPOSIX address checks. At the same time, we could document the design discussion and potential improvements in an issue and revisit it later (for example after the paper submission), when we have more time to refine the design.

As for where exactly the address translation should live if we keep it in glibc, I don't have a strong preference. From the discussion so far, one possible option would also be to let grate developers perform the translation before invoking the syscall. This might also help when testing new RawPOSIX functionality, since it makes the address assumptions explicit.

stupendoussuperpowers:
For now, we can do either of these, and then pick something more robust later:

1. Expose `lind_get_memory_base` to userspace, let grates convert their own addresses. 
2. This PR which lets glibc handle it, once indicated.

Both these are low effort changes (a few lines at best), and would enable chroot and similar grates to work properly. Let me know what the best approach is, and I can submit a PR for that.

For the longer run:

We need to decide a strict boundary on who is responsible for address translations, who does what checks, and why. Both the above stated approaches are inconsistent at best (grates _must_ translate/indicate an address for make_threei_call, but _mustn't_ for copy_data), and confusing for grate authors at worst (rawposix will crash on invalid address instead of reporting a proper error). My suggestion is that rawposix must receive only one kind of address, either user or host, having to support both at the same time will lead to a lot of unnecessary address-coercion codepaths.

src/glibc/lind_syscall/addr_translation.h
```diff
@@ -5,6 +5,11 @@
 #include <stddef.h>
 #include <errno.h>
 
+// When we want the argument to be translated, the cageid supplied must have the MSB set to 1. 
+// These masks help check this.
+#define LIND_ARG_TRANSLATE_FLAG (1ULL << 63)
```
rennergade: im not sure what this is, define a constant? or a better comment?

PR #932, opened 2026-03-12
Fix Exit/Signals for multi-threaded cages

### LINKED ISSUE #914: proposal: fix exit_group and thread_suicide
Here's my proposal to clean up both #908 and #34 which should close a lot of our threading problems. There's an additional issue about blocking syscalls at the end but that can be handled separately. If I can get approval on this should be able to implement it ASAP.

### Problem

`thread_suicide()` (`signal.rs:144`) raises `Trap::Interrupt` to kill a wasm instance. `catch_traps()` (`traphandlers.rs:254`) catches this trap, prints "Terminated", and returns `Ok(())`. This bypasses the entire exit cleanup path:

- `lind_manager.decrement()` never called → lind-boot hangs on `lind_manager.wait()`
- No zombie created → parent's `waitpid()` blocks forever
- No SIGCHLD sent to parent
- No fdtable cleanup

The `exit_call()` method (`lib.rs:967`) already implements proper exit via asyncify unwind — it does cleanup (rm_vmctx, lind_manager.decrement) and sets `OnCalledAction::Finish` so `_start.call()` returns normally with an exit code. But signal termination and thread killing bypass this entirely.

### Root cause

Two exit mechanisms exist that should be one:
1. **Normal exit**: `exit()` → RawPOSIX → 3i → `exit_call()` → asyncify unwind → cleanup ✓
2. **Signal/kill exit**: `signal_handler()` → `thread_suicide()` → trap → no cleanup ✗

### Fix

Replace `thread_suicide()` with `exit_call()` everywhere. `signal_handler` already has a `Caller<'_, T>` and can access `ctx`:

```rust
// signal.rs, Terminate branch (currently lines 69-76):
SignalDefaultHandler::Terminate => {
    cage::cage_record_exit_status(cageid, cage::ExitStatus::Signaled(signo, false));
    cage::signal::epoch_kill_all(cageid);
    // OLD: thread_suicide();
    // NEW: proper asyncify exit
    ctx.exit_call(caller, 128 + signo as i32, 1);
    return 0;
}
```

`exit_call` sets up asyncify unwind + `OnCalledAction::Finish`. The unwind propagates back through the call chain (`signal_handler` → `epoch_callback` → wasm → `_start.call()` returns with exit code). All cleanup happens in `exit_call` before the unwind starts.

Same fix for `thread_check_killed` (line 44-47): instead of `thread_suicide()`, call `exit_call`. For killed non-main threads, pass `is_last_thread=0` since the killing thread handles cage cleanup.

### exit_group semantics (issue #34)

Currently `exit()` only exits the calling thread. Other threads in the cage continue. The fix:

**In RawPOSIX `exit_syscall`** (`sys_calls.rs`): Before `lind_thread_exit()`, call `epoch_kill_all(cageid)` to mark all other threads for death. Non-main threads hit `thread_check_killed` → `exit_call(is_last_thread=0)` → asyncify unwind → their `_start.call()` returns → thread exits. Once all other threads are gone, the calling thread proceeds as last thread with full cage cleanup.

The wait-for-other-threads-to-die can check `epoch_handler.len() == 1` (only calling thread remains). May need a condvar or short spin since threads die asynchronously via asyncify.

### exec

Same pattern: `epoch_kill_all` other threads, wait for them to exit via asyncify, then proceed with module re-instantiation.

### What this doesn't solve

Threads blocked in host syscalls (`libc::read`, `futex_wait`, etc.) won't see the epoch because wasm isn't executing. They need a separate interruption mechanism (likely storing host `pthread_t` per thread and using `pthread_kill` to deliver EINTR). This is a separate issue from the exit path itself — the asyncify exit fix handles all threads that are executing wasm, which is the common case.

rennergade:
Closes #914 (hopefully)

Here's a working implementation that fixes thread exiting completely. I found that its actually quite a complex scenario with race conditions especially including 3i. I think this is at least a good starting point that gets us to short term goals. Some of this could be replaced with or encompass a harsh_cage_exit implementation I believe.

### Problem

Thread and process exit in Lind had two fundamental issues:

1. **`thread_suicide()` bypassed all cleanup.** It raised `Trap::Interrupt` to kill threads during exit and signal termination. This skipped zombie creation, SIGCHLD delivery, and `lind_manager.decrement()`, causing:
   - **waitpid hangs**: Parent blocks forever because the child never becomes a zombie
   - **lind-boot hangs**: `LindCageManager::wait()` never returns because `cage_count` is never decremented
   - **Grate dispatch panics**: In-flight grate-forwarded syscalls hit a removed cage and panic on `get_cage().unwrap()`
   - **Race conditions**: Multiple threads calling cage cleanup concurrently could double-decrement `cage_count`

2. **No distinction between exit and exit_group.** Both glibc's `exit()` (process termination) and `start_thread`'s `_exit(0)` (single thread returning) mapped to syscall 60, which treated every `_exit()` as exit_group. When a non-main thread returned from its function, `_exit(0)` triggered `epoch_kill_all`, killing the main thread before it could finish its work. This broke all multi-threaded tests where non-main threads returned before the main thread exited.

### Solution

Two changes, layered on top of each other:

**1. Replace `thread_suicide()` with asyncify-based exit.**

All `thread_suicide()` calls are replaced with `ctx.exit_call()`, which initiates a proper asyncify unwind. Cage cleanup is deferred to each thread's `OnCalledAction` callback after asyncify completes. The actual last thread to exit handles `cage_finalize` (zombie creation, SIGCHLD, fdtable removal, cage removal).

**2. Split exit (syscall 60) from exit_group (syscall 231) to match Linux semantics.**

- **Syscall 60 (`exit`)**: Thread-only exit. Used by glibc's `start_thread` when a non-main thread returns from its thread function. Records exit status and triggers asyncify unwind. Does NOT initiate cage-wide shutdown.
- **Syscall 231 (`exit_group`)**: Process exit. Used by glibc's `exit()` after running atexit handlers and flushing stdio. Wins `try_initiate_exit_group` CAS → sets `is_dead` → inserts into `EXITING_TABLE` → `epoch_kill_all` → removes grate handlers → asyncify unwind → deferred cleanup in `OnCalledAction`.

### Architecture

**Exit flow (before — broken):**
```
exit()/_exit() → syscall 60 → exit_syscall
  → epoch_kill_all (always, even for single thread exit)
  → thread_suicide (Trap::Interrupt, no cleanup)
  → zombie/SIGCHLD sometimes skipped
```

**Exit flow (after):**
```
Thread return → _exit(0) → syscall 60 → exit_syscall
  → record exit status
  → make_syscall to wasmtime → exit_call → asyncify unwind
  → OnCalledAction: lind_thread_exit
      → if last thread: cage_finalize (zombie, SIGCHLD, fdtable, remove_cage)
      → rm_vmctx, lind_manager.decrement()

exit() → __run_exit_handlers → __lind_exit_group → syscall 231 → exit_group_syscall
  → try_initiate_exit_group CAS (first thread wins)
  → is_dead=true, EXITING_TABLE.insert, epoch_kill_all, _rm_grate_from_handler
  → make_syscall to wasmtime → exit_call → asyncify unwind
  → OnCalledAction: same deferred cleanup as above
```

Each thread independently unwinds via asyncify and removes itself from `epoch_handler`. The last thread to finish handles all cage teardown. This eliminates the need for any thread-draining wait loops.

### Race condition fixes

**Double asyncify unwind** (`signal.rs`): When `epoch_kill_all` fires, killed threads enter `signal_handler`. If a thread was already in asyncify unwind (e.g. from its own `exit_call` via syscall 60), calling `exit_call` again would corrupt the unwind state and cause OOB memory faults. Fix: check `AsyncifyState::Unwind` before calling `exit_call`.

**Asyncify rewind from exit_call** (`signal.rs`, `lind-common/lib.rs`): During asyncify rewind from an `exit_call`, the epoch can fire and re-enter `signal_handler`, or the syscall entry point can be hit. Both paths check for rewind data (`get_current_signal_rewind_data` / `get_current_syscall_rewind_data`) and unwrap. But exit rewinds have no signal/syscall data. Fix: return early on `None` instead of panicking.

**Late thread after cage removal** (`cage.rs`): A thread can be between `futex_wake` (signaling pthread_join done) and `_exit(0)` when `epoch_kill_all` fires. Since epoch only takes effect in WASM code, the thread continues through native rawposix `exit_syscall` and hits `cage_record_exit_status` after `cage_finalize` has already removed the cage. Fix: `cage_record_exit_status` returns silently when `get_cage()` returns None.

**Grate handler table cleanup ordering** (`signal.rs`, `threei.rs`): Grate handler table entries must be cleaned up before VMContext removal on exit. `_rm_grate_from_handler` is called during exit to prevent new grate dispatches to a dying cage.

**EXITING_TABLE ordering**: Must be set before `epoch_kill_all` to close the window where a cross-cage call could slip through after the cage is marked for death but before killed threads have exited.

**EXITING_TABLE** (`DashSet<u64>`): Persists cage IDs after `remove_cage()`. `make_syscall` checks both `is_cage_dead()` and `EXITING_TABLE.contains()` to block cross-cage calls to fully removed cages.

**grate_inflight** (`AtomicU64`): Incremented before grate dispatch, decremented after. `cage_finalize` spins until it reaches 0. Double-checked against `is_dead` after increment to close the race window.

**SIGUSR2 in `epoch_kill_all`**: After writing `EPOCH_KILLED` to each thread's epoch pointer, sends `SIGUSR2` via `tkill` to interrupt threads blocked in host syscalls (e.g. `futex`).

#### Asyncify-based exit (`cage/`, `lind-multi-process/`, `rawposix/`)
- **`exit_syscall` (60)**: Thread-only exit. Records status and triggers asyncify unwind. No CAS, no epoch_kill_all.
- **`exit_group_syscall` (231)**: New. Full exit_group path with `try_initiate_exit_group()` CAS, `is_dead`, `EXITING_TABLE`, `epoch_kill_all`, `_rm_grate_from_handler`.
- **`try_initiate_exit_group()`**: `AtomicBool` CAS so only the first thread initiates shutdown; other threads just exit individually.
- **`is_dead` flag**: `AtomicBool` on `Cage` struct, set when exit begins. Cross-cage `make_syscall` checks this and returns `-ESRCH`.
- **Deferred `cage_finalize`**: Called from `OnCalledAction` (after asyncify unwind), not inline during exit. Waits for `grate_inflight` to drain before removing the cage.
- **`lind_thread_exit` restructured**: Remove-first then `is_empty()` check eliminates DashMap iterator race that could miss entries and misidentify the last thread.
- **`cage_record_exit_status`**: Gracefully handles already-removed cages (returns silently on None) for late threads that miss the epoch.
- **Killed threads** (signal_handler): Use `exit_call` instead of `thread_suicide`, with guard against double asyncify unwind. `lind_thread_exit` deferred to `OnCalledAction`.
- **Asyncify rewind guards**: Both `signal_handler` and syscall entry handle missing rewind data gracefully (exit rewinds have no signal/syscall data).
- **`exec_syscall`**: No longer clears `epoch_handler` before attempting exec. If exec fails (e.g. ENOENT), the thread continues running and still needs its epoch_handler entry for proper exit tracking.

#### glibc changes
- **`exit.c`**: New `__lind_exit_group()` calls `EXIT_GROUP_SYSCALL` (231). `__run_exit_handlers` calls `__lind_exit_group` instead of `_exit` so that `exit()` properly kills all threads.
- **`lind_syscall_num.h`**: Added `EXIT_GROUP_SYSCALL 231`.
- **`_exit()`** remains syscall 60 — single-thread exit for `start_thread`.

#### Race condition fixes (`threei/`, `lind-utils/`)
- **`EXITING_TABLE`**: `DashSet<u64>` persisting cage IDs after removal, checked in `make_syscall`.
- **`grate_inflight`**: Increment/decrement around grate dispatch with `is_dead` double-check.
- **`_rm_grate_from_handler`**: Handler table cleanup during exit.
- **`_get_handler`**: Panics on missing cage/syscall entries (invariant violations). Empty target_map (from `_rm_grate_from_handler` cleanup) returns None gracefully.
- **`LindCageManager`**: `wait()` uses `> 0` (not `!= 0`) and `decrement()` notifies on `<= 0` as defense against `cage_count` going negative.
- **fdtable**: `remove_cage_from_fdtable` handles concurrent removal gracefully instead of asserting.

### Known issue: grate dispatch OOB with mmap/futex interception

Grate-intercepted mmap and futex calls cause intermittent OOB faults and deadlocks under concurrent multi-cage workloads. This is a pre-existing grate dispatch issue, not caused by exit changes:
- **mmap interception**: OOB at addresses near 0xfff59da8/0xfff7ada8 (thread stack region). Likely return value corruption during grate round-trip.
- **futex interception**: VMContext pool exhaustion when many threads make concurrent grate futex calls. A blocked futex_wake prevents pthread_join from returning → deadlock.
- Both issues disappear when grate interception is disabled.
- race-test_grate.c has interception disabled pending investigation.

### New Tests
- **`exit_group_thread.c`**: Child thread calls `exit(42)` while main thread is blocked in `pthread_join`. Verifies exit_group terminates all threads and parent gets correct exit status via `waitpid`.
- **`signal_kill_cleanup.c`**: Parent sends `SIGTERM` to child, verifies `WIFSIGNALED` and proper zombie cleanup via `waitpid`.
- **`race-test.c` / `race-test_grate.c`**: Grate parent with fork+threads+exit. Exercises `grate_inflight` counter and `EXITING_TABLE` guards. Syscall interception disabled pending grate dispatch fix.
- **`kill.c` removed from skip list**: Signal termination now works correctly.

### DIFF
--- a/src/lind-boot/src/lind_wasmtime/execute.rs
+++ b/src/lind-boot/src/lind_wasmtime/execute.rs
@@ -92,8 +92,7 @@ pub fn execute_wasmtime(lindboot_cli: CliOptions) -> anyhow::Result<i32> {
             let cageid = CAGE_START_ID as u64;
             cage::cage_record_exit_status(cageid, cage::ExitStatus::Exited(1));
             if let Some(c) = cage::get_cage(cageid) {
-                c.is_dead
-                    .store(true, std::sync::atomic::Ordering::Release);
+                c.is_dead.store(true, std::sync::atomic::Ordering::Release);
             }
             threei::EXITING_TABLE.insert(cageid);
             threei::handler_table::_rm_grate_from_handler(cageid);
--- a/src/wasmtime/crates/lind-multi-process/src/lib.rs
+++ b/src/wasmtime/crates/lind-multi-process/src/lib.rs
@@ -480,8 +480,7 @@ impl<
                                 cage::ExitStatus::Exited(1),
                             );
                             if let Some(c) = cage::get_cage(child_cageid) {
-                                c.is_dead
-                                    .store(true, std::sync::atomic::Ordering::Release);
+                                c.is_dead.store(true, std::sync::atomic::Ordering::Release);
                             }
                             threei::EXITING_TABLE.insert(child_cageid);
                             threei::handler_table::_rm_grate_from_handler(child_cageid);


src/cage/src/cage.rs
```diff
@@ -139,6 +150,10 @@ pub struct Cage {
     // address of the wasm thread. The epoch is a u64 value that guest thread is frequently checking for
     // and just to host once the value is changed
     pub epoch_handler: DashMap<i32, RwLock<*mut u64>>,
+    // os_tid_map maps lind thread IDs to OS thread IDs (gettid). Used by
+    // epoch_kill_all to send SIGUSR2 to threads blocked in host syscalls,
+    // interrupting them so they can re-enter wasm and see the epoch kill.
+    pub os_tid_map: DashMap<i32, i64>,
```
Yaxuan-w: Can we also mention which one in Dashmap is os's tid which one is Lind's tid?

src/rawposix/src/sys_calls.rs
```diff
@@ -323,96 +313,80 @@ pub extern "C" fn exit_syscall(
...
+    arg3: u64,
+    arg3_cageid: u64,
+    arg4: u64,
+    arg4_cageid: u64,
+    arg5: u64,
+    arg5_cageid: u64,
+    arg6: u64,
+    arg6_cageid: u64,
+) -> i32 {
+    let status = sc_convert_sysarg_to_i32(status_arg, status_cageid, cageid);
```
Yaxuan-w: Also needs to check unused args

rennergade:
> The grate unit test failed. Maybe related to recent commits in this pr?

Ah yeah that's the one I added here need to skip list, has the separate mmap and futex problems I listed above.

src/cage/src/signal/signal.rs
```diff
@@ -15,14 +15,17 @@ pub fn signal_epoch_trigger(cageid: u64) {
...
 
         let threadid_guard = cage.main_threadid.read();
         let main_threadid = *threadid_guard;
-        let epoch_handler = cage
-            .epoch_handler
-            .get(&main_threadid)
-            .expect("main threadid does not exist");
+        let epoch_handler = match cage.epoch_handler.get(&main_threadid) {
+            Some(h) => h,
+            None => return,
```
qianxichen233: not sure if we should silently return here. We should never encounter None here, and if we encountered one, that probably means something went wrong and we should check

Maybe doing return for release build but panic on debug build

Also applies to all other unwarp handling: I would still want to panic if we know the panic should normally never be triggered, this allows us to locate potential bugs much more efficiently

qianxichen233 (approved):

PR #967, opened 2026-03-20
fork: return child_cageid instead of 0 on success.

### LINKED ISSUE #966: Grates interposing on `fork` don't know the cageid of the newly forked cage.
### Issue/Current Behaviour

Setting the scene a bit, 

For IMFS and other such grates, we need the ability to interpose on `fork`, mostly so that we can copy over the file descriptor tables (stored in IMFS's memory space) from the parent cage to the newly forked child cage. This requires the grate handler calling `make_threei_call` on behalf of the parent cage to have access to the cageid of the newly created child. This is not currently possible.

When we call `fork`, the call goes to rawposix's `fork_syscall` which sets up the new cage, copies over the vmmaps, fdtables, etc. then delegates this to the wasmtime. 

Wasmtime uses asyncify to unwind and rewind the execution phases of both the parent and the child cage.

We set the return value for the child wasm module here:
https://github.com/Lind-Project/lind-wasm/blob/a8c8665afe2c07c3688fb95cba3be244e9544042/src/wasmtime/crates/lind-multi-process/src/lib.rs#L441-L444

and then we set the return value for the parent wasm module here:
https://github.com/Lind-Project/lind-wasm/blob/a8c8665afe2c07c3688fb95cba3be244e9544042/src/wasmtime/crates/lind-multi-process/src/lib.rs#L521-L522

This is how the "dual return values" for fork currently work. 

What this means is that `rawposix::fork_syscall` is not the component doing any of this, it only exists in one process, and can only return 1 value, which is currently set to 0. Therefore, anything observing this function also only ever sees 0 (which is what the grate handler is doing). 

### Possible fixes:

- Have `rawposix::fork_syscall` always return the child cageid of the new cage. The value it returns doesn't get seen by the wasm modules, but anything observing it would have information about the new cageid.

stupendoussuperpowers:
Closes #966.

### DIFF
--- a/tests/grate-tests/interpose-fork_grate.c
+++ b/tests/grate-tests/interpose-fork_grate.c
@@ -43,7 +43,7 @@ int fork_grate(uint64_t cageid,
   printf("[Grate|interpose-fork] In fork_grate %d handler for cage: %llu\n",
          getpid(), cageid);
   int self_grate_id = getpid();
-  int ret = make_threei_call(
+  int new_cageid = make_threei_call(
     56, // syscallnum for clone
     0,    // callname is not used in the trampoline, set to 0
     self_grate_id,    // self_grate_id is not used in the trampoline, set to 0
@@ -56,7 +56,11 @@ int fork_grate(uint64_t cageid,
     arg6, arg6cage,
     0 // we will handle the errno in this grate instead of translating it to -1 in the trampoline
   );
-  return ret;
+
+  // Ensure the return value is a valid process ID.
+  assert(new_cageid > 0);
+
+  return new_cageid;
 }
 
 // Main function will always be same in all grates


rennergade:
This seems to pass CI so it doesnt break anything and gives us correct behavior for 3i. However I still don't completely understand why that is.

stupendoussuperpowers:
> Do we expect the fork interposition in grate to only have the parent routine in design?

Suppose you have (G)rate interposing on Cage (C1)'s fork. If G wants to spawn a new cage, the only mechanism for it to do so is through `make_threei_call(SYS_CLONE,...)`. Once it does that and C2 now exists, it needs to know the cage_ids for both C1 and C2. It already has C1's, with this change it also has C2's. 

C1 and C2 are oblivious to what `make_threei_call` or `fork_syscall` return, those values are never seen by C1 and C2, since asyncify overrides them using `Rewind`. 

> 1. Only parent routine is visible to grate, and grate is operating in Asyncify::Unwind mode, which is not safe to run code

There is not much of a "parent" or "child" routine for G. But it might need C1.cage_id and C2.cage_id, if say for example it wants to do some bookkeeping, in the example of IMFS, this includes copying over the fdtables that G maintains. 

> 2. child will not spawn in grate, it will still spawn in cage

This is the behaviour we want. 

The fork shouldn't happen in G, it should happen in C1. Interposing on fork shouldn't change assumptions about C2's parent since that would violate all sorts of assumptions for what it means to C1 when it calls fork. If we fork on G, it's gonna copy over the G memory instead which is essentially just a no-op from C1's perspective.

If for whatever reason a grate wants C2 be a child of G, it can simply call `fork()` itself and deal with that how it wants to. I am not sure why a grate would want that though.

qianxichen233 (approved):
Okay, if that's desired behavior, that's fine. I am not sure what possible scenarios that is in the scope of grate working with fork, but just keep in mind that it will be kind restricted in certain scenario.

Also the `--asyncify` is not needed for grate compilation in this case, the Asyncify feature isn't really used and cannot be used by grate right now. Removing it should slightly improve the performance of grate

rennergade (approved):

PR #976, opened 2026-03-23
Libcpp compilation integration

123R3N321:
### Integrate libcpp compilation into main codebase

1. libcpp compilation toolchain cmake file and compile .sh script used to generate artifacts, stored in artifacts/ directory. Note the toolchain cmake and bash scripts are removed after one-time use to create the artifacts. (According to Alice, they are still tracked in lind-wasm-apps repo?)

2. Additional makkefile script to compile cpp into wasm/ Note we are suppressing throw-exception syntax. This will be a major fix in the future. See scripts/lind_compile_cpp

3. documentation in .md, found in docs/ dir.

4. CI integration with additioanl html rendering in e2e report: added libcpp test report with one singular unit test using the algorithms.h header.


Yaxuan-w (changes requested):
We need to discuss how we want to standardize cpp testing. 

- I'm thinking to have a separate folder under `tests/` called `tests/cpp/` for all cpp unit test
- cpp compilation scripts can be put under `scripts/cpp` and put `lind_compile` into `scripts/c`, rust version into `scripts/rust` (or sth similar)

scripts/lind_compile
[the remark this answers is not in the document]
123R3N321: I think it is better to not touch the existing path for original lind_compile for clang and .c compilation; I plan to simply create a separate lind_compile_cpp.sh at the same directory level as the lind_compile script and add corresponding tasks to the CI flow.

The existing CI flow worries me as the workflow is very complicated and any changes to the path of original lind_compile might cause issues. The call chain: e2e.yml -> dockerfile.e2e-> Makefile->test_runner.py->check_reports.py

123R3N321:
> > @m-hemmings regarding the CI problem:
> > Nick has brought to my attention that the current workaround for libcpp header involves this workflow: git clone llvm-18 ->modification of llvm source code -> compile archive files -> copy over to /build/sysroot -> sysroot provides dependency for clang++ to compile cpp source code.
> > This is problematic. On the one hand, llvm is big and preserving their source code in our repo is confusing; on the other hand, as Alice points out:
> > > Source code and scripts in separate repos will make things hard to understand. We can use prebuilt libcpp.a for ci
> > 
> > 
> > However, this implies that, should there be any upgrade of llvm to version beyond 18, we will need to re-create this entire workflow to compile an updated version of the .a archive(s) needed for libcpp header support. I will do my best to document this workflow and, for future improvements, automate the process as far as I can
> > Nick also points out that we are using llvm-16 in [lind-wasm-apps](https://github.com/Lind-Project/lind-wasm-apps) repo, see commit [0e78686](https://github.com/Lind-Project/lind-wasm-apps/commit/0e7868679b02511acd3181ba32fb929a4cf92c94); while my PR for [lind-wasm](https://github.com/Lind-Project/lind-wasm/tree/main) repo is now using llvm-18, this is consistent with the [docker config](https://github.com/Lind-Project/lind-wasm/blob/936d9ffe6d7177d3a3c66b5670f14bde8c2a6c65/Docker/Dockerfile.dev#L19) in this repo.
> 
> Thanks for writing this up. I agree that the current approach puts us in an awkward spot: keeping LLVM source in the repo is heavy and confusing, but moving all of the logic somewhere else also makes the workflow harder to understand and maintain.
> 
> I think there is a cleaner middle ground. Instead of treating libc++ support as something we rebuild from LLVM source during normal CI, we could treat it as a versioned toolchain artifact.
> 
> Concretely, I would suggest:
> 
> * Keep one documented build script in this repo that knows how to produce the exact libc++ headers and archive files we need for a given LLVM version.
> * Run that script only when we intentionally upgrade LLVM or change the libc++ workflow.
> * Publish the result as a named artifact, for example an LLVM-versioned sysroot overlay or tarball.
> * Have CI download and unpack that artifact instead of cloning and rebuilding LLVM as part of ordinary PR validation.
> * Add one small smoke test in CI that compiles a trivial `.cc` file with `clang++` so we catch regressions immediately.
> 
> That gives us a few benefits:
> 
> * The normal CI path stays much smaller and easier to reason about.
> * We keep the process reproducible because the generation script still lives here.
> * We avoid preserving a large LLVM source tree in the repo.
> * Future LLVM upgrades become an intentional refresh of a pinned artifact.
> 
> On the versioning point, I do think it would help to align on a single LLVM major version across repos if we can. This repo is already pinned to LLVM 18 in Docker, so moving `lind-wasm-apps` to the same version would make the overall story much easier to follow.
> 
> So in short, my suggestion would be:
> 
> 1. Keep the build recipe in-repo.
> 2. Stop keeping LLVM source in-repo for day to day CI.
> 3. Publish a prebuilt, versioned libc++ support artifact for CI consumption.
> 4. Add a simple C++ compile smoke test so the setup stays honest.
> 
> What does everyone think about that?

step 1, 3, 4 complete. I need some more reading and learning before I can determine if it is safe to remove LLVM all together.

123R3N321:
Additional comment from Nick:
integrate the libcpp testing into the report html generated.

123R3N321:
Removed separate libcpp integration testing job in CI as instructed by @Yaxuan-w 

Added ``scripts/harnesses/libcpptestreport.py`` following the styling of ``wasmtestreport.py`` in the same directory as instructed by @rennergade Currently passing 1/1 test.

rennergade:
> Removed separate libcpp integration testing job in CI as instructed by @Yaxuan-w
> 
> Added `scripts/harnesses/libcpptestreport.py` following the styling of `wasmtestreport.py` in the same directory as instructed by @rennergade Currently passing 1/1 test.

this test just verifies compilation? It would be cool to actually run the programs like the main tester does

scripts/lind_compile
```diff
@@ -530,4 +530,4 @@ case "${MODE}" in
   "precompile-only")
     echo "OK: ${OUT_CWASM}"
     ;;
-esac
+esac
```
Yaxuan-w, line 533: Add new line

scripts/lind_compile_cpp
```diff
@@ -0,0 +1,395 @@
...
+      ;;
+    --print-args|--debug-args)
+      PRINT_ARGS="true"
+      ;;
+    --no-default-clang-flags)
+      NO_DEFAULT_CLANG_FLAGS="true"
+      ;;
+    --compile-grate)
+      COMPILE_GRATE="true"
+      ;;
```
Yaxuan-w, line 115: Don't need those in cpp yet?

scripts/lind_compile_cpp
```diff
@@ -0,0 +1,395 @@
...
+  full)
+    echo "OK: ${OUT_CWASM}"
+    ;;
+  "compile-only"|"opt-only")
+    echo "OK: ${OUT_WASM}"
+    ;;
+  "precompile-only")
+    echo "OK: ${OUT_CWASM}"
+    ;;
+esac
```
Yaxuan-w, line 366: Add new line

trial/hello.cpp
Yaxuan-w, line 1: Remove this file

.gitignore
```diff
@@ -20,3 +20,7 @@ report.html
 
 #Ignore python cache
 __pycache__/
+
+#Ignore llvm external dependency needed for libcpp headers
+llvm-project/
+libcxx-*/
```
Yaxuan-w: No need for those two?

build-llvm.sh
Yaxuan-w, line 1: Also this should be removed since we keep in the app repo

docs/internal/libcpp.md
Yaxuan-w, line 1: Put into doc folder

Toolchain-WASI.cmake
Yaxuan-w, line 1: Also remove this

scripts/lind_compile_cpp
123R3N321: The lind_compile_cpp script actually has a ton of space for improvement since it is literally scrambled up from scraps taken from the original lind_compile script.

will we have the need to preserve such a flag in the future? I have it here simply because, well, I haven't deleted it after copy-pasting the original lind_compile script.

There is much to do, as almost everything in this script assumes c source code usage (such as the prompt when you pass in -help flag). I will see to it after getting all the more important parts done

trial/hello.cpp
123R3N321: ok, that will require corresponding changes to my documentation. Currently it is already no longer used by anything in the test suite or CI flow. I am removing it now

.gitignore
123R3N321: wait are we tracking llvm project in git now? 

If not, I understand these directories won't be here unless someone manually curled and then decided to push. The way I see it is that I would keep there here for safety reasons; if for some reasons in the future we want to keep track of directories of these names, someone can easily spot the gitignore situation. (Easy fix to remove it; hard fix to manually clean such large files if someone accidentally commit.)

build-llvm.sh
123R3N321: I might need to ask you what this means... also I would make sure to document this thing before removing it, else it would be very confusing to future devs.

docs/internal/libcpp.md
123R3N321: Done.

ok I'm not sure I want this to show up on the official documentation site yet... I will look into it and possibly rework the entire documentation one this PR is truely ready for merge, so that future devs know what is going on.

Toolchain-WASI.cmake
123R3N321: same question as above. I will ask you in person.

123R3N321:
> > Removed separate libcpp integration testing job in CI as instructed by @Yaxuan-w
> > Added `scripts/harnesses/libcpptestreport.py` following the styling of `wasmtestreport.py` in the same directory as instructed by @rennergade Currently passing 1/1 test.
> 
> this test just verifies compilation? It would be cool to actually run the programs like the main tester does

Done. a couple assumptions: 
1.  lind_run is used as blackbox to execute .wasm
2. I "suppressed" the comipler warning regarding main func argument list style so that the report looks less noisy.

scripts/lind_compile_cpp
Yaxuan-w: We can remove grates related part at least. But yeah this has less priority than others

.gitignore
Yaxuan-w: > If not, I understand these directories won't be here unless someone manually curled and then decided to push. The way I see it is that I would keep there here for safety reasons; if for some reasons in the future we want to keep track of directories of these names, someone can easily spot the gitignore situation. (Easy fix to remove it; hard fix to manually clean such large files if someone accidentally commit.)

The gitignore only tracks the "intended" files in this repo. LLVM source has been placed into app repo, so remaining it in gitignore makes it less readable

build-llvm.sh
Yaxuan-w: Maybe we can include the usage to the app repo in the doc

Yaxuan-w:
Tagging @Uk-jake for CI related reviews

Yaxuan-w (dismissed):

rennergade:
minor formatting changes before merge:

Libc++ integration -> c++ harness

remove this paragraph:
Full lind_compile_cpp (wasm-opt + precompile), lind_run on the .cwasm, and verify native and wasm runs have identical exit codes and output. Native time / Wasm time are total wall time (native compile + native run vs lind_compile_cpp + lind_run), same convention as the deterministic/fail tables above.

I dont know what this is in the output: Native/Wasm parity verified (exit=0). remove

PR #979, opened 2026-03-24
register_handler: calling cage should be the destination grate.

stupendoussuperpowers:
The current model when we interpose on register_handler is: "If anyone tries to interpose this target_cage, send it to me" and not "If this grate is the target for a register handler on any random cage, send it to me"

This PR fixes that and updates the interpose-register test to be compatible with the expected semantics.


stupendoussuperpowers:
Shouldn't the register_handler interposition have 3 grates?

main grate -> regular grate -> cage

and whenever we try to register_handler with target_grateid = regular grate, it goes to main grate first?

Not sure what the test we currently have is doing.

Yaxuan-w (approved):

PR #980, opened 2026-03-25
fix: skipping optimization, precompilation of grates with "-" in their name

drapl0n:
* fix: skipping optimization, precompilation of grates with "-"  in their name.


drapl0n:
Container is not updated with the latest `lind_compile` script. The old script does not copy precompiled file to lindfs.

stupendoussuperpowers (approved):

PR #194, opened 2025-04-30
Add files for PR Checklist comment via github workflows

m-hemmings:
This pull request introduces a GitHub workflow that automatically posts a checklist comment on every newly opened pull request.

How it works:

- The workflow is triggered when a new pull request is opened (pull_request: types: [opened])
- It uses actions/github-script to read .github/templates/PR_Checklist.md
- The contents of that file are posted as a comment on the pull request using GitHub’s REST API
- The checklist provides contributors with a consistent set of review criteria before maintainers begin a review

Template location:
.github/templates/PR_Checklist.md

Workflow location:
.github/workflows/pr-comment.yml

Purpose:
This ensures that contributors are reminded to follow key quality checks, helping to streamline reviews and maintain code quality across the project.

Issue #188, opened 2025-04-18
Add individual contact infos to Team page


lukpueh:
**Team page**:  https://github.com/Lind-Project/lind-wasm-docs/blob/main/docs/community/team.md
**Context**: These may be used for code of conduct violation reporting (https://github.com/Lind-Project/lind-wasm-docs/pull/39#discussion_r2049411765)
**Sub-issues**:
- Add team members (cc @marcelamelara)
- Reference team page in code of conduct (https://github.com/Lind-Project/lind-wasm-docs/pull/39#discussion_r2035640898

PR #986, opened 2026-03-27
Add support for setxattr and listxattr system calls

### LINKED ISSUE #889: Add support for setxattr and listxattr system calls
# Task:
Please provide support for `setxattr()` and `listxattr()` system calls

# Context
While porting Python 3.14.3 to Lind-Wasm, the test suite failed because `os.listxattr` is not yet implemented in the runtime. Error was :
'Function not implemented' 

# How to reproduce error
The following test can reproduce it.

```
#include <stdio.h>
#include <stdlib.h>
#include <sys/xattr.h>
#include <fcntl.h>
#include <unistd.h>
#include <string.h>

int main() {
    const char *path = "test_file_c.txt";

    // Create a dummy file
    int fd = open(path, O_CREAT | O_WRONLY, 0644);
    close(fd);
    // 1. Set an attribute
    if (setxattr(path, "user.demo", "hello-from-c", 12, 0) == 0) {
        printf("setxattr Success\n");
    } else {
        // Use perror to see WHY it failed (e.g., ENOSYS, EACCES, ENOTSUP)
        perror("setxattr failed");
    }
    ssize_t buflen = listxattr(path, NULL, 0);

   if (buflen < 0) {
        perror("listxattr failed");
    }
    unlink(path);
    return 0;
}

```

celinehoang177:
Implements support for setxattr() and listxattr() system calls in Lind-Wasm, enabling compatibility with Python 3.14.3 and other applications that use extended file attributes.

Changes:
- Implements setxattr_syscall() (Linux x86_64 syscall #188) to set extended attributes on files
- Implements listxattr_syscall() (Linux x86_64 syscall #194) to list extended attributes on files
- Registers both syscalls in the SYSCALL_TABLE for both rawposix and wasmtime runtimes
- Properly converts paths to the host namespace and handles errno conversions

Closes #889


vidyalakshmir (commented):
@celinehoang177 What about the changes to glibc?

Yaxuan-w (commented):
LGTM. Is there any reason that we add new files for those two syscalls instead of just changing the source in glibc?

rennergade (changes requested):
Pointed out a few issues, also need to add a test.

vidyalakshmir:
@celinehoang177 Please address the comments above.

celinehoang177:
> LGTM. Is there any reason that we add new files for those two syscalls instead of just changing the source in glibc?

I fixed it. It's now just changing the source in glibc.

src/rawposix/src/fs_calls.rs
```diff
@@ -4461,3 +4461,140 @@ pub extern "C" fn getrandom_syscall(
...
+        Ok(path) => path,
+        Err(e) => return syscall_error(e, "setxattr", "path conversion failed"),
+    };
+
+    let name = match sc_convert_path_to_host(name_arg, name_cageid, cageid) {
+        Ok(name) => name,
+        Err(e) => return syscall_error(e, "setxattr", "name conversion failed"),
+    };
+
+    let value = value_arg as *const libc::c_void;
```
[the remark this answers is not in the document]
celinehoang177: I fixed it

src/rawposix/src/fs_calls.rs
```diff
@@ -4461,3 +4461,140 @@ pub extern "C" fn getrandom_syscall(
...
+    arg6: u64,
+    arg6_cageid: u64,
+) -> i32 {
+    // Type conversion
+    let path = match sc_convert_path_to_host(path_arg, path_cageid, cageid) {
+        Ok(path) => path,
+        Err(e) => return syscall_error(e, "listxattr", "path conversion failed"),
+    };
+
+    let list = if list_arg == 0 {
```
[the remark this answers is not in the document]
celinehoang177: I fixed it

src/rawposix/src/fs_calls.rs
```diff
@@ -4567,3 +4567,151 @@ pub extern "C" fn getrandom_syscall(
...
+        Err(e) => return syscall_error(e, "listxattr", "path conversion failed"),
+    };
+
+    // Type conversion for list buffer (may be NULL)
+    let list = if list_arg == 0 {
+        std::ptr::null_mut()
+    } else {
+        sc_convert_buf(list_arg, list_cageid, cageid) as *mut libc::c_char
+    };
+    let size = sc_convert_sysarg_to_usize(size_arg, size_cageid, cageid);
```
vidyalakshmir, line 4926: In the earlier line, guest memory pointers are translated using `sc_convert_buf` without validating the guest's provided size argument. This allows a malicious guest to allocate a tiny buffer but pass a massive size, tricking the host kernel into writing outside the guest's memory boundaries (a buffer overflow).

**Proposed Fix:** We need to add a bounds check to validate that the entire contiguous memory region from pointer to pointer + size resides completely within the allowed cage memory. (pointer is `list_arg`).
@Yaxuan-w @rennergade @qianxichen233

src/rawposix/src/fs_calls.rs
```diff
@@ -4567,3 +4567,151 @@ pub extern "C" fn getrandom_syscall(
...
+        Err(_) => return syscall_error(Errno::EINVAL, "setxattr", "name contains null byte"),
+    };
+
+    // Type conversion for value buffer (may be NULL)
+    let value = if value_arg == 0 {
+        std::ptr::null()
+    } else {
+        sc_convert_buf(value_arg, value_cageid, cageid) as *const libc::c_void
+    };
+    let size = sc_convert_sysarg_to_usize(size_arg, size_cageid, cageid);
```
vidyalakshmir, line 4860: Similar to the comment for listxattr, there can be an out-of-bounds read here. We need to check if `pointer + size` is within the allowed cage memory.

If the guest allocates a 5-byte buffer containing "hello", but calls `setxattr` with a size of 1000, the host kernel will read 1000 bytes starting from that pointer and save it to the file's extended attributes. The guest just forced the host kernel to read 995 bytes of adjacent memory and save it to the disk. The guest can then read that file's extended attributes to steal the host's memory. 

@Yaxuan-w @rennergade @qianxichen233

src/rawposix/src/fs_calls.rs
```diff
@@ -4567,3 +4567,151 @@ pub extern "C" fn getrandom_syscall(
...
+        );
+    }
+
+    // Call to kernel listxattr
+    let ret = unsafe { libc::listxattr(path.as_ptr(), list, size) };
+
+    if ret < 0 {
+        let errno = get_errno();
+        return handle_errno(errno, "listxattr");
+    }
```
vidyalakshmir, line 4945: The `setxattr` and `listxattr` implementations allow the guest to operate on arbitrary xattr namespaces. Since the host kernel evaluates these calls using the sandbox runtime's privileges, a malicious guest can write to `security.*` attributes (e.g. `security.capability` to grant root capabilities) or modify ACLs via `system.*` — fully bypassing host access controls. All xattr operations must be restricted to the `user.*` namespace, with anything else rejected at the shim layer.
@Yaxuan-w @rennergade @qianxichen233

vidyalakshmir (changes requested):

vidyalakshmir (dismissed):
Approving this. The changes requested would be tackled in a [different issue](https://github.com/Lind-Project/lind-wasm/issues/1082) and [this issue](https://github.com/Lind-Project/lind-wasm/issues/1083). Please fix the merged conflicts.

Yaxuan-w:
Needs to resolve merge conflicts before merging

Yaxuan-w (approved):

src/rawposix/src/fs_calls.rs
vidyalakshmir: Resolving this as it would be tackled in [different issue](https://github.com/Lind-Project/lind-wasm/issues/1082) and [this issue](https://github.com/Lind-Project/lind-wasm/issues/1083).

src/rawposix/src/fs_calls.rs
vidyalakshmir: Resolving this as it would be tackled in [different issue](https://github.com/Lind-Project/lind-wasm/issues/1082) and [this issue](https://github.com/Lind-Project/lind-wasm/issues/1083).

vidyalakshmir: Resolving this as it would be tackled in [different issue](https://github.com/Lind-Project/lind-wasm/issues/1082) and [this issue](https://github.com/Lind-Project/lind-wasm/issues/1083).

src/rawposix/src/fs_calls.rs
vidyalakshmir: Resolving this as it would be tackled in [different issue](https://github.com/Lind-Project/lind-wasm/issues/1082) and [this issue](https://github.com/Lind-Project/lind-wasm/issues/1083).

src/rawposix/src/fs_calls.rs
vidyalakshmir: Resolving this as it would be tackled in [different issue](https://github.com/Lind-Project/lind-wasm/issues/1082) and [this issue](https://github.com/Lind-Project/lind-wasm/issues/1083).

vidyalakshmir (approved):

PR #994, opened 2026-03-29
symlink implementation

### LINKED ISSUE #993: symlink/symlinkat syscalls not implemented, ln -s fails
ln -s silently fails because symlink and symlinkat are not implemented in rawposix. Need to implement both syscalls to support symbolic link creation.

moogchi:
## Summary
Implements the `symlink` and `symlinkat` syscalls to support symbolic link creation in the Lind sandbox. This unblocks `ln -s`, which requires `symlink`/`symlinkat` to create symbolic links.

**Before:** `ln -s linkoriginal newlink` silently failed with no symlink created.
**After:** `ln -s linkoriginal newlink` successfully creates the symbolic link.

### RawPOSIX (Rust handler)
- **`src/rawposix/src/fs_calls.rs`**: Added `symlink_syscall` and `symlinkat_syscall` handlers. Both convert path arguments via `sc_convert_path_to_host` and delegate to `libc::symlink` and `libc::symlinkat` respectively. `symlinkat` handles `AT_FDCWD` as a special case, falling back to `libc::symlink`.
- **`src/rawposix/src/syscall_table.rs`**: Registered `symlink_syscall` at syscall number 88 and `symlinkat_syscall` at 266.

### glibc
- **`src/glibc/lind_syscall/lind_syscall_num.h`**: Added `SYMLINK_SYSCALL 88` and `SYMLINKAT_SYSCALL 266`.
- **`src/glibc/sysdeps/unix/sysv/linux/symlink.c`**: Replaced original implementation with `MAKE_LEGACY_SYSCALL` to route through the 3i dispatcher. Uses `TRANSLATE_GUEST_POINTER_TO_HOST` for pointer translation.
- **`src/glibc/sysdeps/unix/sysv/linux/symlinkat.c`**: Created new file (previously only a stub existed in `io/symlinkat.c` returning `ENOSYS`). Routes through 3i dispatcher with pointer translation for both path arguments.

### Tests
- **`tests/unit-tests/file_tests/deterministic/symlink.c`**: Verifies `symlink` and `symlinkat` create symbolic links correctly, validates target via `readlink`, verifies `lstat` detects symlink type, confirms read-through works, and checks `EEXIST` error case.

## How it works
glibc's `ln -s` calls `__symlink()` which now uses `MAKE_LEGACY_SYSCALL` to route through the 3i dispatcher to the `symlink_syscall` Rust handler. The handler converts the path arguments and delegates directly to `libc::symlink` on the host. `symlinkat` follows the same pattern, mirroring the existing `readlinkat` implementation.

## Impact
- Enables symbolic link creation (`ln -s`) in the Lind sandbox
- Enables any application that calls `symlink`/`symlinkat`

## Related
- Closes #993


src/glibc/sysdeps/unix/sysv/linux/symlinkat.c
```diff
@@ -0,0 +1,38 @@
...
+int
+symlinkat (const char *from, int fd, const char *to)
+{
+  uint64_t host_from = TRANSLATE_GUEST_POINTER_TO_HOST(from);
+  uint64_t host_to = TRANSLATE_GUEST_POINTER_TO_HOST(to);
+
+  return MAKE_LEGACY_SYSCALL(SYMLINKAT_SYSCALL, "syscall|symlinkat",
+      host_from, (uint64_t) fd, host_to,
+      NOTUSED, NOTUSED, NOTUSED, TRANSLATE_ERRNO_ON);
+}
```
Yaxuan-w: add new line

src/rawposix/src/fs_calls.rs
```diff
@@ -4461,3 +4523,165 @@ pub extern "C" fn getrandom_syscall(
...
+        && sc_unusedarg(arg5, arg5_cageid)
+        && sc_unusedarg(arg6, arg6_cageid))
+    {
+        panic!(
+            "{}: unused arguments contain unexpected values -- security violation",
+            "symlinkat_syscall"
+        );
+    }
+
+    let ret = if virtual_fd == libc::AT_FDCWD {
```
Yaxuan-w: Use imports from sysdefs

Yaxuan-w (changes requested):

tests/testfiles/lstat_link
rennergade: these are empty files and shouldnt be commited right?

vidyalakshmir:
> I think this is basically ready but a handful of nits need to be fixed
> 
> * fix merge conflicts
> * remove empty file commits
> * fix trailing newlines
> * rest of alices comments need to be resolved

@moogchi

tests/testfiles/lstat_link
moogchi: The empty files are intentional, they're needed as test targets for the symlink/lstat tests. The test is checking that we can create a symlink pointing to an existing file, so the file needs to be there in the repo. I added a line of comment in those files so its not an empty file.

vidyalakshmir: @moogchi Would it be better if you create those files within the tests and then test the symlink and then remove those after the test?

rennergade (approved):

vidyalakshmir (approved):

PR #1003, opened 2026-04-05
Fix readdir by converting dirent64 records into plain dirent records

### LINKED ISSUE #956: lind-bash: glob is blocking - fix readdir
I was trying to run bash's own testsuite but it looks glob feature is currently broken:
`lind-wasm bin/bash -c "cd bash_tests; ls run-*"` will hang, while `lind-wasm bin/bash -c "cd bash_tests; ls"` is working

DanielZ224:
Fixes #956

## Summary

This PR fixes the broken plain `readdir` path in Lind by converting returned `dirent64`-style records into plain `struct dirent` records inside glibc's Linux `readdir` implementation.

Previously:

- plain `readdir` was broken in Lind
- `readdir64` worked correctly
- `readdir` also worked when compiled with `_FILE_OFFSET_BITS=64`

After this change:

- plain `readdir` works correctly
- `readdir_basic.cwasm` prints directory entries normally without garbled output

## Root Cause

Debugging showed that Lind's low-level path ultimately receives valid `getdents64` / `dirent64`-style records.

We verified that:

- the raw buffer returned by the low-level syscall was valid
- parsing the first record as `dirent64` inside `readdir.c` was also valid
- however, simply casting `dirent64 *` to `dirent *` was not sufficient; the caller still observed garbled output

So the issue was not in the low-level syscall return itself, but in the mismatch between the returned `dirent64`-style data and the plain `readdir` caller expectation.

## Fix

Instead of returning a casted pointer, this patch performs an explicit field-by-field conversion from `struct dirent64` to plain `struct dirent` in `src/glibc/sysdeps/unix/sysv/linux/readdir.c`.

The updated logic:

- interprets the current record as `struct dirent64`
- stores source fields in local variables first
- writes the converted values into a plain `struct dirent`
- advances `dirp->offset` and `dirp->filepos` using the source record metadata
- returns the converted plain `struct dirent *`

This avoids the previous broken behavior where a plain cast did not provide a correct plain `readdir` result.

## Validation

We validated the fix in several stages:

1. Reproduced the original issue
   - plain `readdir` failed in Lind
   - `readdir64` worked
   - plain `readdir` with `_FILE_OFFSET_BITS=64` also worked

2. Verified low-level behavior
   - the low-level path returned valid `getdents64` records
   - the first returned record decoded correctly

3. Verified `readdir.c` parsing
   - parsing the returned buffer as `dirent64` inside `readdir.c` was correct
   - a plain cast still produced garbled output in the caller

4. Verified the conversion-based fix
   - after field-by-field conversion, `readdir_basic.cwasm` printed directory entries correctly
   - after removing temporary debug output, the test continued to pass

Example successful output from `readdir_basic.cwasm`:

- `.`
- `etc`
- `sbin`
- `var`
- `testfiles`
- `..`
- `home`

## Tests

This PR also includes three minimal test programs used during reproduction and validation:

- `readdir_basic.c`
- `readdir64_basic.c`
- `readdir_fob64_basic.c`

## Files Changed

- `src/glibc/sysdeps/unix/sysv/linux/readdir.c`

Temporary debugging changes used during investigation were removed before finalizing this patch.

## Notes

This patch fixes the current plain `readdir` behavior by converting `dirent64` records into plain `dirent` records at the `readdir.c` level.


Yaxuan-w (changes requested):
Could we also include test case for this?

Yaxuan-w (dismissed):
Awesome! Thanks

qianxichen233 (dismissed):

qianxichen233 (dismissed):

src/glibc/sysdeps/unix/sysv/linux/readdir.c
```diff
@@ -25,37 +26,62 @@
...
+  char src_name[256];
+
+  snprintf(src_name, sizeof(src_name), "%s", src->d_name);
+
+  /* Convert the dirent64 record into a plain dirent record.  */
+  dst->d_ino = (__ino_t) src_ino;
+  dst->d_off = (__off_t) src_off;
+  dst->d_reclen = src_reclen;
+  dst->d_type = src_type;
+  snprintf(dst->d_name, sizeof(dst->d_name), "%s", src_name);
```
rennergade: I think using memmove instead of snprintf for this makes more sense.

src/glibc/sysdeps/unix/sysv/linux/readdir.c
```diff
@@ -25,37 +26,62 @@
...
+     rewriting does not interfere with reading the original record.  */
+  /*TODO: directing casting 64bit field to 32bit field is not safe and could cause potential problems. 
+     It is rare to have d_ino and d_off to be large enough to cause the problem but should still might need to be fixed in the future*/
+  __ino64_t src_ino = src->d_ino;
+  __off64_t src_off = src->d_off;
+  unsigned short src_reclen = src->d_reclen;
+  unsigned char src_type = src->d_type;
+  char src_name[256];
+
+  snprintf(src_name, sizeof(src_name), "%s", src->d_name);
```
rennergade: Like I think you can just memmove d_name from src to dst

src/glibc/sysdeps/unix/sysv/linux/readdir.c
```diff
@@ -17,6 +17,7 @@
    <https://www.gnu.org/licenses/>.  */
 
 #include <dirent.h>
+#include <stdio.h>
```
rennergade: can remove if you replace snprintf

src/glibc/sysdeps/unix/sysv/linux/readdir.c
```diff
@@ -25,37 +26,62 @@
...
+  unsigned short src_reclen = src->d_reclen;
+  unsigned char src_type = src->d_type;
+  char src_name[256];
+
+  snprintf(src_name, sizeof(src_name), "%s", src->d_name);
+
+  /* Convert the dirent64 record into a plain dirent record.  */
+  dst->d_ino = (__ino_t) src_ino;
+  dst->d_off = (__off_t) src_off;
+  dst->d_reclen = src_reclen;
```
rennergade: this isn't adjusted so its technically wrong I think? Probably won't affect many things but worth being correct

rennergade (changes requested):
This is pretty well done, left some minor comments that I think are worth fixing

rennergade (approved):

PR #1019, opened 2026-04-08
optimize compiled grate wasm bin with static flag

drapl0n:
Default optimizations on the grate's compiled wasm binary are done by passing additional flags to `wasm-opt` that are required for dynamic building. This results in export related issues for building rust grates, thus adding `-s` flag to `lind_compile --opt-only` optimizes the binary for static build.


stupendoussuperpowers:
Do we only need to pass this for opt and not also for compile?

Yaxuan-w:
> Do we only need to pass this for opt and not also for compile?

@stupendoussuperpowers we only use lind_compile for pre-compilation which should be same on both build

stupendoussuperpowers (approved):

Yaxuan-w (approved):

rennergade (approved):

PR #1023, opened 2026-04-08
Fix race condition when decreasing fdcount

### LINKED ISSUE #988: test: getpid_syscall.c occasionally fail
`getpid_syscall.c` under process_tests would occasionally fail with below error message:
```
thread 'main' (60609) panicked at /home/lind/lind-wasm/src/wasmtime/crates/fdtables/src/dashmaparrayglobal.rs:449:46:
called `Option::unwrap()` on a `None` value
```
The chance is around 1 in 10

### LINKED ISSUE #1018: FDCOUNT race condition in multi-cage scenario leads to inconsistent refcount and panic
In a multi-cage execution scenario, `FDCOUNT` (a global `DashMap<(u32, u64), u64>`) can become inconsistent due to non-atomic read-modify-write operations. This leads to incorrect reference counts and eventually causes panics (e.g., decrement on missing key mentioned in #988).

Although DashMap provides thread-safe access at the operation level, the current implementation of `_increment_fdcount` and `_decrement_fdcount` performs multi-step updates (get -> compute -> insert) which are not atomic.

```rust
let newcount:u64 = FDCOUNT.get(&mytuple).unwrap().value() - 1;
```

Under concurrent operations across cages, the same `(fdkind, underfd)` key may exhibit inconsistent updates.

Example log:

```txt
[_decrement_fdcount] Updating count for key (0, 0) to 1
[_increment_fdcount] Incrementing count for key (0, 0) from 2 to 3
...
[_increment_fdcount] Incrementing count for key (0, 0) from 1 to 2
...
```

The count for (0,0) was updated to 1 but a concurrent increment observed the old value 2, resulting in inconsistent state transitions.

Yaxuan-w:
Resolves #988 
Resolves #1018 

Note: muthashmaxglobal and vanillaglobal have problems on their implementation. Need further debugging. So only fix dash map now.

### DIFF
--- a/src/fdtables/src/dashmaparrayglobal.rs
+++ b/src/fdtables/src/dashmaparrayglobal.rs
@@ -458,29 +458,39 @@ fn _decrement_fdcount(entry: FDTableEntry) {
     }
     drop(closehandlers);
 
-    let mut call_last = false;
     let newcount;
+    let call_last;
 
-    if let Some(mut count) = FDCOUNT.get_mut(&mytuple) {
-        let old = *count;
-        newcount = old.checked_sub(1).unwrap_or_else(|| {
-            panic!("FDCOUNT underflow for key {:?}", mytuple);
-        });
+    match FDCOUNT.entry(mytuple) {
+        dashmap::mapref::entry::Entry::Occupied(mut occ) => {
+            let old = *occ.get();
+            newcount = old.checked_sub(1).unwrap_or_else(|| {
+                panic!("FDCOUNT underflow for key {:?}", mytuple);
+            });
 
-        if newcount > 0 {
-            *count = newcount;
-        } else {
-            *count = 0;
-            call_last = true;
+            if newcount > 0 {
+                *occ.get_mut() = newcount;
+                call_last = false;
+            } else {
+                occ.remove();
+                call_last = true;
+            }
         }
-    } else {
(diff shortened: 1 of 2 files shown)


src/fdtables/src/dashmaparrayglobal.rs
```diff
@@ -442,40 +442,46 @@ pub fn register_close_handlers(fdkind:u32, intermediate: fn(FDTableEntry,u64), l
...
+
+    if let Some(mut count) = FDCOUNT.get_mut(&mytuple) {
+        let old = *count;
+        newcount = old.checked_sub(1).unwrap_or_else(|| {
+            panic!("FDCOUNT underflow for key {:?}", mytuple);
+        });
+
+        if newcount > 0 {
+            *count = newcount;
+        } else {
```
rennergade: When newcount == 0, it sets *count = 0 then drops the guard, then calls FDCOUNT.remove(&mytuple) separately. There's a small window between dropping the 
  get_mut guard and the remove where another thread could see the key with count 0. The old code had the same issue (FDCOUNT.remove was also outside the guard). I think that problem was there before basically but probably needs to be addressed

rennergade (changes requested):
One small comment

rennergade (approved):

PR #1032, opened 2026-04-10
Create smoke test for dlopen

### LINKED ISSUE #1004: support dlopen related test in CI
To test dynamic loading, there are two features need to be tested, one is preloading library, and another one is library loaded by `dlopen` functions. Preloading library can be implicitly tested by running the normal unit testsuite with glibc dynamically loaded, but to test `dlopen`, some dedicated pipeline might be needed in CI to make it work.

A ideal `dlopen` test would consist of a main module (e.g. main.c) and some libraries (e.g. lib.c), I already made `lind_compile` being able to compile both target: `lind_compile main.c` for building main module and `lind_compile --compile-library lib.c` for building library module (note that `lind_compile` is doing dynamic build by default right now). Therefore, to run a dlopen related test, the library needs to be first compiled with `--compile-library` flag, after then we can run the test correctly

tashabits:
### Why

We need explicit CI signal for dynamic library loading via `dlopen`, while keeping the current harness architecture stable.

The existing unit harness (`scripts/test_runner.py` + `scripts/harnesses/wasmtestreport.py`) is still the main path, but `dlopen` positive-path validation needs a two-phase compile sequence:

1. compile library with `--compile-library`
2. compile main binary
3. run main and verify it loaded the library successfully

### What Changed

- Added a dedicated script: `scripts/dlopen_smoke_test.sh`
- Integrated one call to it in the existing `make test` command chain
- Added `reports/dlopen.json` as a normal report artifact consumed by existing report checks
- Kept existing harness scripts unchanged

### How the `dlopen` Smoke Test Works

The script executes four explicit checks and logs each one in JSON:

1. `compile_library`
2. `compile_main`
3. `run_main`
4. `verify_output_substring`

For each check, the script records:

- `name`
- `status` (`pass`, `fail`, or `skipped`)
- `command`
- `output` (captured stdout/stderr or skip reason)

At the end it writes a summary:

- `summary.total`
- `summary.passed`
- `summary.failed`
- `summary.skipped`

Top-level status is reflected as:

- `status: pass|fail`
- `number_of_failures: <failed count>`

### Expected Success Condition

The smoke test is considered successful when:

- library compilation succeeds
- main compilation succeeds
- main execution succeeds
- runtime output contains:

`Hello, main module! (from shared library)`

### How It Folds into Existing `make test`

This is integrated directly into the current `Makefile` `test` target sequence:

1. run unified harness (`python3 ./scripts/test_runner.py --export-report report.html`)
2. run `./scripts/dlopen_smoke_test.sh`
3. continue with existing JSON scan + `scripts/check_reports.py`

This keeps behavior consistent with current E2E flow and avoids creating a separate CI path.

### CI / Buildx Behavior

No workflow-level changes are required.

The normal E2E build command still works and now includes the new signal:

```bash
docker buildx build \
  --platform linux/amd64 \
  --file Docker/Dockerfile.e2e \
  --tag e2e:latest \
  --output type=local,dest=test-reports \
  .
```

Artifacts now include:

- `test-reports/reports/dlopen.json`
- existing reports (`wasm.json`, `grates.json`, `report.html`, etc.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,7 @@ test: lindfs
 	fi; \
 	if LIND_WASM_BASE=. LINDFS_ROOT=$(LINDFS_ROOT) \
 	python3 ./scripts/test_runner.py --export-report report.html && \
+	./scripts/dlopen_smoke_test.sh && \
 	find reports -maxdepth 1 -name '*.json' -print -exec cat {} \; && \
 	if [ "$(LIND_DEBUG)" = "1" ]; then \
 	  python3 ./scripts/check_reports.py --debug; \
@@ -119,6 +120,7 @@ test: lindfs
 	  if [ ! -f reports/report.html ]; then cp report.html reports/report.html; fi; \
 	  if [ ! -f reports/wasm.json ]; then printf '%s\n' '{"number_of_failures":1,"results":[],"error":"missing wasm report"}' > reports/wasm.json; fi; \
 	  if [ ! -f reports/grates.json ]; then printf '%s\n' '{"number_of_failures":1,"results":[],"error":"missing grate report"}' > reports/grates.json; fi; \
+	  if [ ! -f reports/dlopen.json ]; then printf '%s\n' '{"number_of_failures":1,"results":[],"error":"missing dlopen report"}' > reports/dlopen.json; fi; \
 	fi; \
 	exit 0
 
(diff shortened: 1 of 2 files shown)


qianxichen233:
a couple of more dlopen related tests are added to main, could you take a look and try to make them compatible with the pipeline as well? feel free to reorganize the file structure in dylink tests if you want (e.g. split library file and main module file into seperate directory)

qianxichen233:
resolved by https://github.com/Lind-Project/lind-wasm/pull/1218

PR #1035, opened 2026-04-10
Fix inline random for Arc4random

rennergade:
arc4random fails to get entropy                                                                                                                                                   
                
__getrandom_nocancel in not-cancel.h used INLINE_SYSCALL_CALL(getrandom, ...) which bypasses 3i and fails in WASM. arc4random calls this internally and hits "Fatal glibc error:   cannot get entropy for arc4random".
                                                                                                                                                                                     
Fix: route __getrandom_nocancel and __getrandom_nocancel_nostatus through __getrandom() which uses MAKE_LEGACY_SYSCALL to reach RawPOSIX.


rennergade:
I removed the test here. Our test suite doesnt support the correct flags to trigger this edge case but i confirmed this fixes clang.

Yaxuan-w (approved):

Issue #1028, opened 2026-04-08
dlopen with multiprocess/multithread


qianxichen233:
plans to implement support for dlopen in multiprocess/multithreads environment
1. refactor GOT handling, each process holds a new GOT. Modify wasm global clone mechanism to update GOT correctly. This should allow dlopen works for new child
2. Record dlopen-ed libraries into a vector, similar to preloaded module vector. Replay the loading of these dlopen-ed libraries during process creation and thread creation. This should correctly sync up dlopen-ed libraries across threads/processes
3. In order to sync up new dlopen in multithreads environment, use epoch infrastructure to interrupt all other threads within the process when one thread does dlopen, force all other threads replay the same dlopen call, and caller thread only return from dlopen if all other threads finished replayed dlopen
4. Fix dlclose. Correctly unload closed library from Linker and GOT, refactor dlopen-ed library replay mechanism so that it still ensures replayability with libraries dlclosed. This is the hardest one, probably need to completely rewrite the data structure used in Linker.

1 and 2 is going to be a PR, 3 alone is going to be a PR and 4 alone is going to be a PR.

1 and 2 is the most important, 3 is not common but might still needs to implement. 4 isn't going to cause problem in most of the situations

Therefore, I plan to fix 1 and 2, then possibly 3, and probably want to leave 4 for now

rennergade:
Awesome sounds like a good plan. For 2 I think order could matter so probably worth using a data structure that captures that

PR #1037, opened 2026-04-10
 dlopen support for fork and thread

qianxichen233:
fix task 1 and task 2 in #1028
                                                                                                                                                                             
  What's fixed                                
                                                                                                                                                                             
  - dlopen'd libraries are now correctly synced to child processes (fork) and threads (pthread_create). Before this, forked children and spawned threads could not call functions from libraries opened via dlopen in the parent.                                                                                                                  
  - GOT handling refactored so each child process gets its own GOT, with global values correctly transferred from the parent via snapshots.

  How
                                                                                                                                                                             
  - dlopen_modules vec added to LindCtx to record every dlopen'd library. On fork and thread creation, these are replayed into the child store (analogous to how preloaded libraries were already replayed).       
  - get_global_snapshot changed from a positional Vec to a HashMap keyed by wasm module name, eliminating fragile manual index tracking and backup-instance skipping. Snapshot lookup in child instantiation paths uses the module's intrinsic name directly.                                                                                    
  - get_global_snapshot further optimized to iterate only registered named instances (named_module_instances field on StoreOpaque) instead of all INSTANCE_NUMBER slots, which could be up to 5000 due to backup instance.

  Tests added
                                                                                                                                                                             
  - dlopen_fork.c — dlopen in parent, fork, child uses the library
  - fork_dlopen.c — fork first, then dlopen independently in each process                                                                                                    
  - dlopen_thread.c — dlopen in main thread, spawned thread calls via inherited pointer and re-resolves via dlsym
  - fork_fork_dlopen.c — two-level fork (grandchild) to verify recursive replay correctness


src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -166,11 +175,19 @@ impl<
...
         self.linker = Some(linker);
     }
 
+    pub fn attach_got_table(&mut self, got_table: Option<LindGOT>) {
+        if let Some(got) = got_table {
+            self.got_table = Some(Arc::new(Mutex::new(got)));
+        }
+    }
+
+    pub fn append_module(&mut self, path: String, module: Module) {
```
Yaxuan-w, line 196: Need comments for those two calls

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -897,8 +1013,57 @@ impl<
                         .instantiate_with_lind_thread(&mut store, &module, false)
                         .unwrap();
 
-                    let snapshots = &global_snapshots[global_snapshots_index].1;
-                    instance.apply_global_snapshots(&mut store, snapshots);
+                    let main_module_name = module.name().unwrap();
```
Yaxuan-w: Need comments on workflow here

rennergade:
Looks really good. Only comment is there's a bunch of unwraps that seems like they may be better off as debug_panics?

Basically anywhere we do module.name().unwrap():
  - lind-multi-process/src/lib.rs:461                                                                                                                                                  
  - lind-multi-process/src/lib.rs:471                                                                                                                                                  
  - lind-multi-process/src/lib.rs:525                                                                                                                                                  
  - lind-multi-process/src/lib.rs:567                                                                                                                                                  
  - lind-multi-process/src/lib.rs:998
  - lind-multi-process/src/lib.rs:1006
  - lind-multi-process/src/lib.rs:1045                                                                                                                                                 
  - lind-multi-process/src/lib.rs:10

rennergade (approved):

PR #1058, opened 2026-04-13
implement preadv and pwritev syscalls

rennergade:
Adds preadv/pwritev support (syscall 295/296) following the same pattern as the existing readv/writev and pread/pwrite implementations.

glibc: replace SYSCALL_CANCEL with MAKE_LEGACY_SYSCALL in preadv64.c and pwritev64.c, using __lind_translate_iov to translate iov_base pointers from guest to host addresses.

rawposix: add preadv_syscall and pwritev_syscall handlers.

test: preadv_pwritev.c covers scatter/gather at offset 0 and non-zero offset, and verifies the file pointer is not modified.


Yaxuan-w:
Need to pass linter

rishabhBudhouliya:
Looks good to me, I am testing it against the postgres build and will report back the result.

rishabhBudhouliya:
@rennergade I was able to run postgres further without hitting the infinite loop bug at postgres. Thanks for making the change.

PR #1059, opened 2026-04-14
sync fdtables implementations and apply rustfmt

### LINKED ISSUE #971: fdtables: include `src/*global.rs` implementation in the module tree for LSP/linting support.
Since files such as `dashmaparrayglobal.rs` or `muthashmaxglobal.rs` etc. are not directly included in the module's include tree, `cargo fmt` ignores these files. 

None of these files are properly linted, which can be checked using `rustfmt --check /path/to/file`. 

We should add a way to automate including these to linters as well, either through manual rustfmt checks, or coming up with a way to include these files in the module include tree.

### LINKED ISSUE #1038: fdtables: muthashmax / vanillaglobal implementations are currently not compilable and out of sync with dashmap updates
The fdtables module currently contains multiple implementations. However, (1) muthashmax and vanillaglobal implementations are not compilable in the current codebase, and (2) recent updates made to the dashmap-based implementation have not been propagated to these alternative algorithms

All fdtables implementations should compile successfully, are API-compatible with the current dashmap implementation, and reflect the same semantics for fd management.

rennergade:
Update muthashmaxglobal.rs and vanillaglobal.rs to match the current dashmaparrayglobal.rs API: FDTableEntry field renames (realfd ->underfd, optionalinfo -> perfdinfo), constant renames (NO_REAL_FD removed, INVALID_FD -> FDT_INVALID_FD, EPOLLFD -> FDT_KINDEPOLL), per-fdkind close handler architecture, updated select/poll/epoll functions, and corrected doc include references.

Apply rustfmt to all four implementation files.

All implementations pass the full test suite (24/24 unit tests).

Fixes #1038
Fixes #971

### DIFF
--- a/src/fdtables/src/dashmaparrayglobal.rs
+++ b/src/fdtables/src/dashmaparrayglobal.rs
@@ -1,4 +1,4 @@
-//  DashMap<u64,[Option<FDTableEntry>;FD_PER_PROCESSS_MAX]>  Space is ~24KB 
+//  DashMap<u64,[Option<FDTableEntry>;FD_PER_PROCESSS_MAX]>  Space is ~24KB
 //  per cage w/ 1024 fds?!?
 //      Static DashMap.  Let's see if having the FDTableEntries be a static
 //      array is any faster...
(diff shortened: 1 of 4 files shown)


rennergade:
For ease of review the only changes to the dashmap impls are from running cargo fmt, I don't believe they need an actual review.

stupendoussuperpowers:
Yet to review but:

instead of using `include_str` and `current_impl` can we just use cfg with feature flags to switch between implementations? that makes sure that LSPs and cargo fmt can see these files/functions

rennergade:
btw I tested these and ran the full test suite for both the fixed impls, no breaking changes

rennergade:
> Yet to review but:
> 
> instead of using `include_str` and `current_impl` can we just use cfg with feature flags to switch between implementations? that makes sure that LSPs and cargo fmt can see these files/functions

that makes sense, can you post this as a separate issue? can get some feedback and then fix that with another minor PR if the group wants it

stupendoussuperpowers (approved):

Yaxuan-w (approved):

Issue #1065, opened 2026-04-15
Install guard page at address 0 and fix invalid NULL-pointer writes in glibc


qianxichen233:
## Background

The Lind memory layout reserves address range `[0, GUARD_SIZE)` (currently 4 KB) as a
guard page. Any memory access to this region should trap to catch null-pointer bugs.
However, this guard page has never been enforced — the vmmap entry for it was not
actually registered, so accesses to addresses near 0 silently succeeded.

A recent refactor ([`early_init_stack`](https://github.com/Lind-Project/lind-wasm/pull/1066)) attempts to properly map this region so that
accesses to it will fault. This exposed several places in the lind-glibc that perform
writes through uninitialized (NULL) pointers of the form:

```c
libc_struct->field = val;  // libc_struct is NULL
```

These writes land at small positive offsets from address 0 (i.e. within the guard
page). They went unnoticed before because the permissive memory management never
faulted on them, and the broken structs happened not to cause visible failures in
practice.

## Goal

1. Properly install the guard page at startup so that any access to `[0, GUARD_SIZE)`
   traps immediately.
2. Audit and fix all glibc sites that perform invalid writes through NULL pointers.

## Known offending sites (identified so far)

- **`__libc_setup_tls` (`csu/libc-tls.c`)** — updates DTV fields via a NULL struct
  pointer. Lind-wasm does not use the DTV mechanism, so this block was already removed
  in a prior cleanup (see `libc-tls.c`). This site is resolved.

- **`__libc_fork`** — at least one statement inside `__libc_fork` triggers a trap once
  the guard page is enforced. The exact statement has not yet been identified. Needs
  investigation.

- **Potentially more** — these are only the two sites surfaced by the initial attempt.
  A full audit of glibc initialization and fork paths is likely needed.

PR #1066, opened 2026-04-15
early alloc stack and refactor brk

qianxichen233:
## Summary

This PR is a chain of fixes, each motivated by the previous:

### 1. Early-initialize the stack before library instantiation

Library constructors (`__attribute__((constructor))`) run during each shared library's
instantiation, before the main module is instantiated and before the normal stack setup.
If a constructor uses `__stack_pointer`, it faults because the stack memory is not yet
backed by an mmap. `early_init_stack` is introduced to mmap the stack region before any
library instantiation begins, ensuring constructors can safely use the stack.

### 2. Refactor `brk_syscall` — a consequence of the above

Previously, stack and data were allocated together in a single mmap, so the vmmap
always had one large chunk starting at page 0. `brk` relied on this by looking up the
heap via a hardcoded `HEAP_INDEX=0`. Splitting stack and data into two separate mmaps
(step 1) breaks this assumption. `brk_syscall` is refactored to remove the hardcoded
index: `vmmap` now stores a `heap_start` page number (fixed at init time to the page
right after the data section), and `brk_syscall` locates the heap entry by that page
number instead. An empty heap entry is synthesized on the first `brk` call so the
logic is uniform regardless of whether any heap pages have been allocated yet.

### 3. Attempt to properly install the guard page — partially done, tracked in #1065

With the stack correctly registered in the vmmap, the natural next step is to enforce
the guard page at `[0, GUARD_SIZE)` so that NULL-pointer dereferences trap immediately.
`GUARD_SIZE` is also corrected from 1 KB to 4 KB since the vmmap operates at page
granularity and the guard must be at least one page. However, enforcing the guard
exposed several places in glibc that write through NULL pointers to fields of
uninitialized structs (writes land at small positive offsets from address 0). One
identified site — DTV initialization in `__libc_setup_tls` — is fixed here by removing
the dead DTV code, since lind-wasm does not use the DTV mechanism. The remaining sites
are tracked in #1065. Until all are resolved, `early_init_stack` maps `[0, stack_end)`
(including the guard region) rather than only the stack, as noted in the TODO comment.


src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -1993,6 +1947,39 @@ pub fn attach_shared_memory<
...
+        cageid,
+        stack_end as u64, // length: guard page + stack
+        cageid,
+        (typemap::PROT_READ | typemap::PROT_WRITE) as u64,
+        cageid,
+        (typemap::MAP_PRIVATE | typemap::MAP_ANONYMOUS | typemap::MAP_FIXED) as u64,
+        cageid,
+        // we need to pass -1 here, but since make_syscall only accepts u64
+        // and rust does not directly allow things like -1 as u64, so we end up with this weird thing
+        (0 - 1) as u64,
```
rennergade: u64::MAX or !0u64 might be less ugly here, not sure it matters

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -1993,6 +1947,39 @@ pub fn attach_shared_memory<
...
+        (MMAP_SYSCALL) as u64, // syscall num
+        0, // since wasmtime operates with lower level memory, it always interacts with underlying os
+        cageid, // target cageid (should be same)
+        0, // map from address 0 (includes the leading guard page per the TODO above)
+        cageid,
+        stack_end as u64, // length: guard page + stack
+        cageid,
+        (typemap::PROT_READ | typemap::PROT_WRITE) as u64,
+        cageid,
+        (typemap::MAP_PRIVATE | typemap::MAP_ANONYMOUS | typemap::MAP_FIXED) as u64,
```
rennergade, line 2234: some context about the flags for future readers would probably help

rennergade (changes requested):
some nits, this is great though.

Would be good to clearly mark the additional fork fixes with a TODO as well.

Yaxuan-w:
A few questions:

1. Is current bug influencing static linking as well?
2. What would be the potential influences of not finishing the todo now? (ie: memory leak / etc.)

qianxichen233:
> A few questions:
> 
> 1. Is current bug influencing static linking as well?
> 2. What would be the potential influences of not finishing the todo now? (ie: memory leak / etc.)

1. if you are talking about early alloc stack bug, no. if you are talking about invalid NULL-pointer writes in glibc issue, yes
2. the guard page isn't enforced right now, which means write to a NULL pointer will sliently success. There are a couple of places in glibc that already did this, which implys potential bugs and might cause some problem in some scenario

rennergade (approved):

Yaxuan-w (approved):

PR #1073, opened 2026-04-15
add static build CI tests

### LINKED ISSUE #1040: CI: add a minimal testsuite for static build wasm binary
as we are switching to dynamic build by default, all the unit tests in CI are currently only tested with dynamic build. However, we should still make sure static build are not broken and can still run.

we can add a new testing type under unit test called `static`, and only contains a few necessary tests to ensure static build is working (e.g. a fork test, a thread test, a tls test, etc). The test under `static` folder also requires special CI test pipeline since they needs to be built with `--static` flag in `lind_compile`

qianxichen233:
fix https://github.com/Lind-Project/lind-wasm/issues/1040

- Add static_tests category under tests/unit-tests/ with deterministic tests for fork, threads, and TLS to cover static WASM builds
- Add statictestreport harness that runs wasmtestreport with --static and -pthread flags; results are included in the combined report.html without generating a separate artifact file
- Add --static flag to wasmtestreport.py (inserts --static before the source file in lind_compile invocations)
- Refactor test_runner.py so harnesses can contribute HTML to the combined report.html purely in memory, without requiring a separate html_filename artifact

### DIFF
--- a/scripts/harnesses/statictestreport.py
+++ b/scripts/harnesses/statictestreport.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+"""Static build test harness.
+
+Runs the unit tests under tests/unit-tests/static_tests/ compiled with the
+--static flag so that static WASM binaries (no dynamic linking) are tested.
+This is a thin wrapper around wasmtestreport.py that pre-sets the static
+compilation flags.
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+import tempfile
+from pathlib import Path
+from typing import Any, Callable
+
+SCRIPT_DIR = Path(__file__).resolve().parent
+REPO_ROOT = SCRIPT_DIR.parents[1]
+WASMTESTREPORT = SCRIPT_DIR / "wasmtestreport.py"
+
+# Flags forwarded to wasmtestreport for static builds:
+#   --run static_tests            : only run tests under static_tests/
+#   --static                      : pass --static before source file in lind_compile
+#   --allow-pre-compiled          : use .cwasm AOT binaries (consistent with dynamic harness)
+#   --compile-flags -pthread -lpthread : link pthread for thread/TLS tests
+_STATIC_HARNESS_ARGS = [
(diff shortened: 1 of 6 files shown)


scripts/harnesses/wasmtestreport.py
Yaxuan-w, line 1: Can we rename this file to dynamic related?

qianxichen233: this test runner invokes `lind_compile` with no additional argument, so this is essentially doing the test with default configuration in `lind_compile`. If default configuration in `lind_compile` is changed, then what this test runner is testing is also changed. Therefore, I'd like to keep the name

scripts/harnesses/statictestreport.py
```diff
@@ -0,0 +1,83 @@
...
+REPO_ROOT = SCRIPT_DIR.parents[1]
+WASMTESTREPORT = SCRIPT_DIR / "wasmtestreport.py"
+
+# Flags forwarded to wasmtestreport for static builds:
+#   --run static_tests            : only run tests under static_tests/
+#   --static                      : pass --static before source file in lind_compile
+#   --allow-pre-compiled          : use .cwasm AOT binaries (consistent with dynamic harness)
+#   --compile-flags -pthread -lpthread : link pthread for thread/TLS tests
+_STATIC_HARNESS_ARGS = [
+    "--run", "static_tests",
```
tashabits, line 28: Since this harness now owns `static_tests`, should we also exclude `static_tests` from the default `wasmtestreport` run? As written, I believe the normal wasm harness still discovers all `tests/unit-tests/**/*.c`, so these tests look like they’ll run once as dynamic-build tests and again here as static-build tests. Would it make sense to add a small guard near the existing `if module_name == "wasmtestreport":` block, perhaps appending a flag?

qianxichen233: just resolved this, could you take a look again

rennergade (approved):

Issue #1052, opened 2026-04-13
Unused vmmap lock in munmap path


rennergade:
`src/rawposix/src/fs_calls.rs:1108-1110`

A read lock on vmmap is acquired and immediately dropped without reading it. The munmap path should validate that the region exists before unmapping.

PR #1075, opened 2026-04-16
munmap: consult vmmap before PROT_NONE to avoid unlinking adjacent mappings

### LINKED ISSUE #1062: munmap() rounding up page logic potentially causes memory faults
## Bug
https://github.com/Lind-Project/lind-wasm/blob/c71505694a265e3a5a3812e26b0d75c9356a7959/src/rawposix/src/fs_calls.rs#L1103-L1106

While bootstrapping Postgres, postgres calls: 
```
munmap(addr=0x7b22848e6000, len=0x1a34
```
This len: **0x1a34** translates to **6,708 bytes**.  Our munmap syscall implementation rounds this up to **2 pages** and then we call: 
```
 mmap(0x7b22848e6000, 0x2000, PROT_NONE, MAP_FIXED | MAP_ANONYMOUS | MAP_PRIVATE, -1, 0)
```
This means there is overlapping memory that might be already assigned that is completely overridden by this calls behavior (MAP_FIXED). 

So effectively, lind wipes: 
```
  Page 1:  0x7b22848e6000 → 0x7b22848e7000   (what postgres asked to unmap)                                                                           
  Page 2:  0x7b22848e7000 → 0x7b22848e8000   (the SHMAT PAGE)
```

This may be the reason why we see the panic: 
```
thread 'lind-fork-2' (292848) panicked at /home/lind/lind-wasm/src/wasmtime/crates/lind-multi-process/src/lib.rs:1190:50:

exec-ed module error: failed to run main module
Caused by:
    1: memory fault at wasm address 0xffae7010 in linear memory of size 0x100000000
    2: wasm trap: out of bounds memory access
```

## Evidence

I changed the round_up to  `round_down(0x1a34) = 0x1000  (1 page, 4,096 bytes)` and didn't see the memory fault and was consequently able to make postgres run

### LINKED ISSUE #1076: shmat_syscall misinterprets any attached user address ≥ 0x80000000 as an error
# Bug
shmat_syscall misinterprets any attached user address ≥ 0x80000000 as an error, silently returning a bogus value and skipping the vmmap entry registration.

# Root Cause
https://github.com/Lind-Project/lind-wasm/blob/1b1f6ed38c9cd18ba3bc3aac6b0823cb73106b70/src/rawposix/src/fs_calls.rs#L4291-L4322

- sys_to_user returns a full u32 wasm address (src/cage/src/memory/vmmap.rs:348).                                                                   
- result as i32 >= 0 is false for any user address with the top bit set (≥ 0x80000000).
- The else branch returns the u32 as a negative i32, which glibc's shmat wrapper reinterprets as a pointer, so callers see a seemingly-valid return.

rishabhBudhouliya:
# Purpose

  - munmap_syscall blindly called mmap(MAP_FIXED|PROT_NONE) over the rounded-up byte range without checking what pages the caller actually owns. When the rounded length extended into an adjacent shm page, it got wiped.                                                                                
  - Now intersects the requested range with vmmap entries and only PROT_NONEs pages belonging to anonymous mappings. SharedMemory-backed entries are left to shmdt.                                                                                                                                      
  - Fixes the dead read-lock/drop at the top of the function (the lock is now a single write lock used for both lookup and mutation).
  
Fixes #1062  and #1052 
Also fixes: #1076 and #1085

## Concern
I was reading Linux's behavior around this un-mapping logic and it seems like for the calling process, the kernel would unlink even the shared memory reference (alongside the anonymous backed region). In our case, we simply ignore that problem. 
I suspect this could cause a divergence in the state of shm at the cage level. However, I couldn't conceptualize/didn't invest enough time to read how that state works.
I was also reading that it is unlikely for VMAs of anonymous and shared memory being adjacent to each other.

Please let me know if I should dig deeper if this fix isn't optimal for our use-case.


rennergade:
This is a really good find/PR. Thanks @rishabhBudhouliya.

It would be good to add a test that would have failed before this fix to confirm this.

As for your questions, I think the inconsistency is only visible if the same cage tries to re-access the address after munmap, which is undefined behavior anyway. I don't think we really need to worry about that.

src/glibc/sysdeps/unix/sysv/linux/mmap.c
```diff
@@ -51,12 +51,12 @@
 // }
 
 // Edit: Dennis
+// Fix: Do NOT translate addr - it's a guest virtual address (offset in WASM linear memory),
```
rennergade: this probably needs a bit more context in the comments that this is handled in rawposix and why

src/rawposix/src/fs_calls.rs
```diff
@@ -1105,44 +1105,55 @@ pub extern "C" fn munmap_syscall(
...
-        panic!(
-            "munmap: MAP_FIXED violation - mmap returned address {:p} but requested {:p}",
-            result as *const c_void, sysaddr as *const c_void
-        );
+    let req_start: u32 = vmmap.sys_to_user(rounded_addr) >> PAGESHIFT;
+    let req_end: u32 = req_start + (rounded_length as u32 >> PAGESHIFT);
+
+    // Collect owned page ranges first to release the iterator borrow before mutating.
+    // interval.end() is inclusive, so +1 converts to exclusive to match req_end.
+    let overlaps: Vec<(usize, usize)> = vmmap
```
rennergade: this is worth refactoring into a function in the vmmap i think.

src/glibc/sysdeps/unix/sysv/linux/mmap.c
```diff
@@ -51,14 +51,17 @@
...
-  
-  return MAKE_LEGACY_SYSCALL(MMAP_SYSCALL, "syscall|mmap", host_addr, (uint64_t) len, (uint64_t) prot, (uint64_t) flags, (uint64_t) fd, (uint64_t) offset, TRANSLATE_ERRNO_ON);
+  return MAKE_LEGACY_SYSCALL (MMAP_SYSCALL, "syscall|mmap",
+			      (uint64_t) (uintptr_t) addr, (uint64_t) len,
+			      (uint64_t) prot, (uint64_t) flags, (uint64_t) fd,
+			      (uint64_t) offset, TRANSLATE_ERRNO_ON);
 }
-weak_alias (__mmap, mmap)
-libc_hidden_def (__mmap)
+weak_alias (__mmap, mmap) libc_hidden_def (__mmap)
```
rennergade: these should be on separate lines

src/rawposix/src/fs_calls.rs
```diff
@@ -1105,44 +1105,45 @@ pub extern "C" fn munmap_syscall(
...
+                act_len,
+                PROT_NONE,
+                (MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED) as i32,
+                -1,
+                0,
+            ) as usize
+        };
+        if result as isize == -1 {
+            let errno = get_errno();
+            panic!(
```
rennergade: I think these should be debug panics if they should never happen? if its just an errno we want to return errno tho

rishabhBudhouliya: @rennergade could you please explain why they should be debug panics? I didn't understand the comment perfectly

rennergade: I don't really understand what's happening here, in what situations are panics being triggered?

rishabhBudhouliya: I actually copy pasted both of these error scenarios from before, that is, I modified the overlapping processing of regions logic but didn't think through the error scenarios coming out of that.

But reading through the man page,

> On failure, it returns -1, and
       [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is set to indicate the error (probably to EINVAL).
The errorno can range from underlying memory not being available to SIGSEGV or misalignment error.

rennergade: we basically should only panic if state is corrupted in some unrecoverable way, but we should use lind_debug_panic for that and not use the generic panic, if its just an errno thats normal we want to return that.

rishabhBudhouliya: Got it, I understand now that this is about a better categorization and error handling of system level failure vs kernel failure.

rennergade (dismissed):

rennergade (dismissed):

tests/unit-tests/memory_tests/deterministic/munmap_adjacent_shm.c
```diff
@@ -0,0 +1,80 @@
...
+            shmctl(shmid, IPC_RMID, NULL);
+            return 1;
+        }
+    }
+
+    printf("PASS: shm intact after unaligned munmap\n");
+
+    // Cleanup disabled: lind-wasm shmdt has null pointer bug
+    // shmdt(shm2);
+    // shmctl(shmid, IPC_RMID, NULL);
```
qianxichen233: what is this bug? is there any issue raised for it

rishabhBudhouliya: Good question, I witnessed a panic in shmdt during the execution of this test. I didn't investigate it at that time but I should and raise an issue for it if there is one.

rishabhBudhouliya: @qianxichen233 turns out I had commented this earlier and now with the double translation address bug fixed in shmdt, I didn't see any NPE while the cleanup of test.

qianxichen233 (approved):

PR #1081, opened 2026-04-17
update dylink blacklist symbol and fix an edge case for syscall retval

qianxichen233:
some misc fixes when working with python


rennergade:
can we get a more descriptive title here?

Yaxuan-w (approved):

PR #1084, opened 2026-04-17
Fix/lind debug printf clean

### LINKED ISSUE #1029: better debug support: guest space debug_printf bypassing fd
sometimes if people use printf or other print-related functions to do logs in glibc side for debugging, the log won't be visible if the process already redirect the stdout to some other file descriptor

we can add a debug_printf by passing the normal fd handling which makes sure the log is always visible as terminal output

one way to do this is adding a wrapper function in lind_debug.c, which uses snprintf to format the string and pass the formated string into lind_debug_str

DanielZ224:
## PURPOSE

This PR fixes issue #1029 by adding a guest-side `lind_debug_printf(...)` helper for glibc-side debugging.

Previously, using normal `printf`-style logging on the glibc side was not reliable for debug visibility, because guest stdout/stderr may already be redirected. This PR adds a debug-only helper that bypasses normal fd-based output and forwards the formatted message through the existing Lind debug string import.

Specifically, `lind_debug_printf(...)`:
- formats the message locally with `vsnprintf(...)`
- sends the formatted string to `__lind_debug_str(...)`

This makes the debug output visible through the host-side Lind debug channel.

## RELATED ISSUES

- Fixes #1029

## CHANGES

Files changed:
- `src/glibc/lind_syscall/lind_debug.c`
- `src/glibc/lind_syscall/lind_debug.h`

Main changes:
- added `lind_debug_printf(const char *fmt, ...)`
- declared the helper in `lind_debug.h`
- implemented the helper using:
  - `va_list`
  - `vsnprintf(...)`
  - `__lind_debug_str(...)`

## TESTING

Tested locally with a smoke test using:

```c
#ifdef LIND_DEBUG
lind_debug_printf("LIND DEBUG SMOKE: hello %d\n", 42);
#endif
``` 

Observed output:

```text
[LIND DEBUG STR]: LIND DEBUG SMOKE: hello 42
```

## NOTES FOR REVIEWERS
- Final diff is limited to `src/glibc/lind_syscall/lind_debug.c` and `src/glibc/lind_syscall/lind_debug.h`
- Local verification files were not included in this PR

### DIFF
--- a/scripts/make_glibc_and_sysroot.sh
+++ b/scripts/make_glibc_and_sysroot.sh
@@ -78,7 +78,6 @@ EXTRA_DEFINES="-include ../include/libc-symbols.h -DPIC -DTOP_NAMESPACE=glibc"
 # Check if LIND_DEBUG is defined (set by build.rs when `lind_debug` is enabled)
 if [ "$LIND_DEBUG" ]; then
   DEFINES="$DEFINES -DLIND_DEBUG"
-  CPPFLAGS="$CPPFLAGS -DLIND_DEBUG"
 fi
 
 # Build glibc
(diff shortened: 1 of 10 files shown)


rennergade (approved):
This looks good.

Yaxuan-w (approved):

PR #1088, opened 2026-04-19
Fix sandbox escape via absolute paths in *at syscalls

### LINKED ISSUE #1077: Security Bug: Sandbox Escape via Absolute Path in `symlinkat_syscall`
### Summary
The `else` branch of `symlinkat_syscall` (when a specific `dirfd` is provided) passes the raw, untranslated guest `linkpath` string directly to the host kernel. This allows a malicious guest to escape the cage sandbox by supplying an absolute path.

### Background
Per POSIX, all `*at` syscalls (e.g. `openat`, `symlinkat`, `renameat`) ignore the `dirfd` argument entirely when the provided path is absolute. The current code assumes that `dirfd` confinement is sufficient to keep the guest's path within the sandbox — but this assumption silently breaks for any path beginning with `/`.

### Vulnerable Code
```rust
// The kernel handles that resolution, so we pass the original guest pointer.
let raw_linkpath = match get_cstr(linkpath_arg) { ... }

unsafe {
    libc::symlinkat(
        target.as_ptr() as *const libc::c_char,
        kernel_fd,
        raw_linkpath.as_ptr() as *const c_char,
    )
}
```

### Attack Scenario
A malicious guest executes:
```c
symlinkat("payload", any_valid_fd, "/etc/cron.d/backdoor");
```
1. The shim extracts the raw string `/etc/cron.d/backdoor`.
2. It passes it directly to the host kernel's `symlinkat`.
3. The host kernel sees an absolute path, ignores `kernel_fd`, and creates the symlink at `/etc/cron.d/backdoor` on the **host filesystem** — fully outside the cage.

No memory corruption or exploit chain is required. A single syscall is sufficient.

### Fix
Before passing `raw_linkpath` to the host kernel, check for absolute paths and either reject them or re-route through `sc_convert_path_to_host` + `AT_FDCWD` (the same path the `dirfd == AT_FDCWD` branch already takes):

```rust
if raw_linkpath.to_bytes().starts_with(b"/") {
    return syscall_error(Errno::EACCES, "symlinkat", "absolute path bypasses dirfd sandbox");
}
```

### Other syscalls to be inspected
Any other `*at` syscall in this codebase using the same pattern — passing raw guest paths to the host kernel and relying on `dirfd` for confinement — is vulnerable to the same escape. The following should be audited:
- `openat`
- `renameat`
- `unlinkat`
- `fchownat`
- Any other `*at` variant

### Severity
High — allows an unprivileged guest process to write arbitrary symlinks to the host filesystem with no preconditions beyond a valid file descriptor.

moogchi:
## Summary

Fixes sandbox escape via absolute paths in `*at` syscalls (fixes #1077).

## Background

`symlinkat`, `openat`, `unlinkat`, and `readlinkat` all pass the raw guest path directly to the host kernel when a real `dirfd` is provided. Per POSIX, the host kernel ignores `dirfd` entirely when the path is absolute — allowing a malicious guest to escape the cage sandbox with a single syscall:

```c
symlinkat("payload", any_valid_fd, "/etc/cron.d/backdoor");
```

## Fix

Reject absolute paths in the `dirfd` branch of all four syscalls with `EACCES` before passing to the host kernel:

```rust
if raw_path.to_bytes().starts_with(b"/") {
    return syscall_error(Errno::EACCES, "symlinkat", "absolute path bypasses dirfd sandbox");
}
```

## Affected syscalls

- `symlinkat_syscall`
- `openat_syscall`
- `unlinkat_syscall`
- `readlinkat_syscall`

### DIFF
--- a/src/rawposix/src/fs_calls.rs
+++ b/src/rawposix/src/fs_calls.rs
@@ -122,6 +122,17 @@ pub extern "C" fn openat_syscall(
             Ok(p) => p,
             Err(_) => return syscall_error(Errno::EINVAL, "openat", "invalid path"),
         };
+
+        // Reject absolute paths — the host kernel ignores dirfd for absolute paths,
+        // which would allow a guest to escape the cage sandbox.
+        if raw_path.to_bytes().starts_with(b"/") {
+            return syscall_error(
+                Errno::EACCES,
+                "openat",
+                "absolute path bypasses dirfd sandbox",
+            );
+        }
+
         let c_path = match CString::new(raw_path) {
             Ok(c) => c,
             Err(_) => return syscall_error(Errno::EINVAL, "openat", "invalid path"),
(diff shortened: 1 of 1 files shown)


vidyalakshmir:
Closing this PR for now, since this issue is not relevant due to chroot mechanism in host. Discussion [here](https://github.com/Lind-Project/lind-wasm/issues/1077#issuecomment-4770780654).

PR #1090, opened 2026-04-19
implement getpgid

### LINKED ISSUE #775: getpgid
git requires the getpgid() syscall: https://github.com/Lind-Project/lind-wasm-apps/issues/74

I just made it work by stubbing getpgid to use getpid. We may just want to hardcode this if we're not going to implement process groups.

rennergade:
Lind does not implement process groups. getpgid always returns the cage's own cageid (matching getpid behavior). This is platform-specific behavior that cannot be changed with a grate since getpgid is handled directly by RawPOSIX.

Fixes https://github.com/Lind-Project/lind-wasm/issues/775


src/rawposix/src/sys_calls.rs
```diff
@@ -675,6 +675,50 @@ pub extern "C" fn getpid_syscall(
...
 }
 
+/// Reference to Linux: https://man7.org/linux/man-pages/man2/getpgid.2.html
+///
+/// Returns the process group ID of the process specified by pid.
+/// If pid is 0, returns the process group ID of the calling process.
+///
+/// Lind does not implement process groups. This always returns the cage's
+/// own cageid, which is platform-specific behavior that cannot be changed
+/// with a grate since getpgid is handled directly by RawPOSIX.
```
Yaxuan-w: Is this because cage (the platform specific) is managed at RawPOSIX layer, so even if calling grate, it'll still need to call things like getpid()?

src/glibc/sysdeps/unix/sysv/linux/i386/i686/getpgid.c
```diff
@@ -1,9 +1,14 @@
 #include <unistd.h>
+#include <sysdep-cancel.h>
```
Yaxuan-w: Seems like not used?

src/glibc/sysdeps/unix/sysv/linux/i386/i686/getpgid.c
```diff
@@ -1,9 +1,14 @@
...
+#include <syscall-template.h>
+#include <lind_syscall_num.h>
+#include <addr_translation.h>
```
Yaxuan-w: Same here

src/rawposix/src/sys_calls.rs
rennergade: I think this comment is misleading ill edit it. We don't implement process groups but we could modify the behavior here with a grate.

Yaxuan-w (approved):

rishabhBudhouliya (approved):

PR #1093, opened 2026-04-20
Add vmmap write checks to WASI environ functions

### LINKED ISSUE #1087: WASI host functions should validate guest addresses against vmmap protections
PR #1086 fixed the sign-extension bug and added bounds checks against linear memory size in the WASI host functions (args_get, environ_get, random_get, etc.). This prevents writes past the 4GB boundary.

However, the bounds check only validates that `offset + len <= memory.data_size()`. It does not check whether the target region has write permissions in the vmmap. A guest could pass a pointer into a PROT_NONE region (e.g. a munmap'd area or the guard page) and the host would write to it, bypassing the protection the vmmap is supposed to enforce.

The vmmap already exposes `check_addr_write(cageid, addr, length)` in `cage::memory` which consults the vmmap entries and verifies PROT_WRITE. The WASI host functions should call this before writing to guest memory.

Affected functions in `src/wasmtime/crates/lind-common/src/lib.rs`:
- `args_sizes_get`
- `args_get`
- `environ_sizes_get`
- `environ_get`
- `random_get`

DanielZ224:
## PURPOSE

This PR fixes issue #1087 by adding vmmap write-permission checks to WASI host functions before they write into guest memory.

PR #1086 already added bounds checks against linear memory size, which prevents writes past the 4GB boundary. However, those checks only verify that the target range is in-bounds. They do not verify whether the target guest address is actually writable according to Lind's vmmap protections.

This PR adds `check_addr_write(cageid, addr, length)` before guest-memory writes so that WASI host functions no longer write into `PROT_NONE`, guard-page, or otherwise non-writable regions.

## RELATED ISSUES

- Fixes #1087

## CHANGES

Updated the following functions in `src/wasmtime/crates/lind-common/src/lib.rs`:

- `args_sizes_get`
- `args_get`
- `environ_sizes_get`
- `environ_get`
- `random_get`

Main changes:
- kept the existing linear-memory bounds checks
- added vmmap write-permission validation with `check_addr_write(...)` before each guest-memory write

## TESTING

- Rebuilt successfully with `make lind-boot`

## NOTES FOR REVIEWERS

- This change is intentionally minimal and only updates the WASI host functions listed in the issue.
- The existing linear-memory bounds checks are preserved; this PR adds the missing vmmap permission check on top of them.

### DIFF
--- a/src/wasmtime/crates/lind-common/src/lib.rs
+++ b/src/wasmtime/crates/lind-common/src/lib.rs
@@ -326,8 +326,8 @@ fn add_environ_funcs_to_linker<
             if off_argc + 4 > mem_size || off_buf_size + 4 > mem_size {
                 return 21; // EFAULT
             }
-            if check_addr_write(cageid, ptr_argc as u32 as u64, 4).is_err()
-                || check_addr_write(cageid, ptr_buf_size as u32 as u64, 4).is_err()
+            if check_addr_write(cageid, base_u64 + off_argc as u64, 4).is_err()
+                || check_addr_write(cageid, base_u64 + off_buf_size as u64, 4).is_err()
             {
                 return 21; // EFAULT
             }
@@ -356,8 +356,8 @@ fn add_environ_funcs_to_linker<
                 if ptr_slot + 4 > mem_size || end > mem_size {
                     return 21; // EFAULT
                 }
-                if check_addr_write(cageid, ptr_slot as u64, 4).is_err()
-                    || check_addr_write(cageid, buf_offset as u64, bytes.len() + 1).is_err()
+                if check_addr_write(cageid, base_u64 + ptr_slot as u64, 4).is_err()
+                    || check_addr_write(cageid, base_u64 + buf_offset as u64, bytes.len() + 1).is_err()
                 {
                     return 21; // EFAULT
                 }
@@ -391,8 +391,8 @@ fn add_environ_funcs_to_linker<
             if off_count + 4 > mem_size || off_buf_size + 4 > mem_size {
                 return 21; // EFAULT
             }
-            if check_addr_write(cageid, ptr_count as u32 as u64, 4).is_err()
-                || check_addr_write(cageid, ptr_buf_size as u32 as u64, 4).is_err()
+            if check_addr_write(cageid, base_u64 + off_count as u64, 4).is_err()
+                || check_addr_write(cageid, base_u64 + off_buf_size as u64, 4).is_err()
             {
                 return 21; // EFAULT
             }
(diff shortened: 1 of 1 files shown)


src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -317,13 +318,19 @@ fn add_environ_funcs_to_linker<
             let cx = get_environ(caller.data());
             let argc = cx.args.len() as u32;
             let buf_size: u32 = cx.args.iter().map(|a| a.len() as u32 + 1).sum();
+            let cageid = wasmtime_lind_multi_process::current_cageid(&mut caller) as u64;
             let (base_u64, mem_size) = get_memory_base_and_size(&mut caller);
             let base = base_u64 as *mut u8;
             let off_argc = ptr_argc as u32 as usize;
             let off_buf_size = ptr_buf_size as u32 as usize;
             if off_argc + 4 > mem_size || off_buf_size + 4 > mem_size {
                 return 21; // EFAULT
```
Yaxuan-w: Could we use constants from sysdefs?

src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -317,13 +318,19 @@ fn add_environ_funcs_to_linker<
...
             }
+            if check_addr_write(cageid, base_u64 + off_argc as u64, 4).is_err()
+                || check_addr_write(cageid, base_u64 + off_buf_size as u64, 4).is_err()
+            {
+                return 21; // EFAULT
```
Yaxuan-w: Could we use constants from sysdefs?

src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -317,13 +318,19 @@ fn add_environ_funcs_to_linker<
...
             let buf_size: u32 = cx.args.iter().map(|a| a.len() as u32 + 1).sum();
+            let cageid = wasmtime_lind_multi_process::current_cageid(&mut caller) as u64;
             let (base_u64, mem_size) = get_memory_base_and_size(&mut caller);
             let base = base_u64 as *mut u8;
             let off_argc = ptr_argc as u32 as usize;
             let off_buf_size = ptr_buf_size as u32 as usize;
             if off_argc + 4 > mem_size || off_buf_size + 4 > mem_size {
                 return 21; // EFAULT
             }
+            if check_addr_write(cageid, base_u64 + off_argc as u64, 4).is_err()
```
Yaxuan-w: What does 4 mean here?

Yaxuan-w (changes requested):
Need to replace hardcode num with constants

src/wasmtime/crates/lind-common/src/lib.rs
DanielZ224: fixed!

src/wasmtime/crates/lind-common/src/lib.rs
DanielZ224: replaced using size_of_u32!

Yaxuan-w (approved):

qianxichen233 (approved):

PR #1096, opened 2026-04-21
Return `EINVAL` instead of panicking during mmap validation

rishabhBudhouliya:
<!-- Before submitting a PR, make sure:

* new code is tested
* tests pass locally
* docs are up-to-date

For details, see https://lind-project.github.io/lind-wasm/contribute/

Please describe **PURPOSE**, reference related **ISSUES**, and include any information that might be helpful to **REVIEWERS** below this line. -->

  - Return `EINVAL` instead of panicking when mmap receives unsupported flags                                                                         
  - Enables PostgreSQL to gracefully fall back when huge pages are unavailable  
  
## Background                                                                                                                                                                                                                                                                           
  PostgreSQL requests huge pages for shared memory using `MAP_HUGETLB` with size encoding:                                                            
                                                                                                                                                      
  ```c
  // sysv_shmem.c:555
  mmap_flags = MAP_HUGETLB | ((shift & MAP_HUGE_MASK) << MAP_HUGE_SHIFT);                                                                             
  // e.g., 0x54040021 for 2MB huge pages                                                                                                              
                                                                                                                                                      
  The upper bits encode the page size (21 = log2(2MB)), which lind-wasm doesn't recognize. Previously this caused a panic, killing the runtime.

PR #1097, opened 2026-04-22
return EINVAL for MAP_HUGETLB, keep debug panic for other unsupported flags

rennergade:
The previous change (PR #1096) returned EINVAL for all unsupported mmap flags. This caused valid but unimplemented flags like PROT_EXEC to silently fail with EINVAL instead of surfacing during development.

MAP_HUGETLB specifically should return EINVAL to match Linux behavior on systems without huge page support. Other unsupported flags use lind_debug_panic so they're caught during development.

### DIFF
--- a/src/rawposix/src/fs_calls.rs
+++ b/src/rawposix/src/fs_calls.rs
@@ -845,18 +845,25 @@ pub extern "C" fn mmap_syscall(
 
     let mut maxprot = PROT_READ | PROT_WRITE;
 
-    // Validate flags - only these four flags are supported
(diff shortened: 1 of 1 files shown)


qianxichen233 (approved):

Yaxuan-w (approved):

JustinCappos:
We should plausibly log something and return EINVAL for situations like
this.  Having log and exit, log and return EINVAL / ENOSYS, or don't log
and return EINVAL / ENOSYS seem like the logical options to have available.

PR #1100, opened 2026-04-22
Implement grate worker execution model with separate stack

### LINKED ISSUE #961: Redesign grate call execution model to fix Wasmtime state corruption under high concurrency
#952 and #953 observed that under high concurrency, interposing on mmap and futex can lead to failures such as memory faults and deadlocks:

```sh
Error: memory fault at wasm address 0xfff7ada8 in linear memory of size 0x100000000

Caused by:
    wasm trap: out of bounds memory access
```

Based on investigation, this issue is likely caused by the current grate call implementation under high concurrency, which allows incorrect execution contexts or concurrent re-entry to corrupt Wasmtime’s internal state. As an experiment, I increased the number of backup instances to 5000 and removed the getpid() call in the grate implementation to reduce pressure on the grate call path. These changes largely improved stability, and the race-test grate now works correctly most of the time. However, crashes still occur occasionally, for example:

```sh
thread 'lind-thread-2' (110000) panicked at /home/lind/lind-wasm/src/wasmtime/crates/wasmtime/src/runtime/func.rs:944:13:
index out of bounds: the len is 55091 but the index is 55091
```

My current understanding is that even though different `vmctx`s are used, multiple threads are still concurrently accessing the same underlying `Store` and its internal structures such as function tables, instance handles, and indices. The current implementation effectively bypasses Wasmtime’s store-level concurrency guarantees, which leads to inconsistent internal state being observed under concurrent access.

**I submitted a temporary fix in PR #960 to mitigate the issue in the current implementation, but this does not address the root cause.** 

Fundamentally, the current design treats Wasmtime’s internal execution context as a shared concurrent entry point, which conflicts with requirements for high concurrency, lifecycle safety, and state consistency. I propose redesigning the grate call execution model to address this.  The system needs to support three key properties: (1) execution can be triggered at arbitrary times from external sources, (2) execution can observe the current runtime state, and (3) the system supports high concurrency. 

Instead of re-entering Wasmtime with arbitrary `vmctx`s, each grate should correspond to a long-lived execution unit implemented as a worker pool. Each worker maintains its own valid Wasmtime execution context, including an independent `Store`, `Instance`, and typed function handles. In this model, requests are submitted to the grate’s execution unit and executed by its workers, rather than directly entering Wasmtime via a raw context pointer.

Concurrency is then controlled at the scheduling layer. For handlers that do not depend on TLS, stack state, or globally consistent shared data (for example, simple forwarding or tracing), requests can be processed in parallel. For handlers that depend on the current execution context or require globally consistent shared state during runtime (for example, accumulating data across requests or updating a global constant that must be immediately visible to subsequent requests), execution must be serialized. This behavior can be specified by the grate via `register_handler`, and the runtime scheduler decides whether to execute requests concurrently or serially based on these annotations.

In this model, there is a designated main worker (the initial execution context) responsible for lifecycle management. The main worker performs final output or post-processing and main worker's exit triggers worker termination. If the grate needs to collect data from all workers, this can be done through a fixed API. Specifically, each worker uses `grate_export_data()` to export its local state, and the main worker implements `grate_gather_data()` to consume and aggregate the collected data. During this phase, the Wasmtime runtime iterates over all workers, invokes `grate_export_data()` on each, and feeds the results into the main worker’s `grate_gather_data()` to complete aggregation and termination. The exact semantics of these functions are defined by the grate developer, while their names and signatures remain fixed to allow the runtime to orchestrate the process. In other words, the runtime is responsible only for data movement, while the grate defines the semantics.

Currently, grates such as chroot, strace, seccomp, and tmp can run fully in parallel under this model. The witness grate requires a final aggregation step to combine attestations from all workers into a final report, but its request handling can still be performed concurrently, with aggregation deferred to the final stage.

Yaxuan-w:
## Summary

Resolves #961 

This PR implements the grate-call execution model based on reusable workers.

A grate call is no longer treated as entering one shared grate instance. Instead,
each call is executed by leasing an independent worker from a per-grate handler
pool, where each worker owns its own:

- Wasmtime `Store`
- instantiated grate `Instance`
- dedicated Wasm stack slot

This PR also introduces the global grate-handler registry used to resolve the
target handler for a grate id at runtime, and documents how the stack arena is
reserved and partitioned when multiple stores attach to the same underlying
linear memory.

Grate calls are not continuation-sensitive in the same way as `fork` / `exit`.
They do not need to resume inside the exact same Wasmtime continuation context.
Instead, they only need a compatible grate execution context.

This PR makes that model explicit by representing grate execution as a pool of
independent workers. That enables concurrency while still keeping stack state
isolated even when multiple stores attach to the same linear memory.

### 1. Add `GrateHandler` / `GrateWorker` execution model
- introduce worker-based grate execution
- support `Parallel` and `Serialized` concurrency modes
- add worker leasing / return logic
- track in-flight calls and shutdown state

### 2. Add global grate-handler registry in `lind-boot`
- initialize a process-wide table of `GrateHandler<HostCtx>`
- register / look up / submit / unregister / clean up grate handlers by cage/grate id
- keep this registry in `lind-boot` because the handler type is tied to the concrete `HostCtx`

### 3. Reserve and document grate stack arena layout
- reserve stack arena space in linear memory during lind-specific instance initialization
- define per-worker stack-slot layout using:
  - `MAX_GRATE_WORKERS`
  - `GRATE_STACK_SLOT_SIZE`
  - `GRATE_STACK_GUARD_SIZE`
- record `STACK_ARENA_BASE` globally for later worker stack-slot computation
- place `STACK_ARENA_BASE` in `sysdefs` to avoid circular dependencies between
  Wasmtime, `lind-3i`, and `lind-multi-process`

## Notes

- worker stacks are isolated by disjoint stack slots inside a shared stack arena
- the global handler registry is intentionally outside `lind-3i` because the
  stored handler type depends on `HostCtx`

## Tests

**Passed simple grate call tests.** 

**Passed grates testing:**

1. IPC grates 
2. fdtables grates 
3. strace grates
4. mtls grates
5. net-namespacing

### DIFF
--- a/src/lind-boot/Cargo.toml
+++ b/src/lind-boot/Cargo.toml
@@ -8,6 +8,7 @@ disable_signals = ["cage/disable_signals", "wasmtime-lind-multi-process/disable_
 secure = ["typemap/secure"]
 lind_debug = ["wasmtime-lind-common/lind_debug"]
 debug-dylink = ["wasmtime-lind-dylink/debug-dylink", "wasmtime-lind-multi-process/debug-dylink", "wasmtime-lind-utils/debug-dylink", "wasmtime/debug-dylink"]
+debug-grate-calls = ["wasmtime-lind-3i/debug-grate-calls"]
 
 [dependencies]
 wasmtime-lind-common = { path = "../wasmtime/crates/lind-common" }
--- a/src/lind-boot/src/lind_wasmtime/execute.rs
+++ b/src/lind-boot/src/lind_wasmtime/execute.rs
@@ -1,4 +1,7 @@
 use crate::lind_wasmtime::host::DylinkMetadata;
+use crate::lind_wasmtime::host::{
+    cleanup_grate_handler, init_grate_pool, register_grate_handler_for_cage,
+};
 use crate::{cli::CliOptions, lind_wasmtime::host::HostCtx, lind_wasmtime::trampoline::*};
 use anyhow::{Context, Result, anyhow, bail};
 use cage::signal::{lind_signal_init, signal_may_trigger};
@@ -21,7 +24,7 @@ use wasmtime::{
     AsContextMut, Engine, Export, Func, InstantiateType, Linker, Module, Precompiled, SharedMemory,
     Store, Val, ValType, WasmBacktraceDetails,
 };
-use wasmtime_lind_3i::{VmCtxWrapper, init_vmctx_pool, rm_vmctx, set_vmctx, set_vmctx_thread};
+use wasmtime_lind_3i::*;
 use wasmtime_lind_common::LindEnviron;
 use wasmtime_lind_dylink::DynamicLoader;
 use wasmtime_lind_multi_process::{
@@ -66,13 +69,13 @@ pub fn execute_wasmtime(lindboot_cli: CliOptions) -> anyhow::Result<i32> {
     // new cage is created
     lind_manager.increment();
 
-    // Initialize vmctx pool
-    init_vmctx_pool();
+    let grate_cleanup_funcptr = cleanup_grate_handler as *const () as usize as u64;
     // Initialize trampoline entry function pointer for wasmtime runtime.
     // This is for grate calls to re-enter wasmtime runtime.
     threei::register_trampoline(
         threei_const::RUNTIME_TYPE_WASMTIME,
         grate_callback_trampoline,
+        grate_cleanup_funcptr,
     );
 
     // Register syscall handlers (clone/exec/exit) with 3i
@@ -570,7 +573,7 @@ fn load_main_module(
     // I don't setup `epoch_handler` since it seems not being used by our previous implementation.
     // Not sure if this is related to our thread exit problem
     let linker = linker_guard.clone();
-    let (instance, cage_instanceid) = linker
+    let (instance, stack_arena_base, cage_instanceid) = linker
         .instantiate_with_lind(
             &mut *store,
             &module,
@@ -653,6 +656,9 @@ fn load_main_module(
     // see comments at signal_may_trigger for more details
     signal_may_trigger(cageid);
 
+    init_grate_pool();
+    init_vmctx_pool();
+
     // The main challenge in enabling dynamic syscall interposition between grates and 3i lies in Rust’s
     // strict lifetime and ownership system, which makes retrieving the Wasmtime runtime context across
     // instance boundaries particularly difficult. To overcome this, the design employs low-level context
@@ -672,32 +678,23 @@ fn load_main_module(
     // This function will be called at either the first cage or exec-ed cages.
     set_vmctx_thread(cageid, THREAD_START_ID as u64, vmctx_wrapper);
 
+    let grate_template = GrateTemplate {
+        engine: module.engine().clone(),
+        module: module.clone(),
+        linker: linker_guard.clone(),
+    };
+
+    let host = store.data().clone();
+
+    // 4) register grate workers for this cage
+    register_grate_handler_for_cage(&grate_template, host, cageid)
+        .with_context(|| format!("failed to register grate workers for cage {}", cageid))?;
+
     // 4) Notify threei of the cage runtime type
     threei::set_cage_runtime(cageid, threei_const::RUNTIME_TYPE_WASMTIME);
 
     let mut linker = linker_guard.clone();
     linker.define_weak_imports_as_traps(&module);
-
-    // 5) Create backup instances to populate the vmctx pool
-    // See more comments in lind-3i/lib.rs
-    for _ in 0..INSTANCE_NUMBER {
-        let (instance, backup_cage_instanceid) = linker
-            .instantiate_with_lind_thread(&mut *store, &module, false)
-            .context(format!("failed to instantiate"))?;
-
-        // Extract vmctx pointer
-        let backup_cage_storeopaque = store.inner_mut();
-        let backup_cage_instancehandler = backup_cage_storeopaque.instance(backup_cage_instanceid);
-        let backup_vmctx_ptr: *mut c_void = backup_cage_instancehandler.vmctx().cast();
-
-        // Put vmctx in a Send+Sync wrapper
-        let backup_vmctx_wrapper = VmCtxWrapper {
-            vmctx: NonNull::new(backup_vmctx_ptr).ok_or_else(|| anyhow!("null vmctx"))?,
-        };
-
-        // Store the vmctx wrapper in the global table for later retrieval during grate calls
-        set_vmctx(cageid, backup_vmctx_wrapper);
-    }
 
     // must drop linker before jump into wasm
     drop(linker);
@@ -707,13 +704,6 @@ fn load_main_module(
         Some(func) => invoke_func(store, func, &args),
         None => Ok(vec![]),
     };
-
-    if !rm_vmctx(cageid) {
-        panic!(
-            "[lind-boot] Failed to remove existing VMContext for cage_id {}",
-            cageid
-        );
-    }
 
     ret
 }
@@ -832,7 +822,7 @@ pub fn precompile_module(cli: &CliOptions) -> Result<()> {
         .with_context(|| format!("failed to read {}", wasm_path.display()))?;
     let cwasm_bytes = engine
         .precompile_module(&wasm_bytes)
-        .context("failed to precompile module")?;
+        .with_context(|| format!("failed to precompile module {}", wasm_path.display()))?;
     std::fs::write(&cwasm_path, cwasm_bytes)
         .with_context(|| format!("failed to write {}", cwasm_path.display()))?;
 
@@ -852,9 +842,14 @@ fn read_wasm_or_cwasm(engine: &Engine, path: &Path) -> Result<Module> {
     // When passing in a .wasm file, the ELF parsing unwinds early. (`ElfFile64::parse(&read_cache)?;`)
     // We can therefore not call .context()? on this function since that would unwind and not run the Module::from_file()
     match engine.detect_precompiled_file(path) {
-        Ok(_) => unsafe { Module::deserialize_file(engine, path) }
-            .context("failed to deserialize precompiled module"),
-        Err(_) => Module::from_file(engine, path).context("failed to compile module"),
+        Ok(_) => unsafe { Module::deserialize_file(engine, path) }.with_context(|| {
+            format!(
+                "failed to deserialize precompiled module {}",
+                path.display()
+            )
+        }),
+        Err(_) => Module::from_file(engine, path)
+            .with_context(|| format!("failed to compile module {}", path.display())),
     }
 }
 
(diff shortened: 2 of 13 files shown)


src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -653,35 +658,23 @@ impl<
...
+                    create_handler_for_cage(
+                        &grate_template,
+                        store.data().clone(),
+                        child_cageid,
+                        ConcurrencyMode::Parallel,
+                    )
+                    .with_context(|| {
+                        format!("failed to register grate workers for cage {}", child_cageid)
+                    })
+                    .expect("create_handler_for_cage failed");
```
rennergade: does it make sense to panic here? I think it does but just bringing it up

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -635,12 +633,19 @@ impl<
...
+
+                    // Notify threei of the cage runtime type
+                    threei::set_cage_runtime(child_cageid, threei_const::RUNTIME_TYPE_WASMTIME);
+
+                    barrier_clone.wait();
+
+                    // update the linker for the child instance, since new linker contains some child-specific defines
+                    let mut new_child_host = store.data_mut();
+                    let new_child_ctx = get_cx(&mut new_child_host);
+                    let cloned_linker = linker.clone();
```
rennergade, line 645: I think you have to attach before clone or the function definitions wont be correct? I dont know exactly how this works

src/wasmtime/crates/lind-multi-process/src/lib.rs
Yaxuan-w: I'd like to panic while we're still under development phase

src/sysdefs/src/constants/lind_platform_const.rs
```diff
@@ -124,3 +124,61 @@ pub enum DylinkErrorCode {
...
+/// a single shared arena formula rather than being recomputed independently by
+/// each worker from instance-local metadata.
+///
+/// This value is placed in `sysdefs` as a shared low-level platform constant
+/// rather than being owned by `lind-3i`, `lind-multi-process`, or Wasmtime
+/// directly. The main reason is dependency hygiene: all three components need
+/// to agree on the same stack-arena base, but placing it in any one of those
+/// layers would introduce circular dependency pressure between runtime,
+/// interposition, and process-management code.
+pub static STACK_ARENA_BASE: OnceLock<u32> = OnceLock::new();
```
rennergade: does this need to be per-cage, a bit confused

rennergade (changes requested):
Left 3 comments trying to understand some of how this works

src/sysdefs/src/constants/lind_platform_const.rs
Yaxuan-w: They are same across cages

Yaxuan-w: Since the initialized memory size are same for all cages

src/wasmtime/crates/lind-multi-process/src/lib.rs
Yaxuan-w: following modification to original linker are all dynamic related, but grate calls only supports static for now, so should be fine

src/lind-boot/Cargo.toml
```diff
@@ -8,6 +8,7 @@ disable_signals = ["cage/disable_signals", "wasmtime-lind-multi-process/disable_
 secure = ["typemap/secure"]
 lind_debug = ["wasmtime-lind-common/lind_debug"]
 debug-dylink = ["wasmtime-lind-dylink/debug-dylink", "wasmtime-lind-multi-process/debug-dylink", "wasmtime-lind-utils/debug-dylink", "wasmtime/debug-dylink"]
+debug-grate-calls = ["wasmtime-lind-3i/debug-grate-calls"]
```
stupendoussuperpowers, line 11: Do we need a separate feature for just debugging? I think these can just be runtime/CLI flags so that we don't need to keep rebuilding things.

src/sysdefs/src/constants/lind_platform_const.rs
qianxichen233: this is probably not true for static build model. In static build, the stack size is retrieved from module information, which is filled by compiler. The compiler may increase the stack size if it wants (e.g. if the wasm module is slightly larger), and would probably make different wasm module has different stack size, therefore different STACK_ARENA_BASE. The dynamic build is not using memory info from module so the stack size if always fixed though

src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -43,216 +43,740 @@
...
+/// This determines whether multiple submitted grate calls may execute
+/// concurrently on different workers, or whether entry must be serialized.
+pub enum ConcurrencyMode {
+    /// Allow multiple calls to execute concurrently as long as distinct
+    /// workers are available in the pool.
+    Parallel,
+
+    /// Allow multiple calls to execute concurrently as long as distinct
+    /// workers are available in the pool.
+    Serialized,
```
drapl0n: Can you update the comment in-order to avoid confusion? 

stating something like "a single call request is addressed at a time even though requests are made concurrently" will make it easy to distinguish between both the modes.

src/sysdefs/src/constants/lind_platform_const.rs
Yaxuan-w: Thanks for pointing this out! Switching to per grate storage.

src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -43,216 +43,743 @@
...
+    ///
+    /// Each worker is an independent `store + instance + call stack` execution
+    /// context for the same grate module. Pre-initializing the full pool allows
+    /// later grate calls to lease a ready-to-run worker without paying the cost
+    /// of instantiation on the fast path.
+    ///
+    /// This worker replication is what enables grate-call concurrency: parallel
+    /// calls execute in different Wasmtime stores rather than contending on a
+    /// single shared execution context.
+    fn init_ten_workers(
```
drapl0n: function name indicates initialization of ten workers while loop iterates `MAX_GRATE_WORKERS` times i.e. 32

src/lind-boot/src/lind_wasmtime/execute.rs
```diff
@@ -66,20 +67,23 @@ pub fn execute_wasmtime(lindboot_cli: CliOptions) -> anyhow::Result<i32> {
...
     // This is for grate calls to re-enter wasmtime runtime.
     threei::register_trampoline(
         threei_const::RUNTIME_TYPE_WASMTIME,
         grate_callback_trampoline,
+        grate_cleanup_funcptr,
     );
 
     // Register syscall handlers (clone/exec/exit) with 3i
     if !register_wasmtime_syscall_entry() {
         panic!("[lind-boot] egister syscall handlers (clone/exec/exit) with 3i failed");
```
stupendoussuperpowers: missing 'r' in register.

src/wasmtime/crates/lind-3i/src/lib.rs
stupendoussuperpowers: Rename this to init_workers, since we are initing MAX_GRATE_WORKERS and not 10.

src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -43,216 +43,743 @@
...
-pub struct VmCtxWrapper {
-    pub vmctx: NonNull<c_void>,
+/// This determines whether multiple submitted grate calls may execute
+/// concurrently on different workers, or whether entry must be serialized.
+pub enum ConcurrencyMode {
+    /// Allow multiple calls to execute concurrently as long as distinct
+    /// workers are available in the pool.
+    Parallel,
+
+    /// Allow multiple calls to execute concurrently as long as distinct
```
rennergade: this seems to be copy and pasted and incorrect

rennergade (dismissed):
Commented on an small error in the doc strings but otherwise I think this is good to go

src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -43,216 +43,743 @@
...
+            println!(
+                "Worker {} handling grate request for cage {}, handler_addr: {:#x}",
+                self.worker_id, req.cageid, req.handler_addr
+            );
+        }
+
+        self.reset_worker_stack();
+
+        let func = self.pass_fptr_func.as_ref().ok_or_else(|| {
+            anyhow::anyhow!("no pass_fptr_func found in worker {}", self.worker_id)
```
stupendoussuperpowers, line 607: In this model does each worker get its own copy of the `pass_fptr_to_wt` func? Can that be shared amongst all workers of a grate?

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -653,36 +658,28 @@ impl<
...
-                        let backup_cage_instancehandler =
-                            backup_cage_storeopaque.instance(backup_cage_instanceid);
-                        let backup_vmctx_ptr: *mut c_void =
-                            backup_cage_instancehandler.vmctx().cast();
-
-                        let backup_vmctx_wrapper = VmCtxWrapper {
-                            vmctx: NonNull::new(backup_vmctx_ptr).unwrap(),
+                    // Grate calls only supports static linking for now, so we only register
+                    // grate workers when dylink is not enabled.
+                    if !dylink_enabled {
```
stupendoussuperpowers, line 663: Not sure where we'd add this, but should we have a panic/error message on grate calls when things are dynamically linked?

stupendoussuperpowers (dismissed):
Minor feedback in some places. 

Overall this looks fine to me, the changes suggested can also be taken up in a different PR after this is merged.

src/wasmtime/crates/lind-3i/src/lib.rs
Yaxuan-w: > In this model does each worker get its own copy of the pass_fptr_to_wt func? 

Yes

> Can that be shared amongst all workers of a grate?

No, since the TypedFunc is associated with vmctx which is different in the new model.

drapl0n (commented):
added minor comments. everything looks good.

src/wasmtime/crates/lind-multi-process/src/lib.rs
Yaxuan-w: At initialization stage, wasmtime doesn't know whether current cage is a grate or normal cage.. but grate call will panic on trampoline when trying to access corresponding grate workers (it will panic with things like "uninitialized workers / etc." )

drapl0n (approved):
LGTM!

stupendoussuperpowers (approved):

rennergade (approved):

PR #1102, opened 2026-04-23
Fix munmap panic when unmapping at the end of the last vmmap entry

### LINKED ISSUE #1101: munmap panic when unmapping at the end of the last vmmap entry
On main branch, when build cage files statically, the interposing mmap failed because of invalid range passed to vmmap:

```sh
lind@d6d71d38db4b:~/lind-wasm$ ./scripts/lind_run interpose-mmap_grate.cwasm interpose-mmap.cwasm
[Grate|interpose-mmap] Registering mmap handler for cage 2 in grate 1 with fn ptr addr: 3
[Grate|interpose-mmap] Handling function ptr: 3 from cage: 1
[Grate|interpose-mmap] In mmap_grate 1 handler for cage: 1
[Grate|interpose-mmap] Handling function ptr: 3 from cage: 1
[Grate|interpose-mmap] In mmap_grate 1 handler for cage: 1

thread 'lind-fork-2' (64203) panicked at /home/lind/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nodit-0.9.2/src/utils.rs:255:3:
invalid interval given to function see here for more details: https://docs.rs/nodit/latest/nodit/#invalid-intervals
```

However, the test successfully runs if compiling cage files dynamically.

Yaxuan-w:
Resolves #1101 

Fixed a boundary bug in find_unmappable_ranges() where munmap() near the end of the address space could construct an invalid Nodit interval (start >= end) and panic. The fix adds an explicit req_start >= last_end check and returns no overlaps instead of querying overlapping() with an empty/reversed interval.


rennergade:
why do we get rid of the test here?

Yaxuan-w:
> why do we get rid of the test here?

It will segfault on native. The original intent of the test was to ensure that the runtime does not interpret edge as still being inside the mapped interval. But munmap(edge, pagesz) is not a no-op at the exclusive end boundary: on native Linux, it may successfully unmap some unrelated adjacent mapping after edge. In other words, I accidentally removed a different page from the process address space, which later caused a segfault. I’m currently working on a more stable way to test this behavior.

drapl0n (approved):

rennergade (approved):

PR #1105, opened 2026-04-23
fix typo in --preload env in lind_run

rishabhBudhouliya:
# Purpose
**Correct**
`--preload env=/lib/libc.cwasm`
**Incorrect**
`--preload env=lib/libm.cwasm `                                                                                                                                                                                
 
Impact: libm.cwasm fails to load with relative path inside chroot.


Yaxuan-w (approved):

vidyalakshmir (approved):
Looks good. I believe having or not having '/' at the beginning, both works. But it is good to keep it uniform.

PR #1112, opened 2026-04-24
Change optimization level from O0 to O2

drapl0n:


drapl0n:
@rennergade `file_tests/deterministic/statfs.c` is still failing.

rennergade:
> @rennergade `file_tests/deterministic/statfs.c` is still failing.

yeah looking at that, I don't really understand whats getting optimized out here.

rennergade:
@Yaxuan-w @qianxichen233 I skip listed statfs, seems to be an obscure error and not really a breaking change. Put up an issue for someone else to look into but not urgent. Think this is good to merge.

rennergade (approved):

Yaxuan-w (approved):

PR #1114, opened 2026-04-25
fix lind fd limit

qianxichen233:
lind host process fd limit is by default 1024 (note this is lind host process system fd limit, not per-cage fd limit). This causes all the running cage can only have a sum of 1024 file descriptors together.


rennergade:
Looks good. Can we put this in its own init function

PR #1120, opened 2026-04-26
Remove set_stack_pointer helper object from glibc/sysroot build

### LINKED ISSUE #978: glibc: depreciate set_stack_pointer.o
Very early on, we are using a wasm object file to modify `__stack_pointer` of the wasm module since we do not find a good way for host to manipulate the `__stack_pointer`, so we added a guest trampoline to do this for us.

But now I think it should be much more easier for us to do it in host. We can depreciate this file and have host to directly manipulate `__stack_pointer` using wasmtime API

This is not an urgent issue and is a pretty good task for beginners who want to starts to get touch with wasmtime codebase

DanielZ224:
## Purpose

This PR removes the deprecated `set_stack_pointer.o` helper from the glibc/sysroot build path.

Previously, `set_stack_pointer.o` was built from `set_stack_pointer.s`, copied into the wasm32-wasi sysroot, included in glibc object lists, and explicitly linked during dynamic compilation. The stack pointer update is now handled on the Wasmtime side, so this standalone helper object is no longer needed.

This change:
- removes `src/glibc/csu/wasm32/set_stack_pointer.s`
- removes `set_stack_pointer.o` / `set_stack_pointer.os` from glibc object lists
- stops building and copying `set_stack_pointer.o` into the sysroot
- stops linking `set_stack_pointer.o` in `scripts/lind_compile`
- removes the dummy `set_stack_pointer` reference from `pthread_create.c`
- updates the toolchain documentation accordingly

## Related Issue

Closes #978  
Related to: `glibc: depreciate set_stack_pointer.o`

## Validation

I checked that there are no remaining script/source/document references to:

- `set_stack_pointer.s`
- `set_stack_pointer.o`
- `set_stack_pointer.os`

using `grep` across `scripts`, `src`, and `docs`, excluding generated/binary artifacts.

I also ran:

```bash
cd ~/lind-wasm/src/wasmtime
cargo check -p wasmtime --lib

### CODE REVIEW

File: src/wasmtime/crates/wasmtime/src/runtime/instance.rs [hunk 1]
```diff
@@ -795,6 +795,23 @@ impl Instance {
         return Err(());
     }
 
+    pub fn set_stack_pointer(&self, mut store: impl AsContextMut, sp: i32) -> Result<(), ()> {
```
Yaxuan-w, line 801: Can we also have comments for this function?

DanielZ224: Thanks! I added comments for this function to explain that it directly updates the exported `__stack_pointer` global from the Wasmtime side and replaces the previous `set_stack_pointer.o` helper used during pthread creation.

Yaxuan-w (dismissed):

src/wasmtime/crates/wasmtime/src/runtime/instance.rs
```diff
@@ -916,6 +795,33 @@ impl Instance {
...
     }
 
+    /// Updates the exported `__stack_pointer` global for this instance.
+    ///
+    /// Lind uses this during pthread creation to initialize the child thread's
+    /// stack pointer directly from the Wasmtime side. Previously this was done
+    /// through a small wasm helper object (`set_stack_pointer.o`) linked into
+    /// glibc. After removing that helper object, the runtime updates the
+    /// mutable `__stack_pointer` global directly instead.
+    ///
```
qianxichen233: this part of comments are not necessary

qianxichen233:
should have a parallel PR on lind-wasm-apps side to remove all the reference to `set_stack_pointer.o` as well. e.g. https://github.com/Lind-Project/lind-wasm-apps/blob/97b7e6f7185f5e85bb39dff07f84d8d6e7069bad/cpython/compile_cpython.sh#L297

Look into `compile_*.sh` under each application in lind-wasm-apps to search for the reference

Both PR needs to be merged together otherwise it will break builds on app side

src/wasmtime/crates/wasmtime/src/runtime/instance.rs
DanielZ224: Comments deleted.

qianxichen233 (approved):

PR #1121, opened 2026-04-26
handle error in epoll_wait_syscall()

drapl0n:
- replaces unsafe `unwrap()` with error handeling based on Result<>
- returns appropriate error code

This PR resolves the issue causing failure of `test_epoll` test in `cpython` app.


rennergade:
@drapl0n add a test here please

drapl0n:
> @drapl0n add a test here please

Failed test:  [test_fromfd()](https://github.com/Lind-Project/lind-wasm-apps/blob/main/cpython/Lib/test/test_epoll.py#L142). It expects EBADF as the return value since fd was closed when `ep` exited.

PR #1124, opened 2026-04-27
create /dev/urandom and /dev/random and add ./resource/Versions

rishabhBudhouliya:
# Purpose
During building PostgreSQL, we found that `/dev/random` was missing from lindfs to generate secret authorization token
and env::getrusage has not been defined.

To fix that:

1. Makefile: Copies /dev/urandom and /dev/random during make lindfs                                                                                 
2. version-path-minimal.txt: Adds ./resource/Versions for getrusage export

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,10 @@ lindfs:
 		mkdir -p $(LINDFS_ROOT)/$$d; \
 	done
 	touch $(LINDFS_ROOT)/dev/null
+	@if [ -c /dev/urandom ]; then \
+		cp -a /dev/urandom $(LINDFS_ROOT)/dev/; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -46,6 +46,10 @@ lindfs:
 		mkdir -p $(LINDFS_ROOT)/$$d; \
 	done
 	touch $(LINDFS_ROOT)/dev/null
+	@if [ -c /dev/urandom ]; then \
+		cp -a /dev/urandom $(LINDFS_ROOT)/dev/; \
```
rennergade, line 50: I'm not really sure if it makes sense to do it like this long term though its probably fine as a hack to get things working for testing.

Thoughts @qianxichen233 @JustinCappos

PR #1125, opened 2026-04-27
Change cage epoch to atomicptr

### LINKED ISSUE #1123: Replace raw pointer in Cage::epoch_handler with AtomicPtr
While fixing #1050, I tried replacing `CAGE_MAP` with a thread-safe structure, e.g.:

```rust
Lazy<Vec<RwLock<Option<Arc<Cage>>>>>
```

This exposed a separate thread-safety issue in `Cage`:

```rust
pub epoch_handler: DashMap<i32, RwLock<*mut u64>>,
```

Since the new `CAGE_MAP` stores `Arc<Cage>` in a shared static, Cage must be `Send` + `Sync`. However, raw pointers like `*mut u64` are not `Send` or `Sync` by default, so the compiler rejects the change with:

```sh
error[E0277]: `*mut u64` cannot be shared between threads safely
```

My current solution is to replace the raw pointer slot with `AtomicPtr<u64>`

```rust
pub epoch_handler: DashMap<i32, AtomicPtr<u64>>,
```

Yaxuan-w:
Resolves #1123 

1. Change cage epoch to atomicptr
2. Replace debug_assertions with lind_debug
3. Remove unnecessary cage helper function

### DIFF
--- a/src/cage/src/signal/signal.rs
+++ b/src/cage/src/signal/signal.rs
@@ -1,8 +1,7 @@
-use crate::cage::{get_cage, get_cage_or_debug_panic};
-use parking_lot::RwLock;
-use std::sync::atomic::Ordering;
+use crate::cage::get_cage;
+use std::sync::atomic::{AtomicPtr, Ordering};
 use sysdefs::constants::{SA_NODEFER, SA_RESETHAND, SIG_DFL};
-#[cfg(debug_assertions)]
+#[cfg(feature = "lind_debug")]
 use sysdefs::logging::lind_debug_panic;
 
 const EPOCH_NORMAL: u64 = 0;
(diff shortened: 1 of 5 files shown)


rennergade (approved):

src/cage/src/signal/signal.rs
```diff
@@ -17,31 +16,49 @@ pub fn signal_epoch_trigger(cageid: u64) {
 
     #[cfg(not(feature = "disable_signals"))]
     {
-        let Some(cage) = get_cage_or_debug_panic(cageid, "signal_epoch_trigger") else {
-            return;
+        let cage = match get_cage(cageid) {
+            Some(c) => c,
+            None => {
+                #[cfg(feature = "lind_debug")]
```
qianxichen233, line 22: why wrap this under feature? if the cage is not found, that probably means there is some fatal bugs in our system which caused some corrupted internal state. This should get noticed immediately rather than ignoring it

Yaxuan-w: This part came from @rennergade’s earlier thread-exiting changes. I’m not fully sure whether some of the current thread-exiting paths are supposed to silently return on error, so @rennergade may be the best person to clarify the intended behavior here. I’m happy to adjust this once we confirm the expected semantics.

rennergade: yes this was my change. I think removing this may cause issues. If we want to look into that I suggest we open a new issue and have someone look into it separately. Basically there are times where this will return None when cages are racing on teardown.

rennergade (approved):
Commented on the thread there but this looks good to me. Approved.

PR #1133, opened 2026-04-29
Propagate interposed clone return value across Asyncify rewind

### LINKED ISSUE #1132: Can't change return PID when interposing on fork.
Suppose we have a handler for fork that does the following:

1. Fork an actual child through rawposix & wasmtime.
2. Instead of returning the actual PID of the new child cage, return a different value. 

Something like this:
```c
int fork_grate(uint64_t cageid,
    uint64_t arg1, uint64_t arg1cage,
    uint64_t arg2, uint64_t arg2cage,
    uint64_t arg3, uint64_t arg3cage,
    uint64_t arg4, uint64_t arg4cage,
    uint64_t arg5, uint64_t arg5cage,
    uint64_t arg6, uint64_t arg6cage) {
  printf("[Grate|interpose-fork] In fork_grate %d handler for cage: %llu\n",
         getpid(), cageid);

  int self_grate_id = getpid();

  int new_cageid = make_threei_call(
    56, // syscallnum for clone
    0,    // callname is not used in the trampoline, set to 0
    self_grate_id,    // self_grate_id is not used in the trampoline, set to 0
    arg1cage,    // target_cageid is not used in the trampoline, set to 0
    arg1, arg1cage,
    arg2, arg2cage,
    arg3, arg3cage,
    arg4, arg4cage,
    arg5, arg5cage,
    arg6, arg6cage,
    0 // we will handle the errno in this grate instead of translating it to -1 in the trampoline
  );

  return 10;
}
```

The expectation would be when the target cage does:

```c
cpid = fork();
printf("CPID=%d\n", cpid);
```

The output would be `10` in the target cage. 

However that is not what happens, because the return value of `fork` inside the wasm modules is not routed through any of the grate handlers once we call wasmtime's `fork_call` function. The return value from the grate handler is only useful when a grate above above it calls `make_threei_call(...)` and never within the wasm module of the actual cage that called fork. 

The return value that fork sees in both parent and child are handled through Asyncify: https://github.com/Lind-Project/lind-wasm/blob/44ca48332f849cf9cf47aef5473031df82910333/src/wasmtime/crates/lind-multi-process/src/lib.rs#L282

TL;DR this is the control flow of a process calling fork when it's interposed. 

Image

1. Cage calls fork() and it goes to the grate handler.

2. Grate handler calls make_threei_call which goes to wasmtime.

3(A). Wasmtime returns CPID to the Grate handler. 

3(B). Wasmtime ALSO directly sets the child cage id within the Cage through asyncify. 

4. The PID returned by the grate handler is ignored.

Yaxuan-w:
Resolves #1132 

This PR preserves a grate-defined return value for interposed `clone`/`fork` calls across the Asyncify unwind/rewind boundary.

Previously, when a grate interposed on `clone` and returned a custom value, the parent cage still observed the real child cage id. This happened because the guest-visible `fork()` return value is produced during the later Asyncify rewind replay, where `catch_rewind()` returns the saved rewind value directly and bypasses the normal syscall/interposition return path.

This PR adds a pending visible clone return slot to the Lind context and consumes it during the parent-side clone rewind

### DIFF
--- a/src/lind-boot/src/lind_wasmtime/host.rs
+++ b/src/lind-boot/src/lind_wasmtime/host.rs
@@ -37,6 +37,10 @@ impl LindHost<HostCtx, CliOptions> for HostCtx {
     fn get_ctx(&self) -> LindCtx<HostCtx, CliOptions> {
         self.lind_fork_ctx.clone().unwrap()
     }
+
+    fn get_ctx_mut(&mut self) -> &mut LindCtx<HostCtx, CliOptions> {
+        self.lind_fork_ctx.as_mut().unwrap()
+    }
 }
 
 pub struct DylinkMetadata {
--- a/src/wasmtime/crates/lind-common/src/lib.rs
+++ b/src/wasmtime/crates/lind-common/src/lib.rs
@@ -129,6 +129,20 @@ fn add_syscall_to_linker<
             // check here to early-return when we are on a rewind replay path.
             if call_number as i32 == CLONE_SYSCALL {
                 if let Some(rewind_res) = wasmtime_lind_multi_process::catch_rewind(&mut caller) {
+                    if rewind_res > 0 {
+                        // On Asyncify rewind replay, `clone` must not be executed again.
+                        // The positive rewind result is the real child cage id returned to the
+                        // parent guest by default. If a grate supplied a pending visible return
+                        // value during the first normal execution, consume it here and return it
+                        // instead. The child-side rewind result is 0 and must not be overridden.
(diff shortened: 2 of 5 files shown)


rennergade (dismissed):
this looks good to me

stupendoussuperpowers (commented):
The changes look good to me, can we add a grate unit test for this? It can just be the same as the interpose-fork_grate but it returns some constant value like 10.

stupendoussuperpowers (approved):

stupendoussuperpowers:
Approved, I think we should merge this once the CI tests finish.

PR #1119, opened 2026-04-26
Fix debug build

rennergade:
Closes #1118 

Needed to add lind_debug_printf to shared script.

PR #1136, opened 2026-04-29
Restore lind debug import helper for debug builds

DanielZ224:
## Purpose

This PR addresses the requested change on #1119 regarding the Lind debug build regression.

A previous cleanup accidentally removed the debug import-preserving helper logic from `lind_debug.c`, including `__lind_debug_num` and `__lind_debug_import`. A later change also replaced the debug-mode startup call in `crt1.c` from `__lind_debug_import()` to `lind_debug_printf(...)`, which was not the expected behavior.

This PR restores the original import-preserving path while keeping `lind_debug_printf()` available as a formatted debug-print helper.

## Related PR / Review Comment

Addresses the requested change on #1119.

Related commits discussed in the review:
- `0f386b3`: removed the debug import helper path from `lind_debug.c`
- `4265012`: replaced the `crt1.c` debug startup call with `lind_debug_printf(...)`

## Changes

- Restored the raw debug imports under `LIND_DEBUG`:
  - `__lind_debug_num`
  - `__lind_debug_str`
- Restored `__lind_debug_import()` in `lind_debug.c` to force the debug imports to be kept.
- Added the corresponding declarations back to `lind_debug.h`.
- Changed the `crt1.c` debug-mode startup call back to `__lind_debug_import()`.
- Kept `lind_debug_printf()` available for formatted debug output.

## Testing

I ran the relevant local build/check command and confirmed that it completed without errors.

## Notes for Reviewers

This PR is intended to minimally restore the missing debug import-preserving behavior requested in #1119, without removing the newer `lind_debug_printf()` helper.

### DIFF
--- a/scripts/lind_compile
+++ b/scripts/lind_compile
@@ -405,7 +405,7 @@ do_dynamic_compile() {
   fi
 
   run_cmd "${CLANG_BIN}" "${default_clang_flags[@]}" "${mandatory_clang_flags[@]}"\
-    "${EXTRA_CLANG_ARGS[@]}" "${SRC}" "$SYSROOT/lib/wasm32-wasi/crt1_shared.o" "$SYSROOT/lib/wasm32-wasi/lind_utils.o" -o "${OUT_WASM}"
+    "${EXTRA_CLANG_ARGS[@]}" "${SRC}" "$SYSROOT/lib/wasm32-wasi/crt1_shared.o" "$SYSROOT/lib/wasm32-wasi/lind_utils.o" "$SYSROOT/lib/wasm32-wasi/lind_debug.o" -o "${OUT_WASM}"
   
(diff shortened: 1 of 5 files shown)


qianxichen233:
Could you try rebuild lind with debug mode (use `make lind-debug`) and run any testcase and see if it works

DanielZ224:
@qianxichen233 
Thanks for the suggestion. I rebuilt Lind with debug mode using `make lind-debug` and tested a unit testcase with the normal compile/run flow:

```bash
lind-clang tests/unit-tests/file_tests/deterministic/statfs.c
lind-wasm tests/unit-tests/file_tests/deterministic/statfs.cwasm
```

Initially, the testcase failed with:

```bash
unknown import: `env::__lind_debug_import` has not been defined
```

I found that lind_debug.o was built, but it was not installed into the sysroot or linked by lind_compile in the dynamic build path. I updated the build scripts so lind_debug.o is copied into the sysroot and linked automatically.

After rebuilding with make lind-debug, the same testcase now runs successfully and prints:

```bash
[LIND DEBUG NUM]: 0
[LIND DEBUG STR]: LIND DEBUG INIT
```

qianxichen233 (approved):

PR #1137, opened 2026-04-29
Add rdynamic support

### LINKED ISSUE #1135: dynamic loading: exporting main module symbols
Scenarios that library function reference back to main module's symbol require main module exporting its symbols as well

This scenario is observed in `plpgsql.so` in postgres

Current main module compilation pipeline does not generate the required export symbols:
```
    --target=wasm32-unknown-wasi
    "--sysroot=${SYSROOT}"
    "-fvisibility=default"
    -Wl,-pie
    -Wl,--import-table
    -Wl,--import-memory
    -Wl,--allow-undefined
    -Wl,--unresolved-symbols=import-dynamic
```

In library compilation, the exported symbol seems to be generated by `-shared` option from wasm-ld. And this is conflicting with `-pie` option in current main module compilation pipeline: `wasm-ld: error: -shared and -pie may not be used together`

Removing `-pie` would make main module failed to link crt1 and could have other potential side effects.

We need to find a way to make main module exports its public symbols as well without breaking its dynamic loading semantics

qianxichen233:
fix https://github.com/Lind-Project/lind-wasm/issues/1135

add rdynamic support to allow library reference symbols back to main module

Added an additional optional lind_compile argument e.g. `lind-clang -rdynamic main.c`


rennergade (dismissed):
This looks good to me.

scripts/lind_compile
```diff
@@ -50,6 +50,9 @@ usage: $(basename "$0") [OPTIONS] <source> [-- <clang args>]
     --fpcast-emu                enable function-pointer-cast emulation via wasm-opt
                                 (applies to full and --opt-only modes; for dynamic/library
                                 builds also passes --pass-arg=relocatable-fpcast)
+    -rdynamic                   export all symbols from the main module so dynamically loaded
```
Yaxuan-w, line 53: Can we use `--rdynamic` here to align with other flags?

Yaxuan-w (changes requested):

scripts/lind_compile
qianxichen233: this aligns with standard clang option which uses `-rdynamic` rather than `--rdynamic`

rennergade:
Needs to fix merge conflict, why are the tests skipped here btw?

qianxichen233:
> Needs to fix merge conflict, why are the tests skipped here btw?

since our CI still cannot support dlopen related tests yet: https://github.com/Lind-Project/lind-wasm/issues/1004

rennergade:
ok can you resolve the conflict

PR #1138, opened 2026-04-29
Fix statfs guest buffer handling under -O2

### LINKED ISSUE #1122: statfs returns EFAULT at -O2 optimization level
`statfs()` returns -1 with errno=14 (EFAULT) when the test binary is compiled at -O2, but works at -O0.

The buffer pointer is valid (`0x800f90`, normal stack range, well below 2GB). The glibc wrapper translates via `TRANSLATE_GUEST_POINTER_TO_HOST` before the syscall. The RawPOSIX handler at `fs_calls.rs:1779` casts `statbuf_arg` directly to `*mut libc::statfs` and passes to the host `libc::statfs`.

Possible causes:
- Host `struct statfs` (x86-64) is larger than wasm32 `struct statfs`. At -O2 the compiler packs the stack tighter, and the host write overflows into adjacent stack data
- The translated pointer value changes at -O2 due to inlining or register allocation in the glibc wrapper

Test: `tests/unit-tests/file_tests/deterministic/statfs.c`
Currently skipped in `skip_test_cases.txt`.

Discovered in PR #1112 (change optimization to -O2).

DanielZ224:
## Purpose

This PR fixes the `statfs` syscall path so it correctly writes filesystem metadata back into guest memory under `-O2` optimization.

Previously, `statfs_syscall` converted the path argument correctly, but directly cast the guest output buffer pointer to a host `libc::statfs *`. This could cause the host `libc::statfs` call to write to an invalid host address and return `EFAULT`.

The updated implementation now follows the safer pattern already used by `fstatfs_syscall`: it calls host `libc::statfs` into a local host-side buffer, translates the guest output pointer with `sc_convert_addr_to_fstatdata`, and then copies the result back into guest memory with `convert_fstatdata_to_user`.

## Related Issue

Fixes #1122

## Changes

- Updated `statfs_syscall` to avoid directly casting the guest statfs buffer pointer to a host pointer.
- Added guest-memory translation for the statfs output buffer.
- Reused the existing `FSData` conversion path used by `fstatfs_syscall`.
- Renamed `statbuf_cageid` so it can be used for guest buffer address conversion.

## Testing

Ran:

```bash
cd ~/lind-wasm/src/wasmtime
cargo check -p rawposix --lib
```

## Result

- `rawposix` builds successfully.
- Existing warnings remain, but there are no new compile errors from this change.


rennergade:
@DanielZ224 O2 branch is merged please remove statfs test from the skip list so it can run CI

DanielZ224:
@rennergade 
Done. I removed `file_tests/deterministic/statfs.c` from `skip_test_cases.txt`, so the statfs test should now run in CI after the O2 branch merge.

qianxichen233:
there are conflicts you need to resolve. merge with main and resolve the conflicts to enable CI run

DanielZ224:
Done. I merged the latest `origin/main` into `fix-statfs-efault-o2`, resolved the conflicts, and pushed the updated branch.

I kept the new `rdynamic_*` skip entries from `main`, removed `file_tests/deterministic/statfs.c` from `skip_test_cases.txt`, and preserved the new `cage_map_bench` bench config from `main`. CI should be able to run now.

DanielZ224:
Since `error_handling_net.c` seems unrelated to the statfs fix and passes locally, I’ll keep this PR focused on `statfs` and open a separate issue/PR to track and fix the networking test failure.

Yaxuan-w (approved):

PR #1140, opened 2026-04-30
Increase MAX_CAGEID to 2048 and fix alloc off-by-one

rennergade:
Raise the cage ID limit from 1024 to 2048. Fix off-by-one in alloc_cage_id where the old condition (v <= MAX_CAGEID) could return a cage ID equal to MAX_CAGEID, which is out of bounds for CAGE_MAP. Add fork_max_cages test that exercises the full cage ID space.


rennergade:
Skip listed the test since itll timeout on CI, runs locally

Yaxuan-w (approved):

PR #1147, opened 2026-05-03
Add wasm-opt wrapper script

qianxichen233:
## Summary

Introduces `scripts/lind-wasm-opt`, a wrapper around `wasm-opt` that encapsulates lind's standard instrumentation flags, and migrates all direct `wasm-opt` invocations in the build scripts to use it.

**`lind-wasm-opt`** selects the correct flag set based on a required build target:
- `--target=main` — dynamic main-module binary (epoch-import + epoch-main-module + asyncify-import-globals)
- `--target=library` — dynamic shared library (epoch-import + asyncify-import-globals, no epoch-main-module)
- `--static` — static binary (bare epoch + asyncify, with -O2)

Optional flags: `--fpcast-emu` (adds `--pass-arg=relocatable-fpcast` automatically for dynamic targets), `--without-asyncify`, `--without-epoch`, `--verbose`.

The wrapper enforces the correct wasm-opt argument ordering (in particular, `-O2` after `--asyncify`) and rejects unknown options, making it harder to accidentally produce a mis-instrumented binary.

Also includes a rebuilt `libbinaryen.so` with two fixes to `FuncCastEmulation`:
- **Crash fix**: `makeThunk` now returns `nullptr` for functions exceeding `max-func-params` (default 18) instead of emitting out-of-bounds `local.get` indices, which caused heap corruption in `SimplifyLocals` on large modules like postgres
- **Whitelist support**: new `--pass-arg=fpcast-whitelist@name1,glob*` argument limits thunk generation to a named subset of functions

Related changes with `wasm-opt` are at https://github.com/Lind-Project/lind-wasm-binaryen/commit/f13fa3298ce3f26acd707299fe07e5f4f4f53a79 and https://github.com/Lind-Project/lind-wasm-binaryen/commit/f85da490f35412c5974eb6b38171a124d6d02b51

### DIFF
--- a/scripts/lind-wasm-opt
+++ b/scripts/lind-wasm-opt
@@ -1,0 +1,176 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+# lind-wasm-opt — wasm-opt wrapper that applies lind's standard instrumentation
+# flags for a given compilation target.
+#
+# Usage:
+#   lind-wasm-opt (--target=<target> | --static) [OPTIONS] input.wasm -o output.wasm
+#
+# Exactly one of --target or --static must be given:
+#   --target=main      Dynamic main-module binary.
+#   --target=library   Dynamic shared-library binary.
+#   --static           Static (non-dynamic) binary.
+#
+# Options:
+#   --fpcast-emu          Enable function-pointer-cast emulation.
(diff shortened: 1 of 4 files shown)


rennergade (approved):
Looks good on review. Im fine to fast track this for now.

Yaxuan-w (approved):

PR #1149, opened 2026-05-03
fix rdynamic across fork

qianxichen233:
## Summary

This PR fixes postgres support under lind-wasm by addressing several issues in the dynamic linking (dlopen/GOT), process fork, and thread creation paths.

### GOT symbol cache for rdynamic / late-dlopen (`lind-utils`, `instance.rs`, `linker.rs`)

Adds a `symbol_cache` field to `LindGOT`. Previously, GOT resolution only worked when the consumer module (the one with the GOT cell) loaded before or at the same time as the provider. If a provider module was loaded first (rdynamic or preloaded library), the consumer's GOT cell would never be filled.

`symbol_cache` records every resolved symbol value unconditionally at load time via `cache_symbol`. When a consumer later registers a GOT cell via `new_entry`, the cached value is written in immediately. Both directions are now handled:
- Consumer before provider: `update_entry_if_unresolved` fills the cell when the provider loads
- Provider before consumer: `new_entry` pre-fills the cell from the cache when the consumer links

`clone_with_cache` carries the cache into forked processes so child cages inherit parent symbol resolutions.

GOT.mem globals that are not i32 now emit a warning instead of silently skipping, to catch violated ABI assumptions.

### Fork correctly propagates GOT into child process

Child processes now receive a `clone_with_cache` copy of the parent's GOT rather than starting with an empty `LindGOT::new()`. This ensures dlopen'd libraries whose symbols are already resolved in the parent are immediately available in the child without re-resolving.

### `signal_callback` export registered after fork and thread creation

After forking a process or creating a thread, `instance_dylink` is called to register `signal_callback` into the child's linker, matching what the parent's initial setup does.

### Thread fork path unified with process fork (`lind-multi-process`, `host.rs`, `execute.rs`)

- Removed the hand-written `Clone` impl on `LindCtx` (which silently called `fork_thread`) and replaced it with `#[derive(Clone)]` plus an explicit `fork_thread` method parallel to `fork_process`.
- `fork_host` closure now takes a `bool` (`is_thread`) so the same closure handles both cases: `fork_call` passes `false`, `pthread_create_call` passes `true`. Previously thread creation used `caller.data().clone()` which bypassed the host fork logic entirely.
- `HostCtx::fork_thread` is now explicitly wired up. `lind_environ` is wrapped in `Arc` so threads share the same environ object (clone of the `Arc`, not a copy of the data), while process fork still gets an independent copy via `Arc::new(e.fork())`.


src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -1801,6 +1820,13 @@ impl<
 
     // fork the state for new thread
     pub fn fork_thread(&self) -> Self {
+        let cloned_got = if let Some(got) = self.got_table.as_ref() {
```
rennergade, line 1823: is this ever used?

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use sysdefs::constants::lind_platform_const::{
     unset_stack_arena_base, UNUSED_ARG, UNUSED_ID, UNUSED_NAME,
 };
 use sysdefs::constants::syscall_const::{EXEC_SYSCALL, EXIT_SYSCALL, FORK_SYSCALL};
-use sysdefs::constants::{Errno, MAX_SHEBANG_DEPTH, MMAP_SYSCALL};
+use sysdefs::constants::{Errno, DEFAULT_STACKSIZE, GUARD_SIZE, MAX_SHEBANG_DEPTH, MMAP_SYSCALL};
```
rennergade: unused?

src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -1779,9 +1791,16 @@ impl<
...
+            Some(Arc::new(Mutex::new(got_guard.clone_with_cache())))
+        } else {
+            None
+        };
+
         let forked_ctx = Self {
-            linker: None,    // Linker is explicitly set up by the caller
-            got_table: None, // new process should use a new GOT
+            linker: None,          // Linker is explicitly set up by the caller
+            got_table: cloned_got, // new process should use a new GOT
```
rennergade: stale comment?

src/wasmtime/crates/wasmtime/src/runtime/linker.rs
```diff
@@ -1520,24 +1520,27 @@ impl<T> Linker<T> {
...
                         #[cfg(feature = "debug-dylink")]
                         println!("[debug] update GOT.mem.{} to {}", name, val);
                     }
+                    // Only expose as dlsym-resolvable if it's a known GOT data symbol.
+                    if got.has_entry(&name) {
+                        symbol_map.add(name.clone(), val);
+                    }
                 }
 
+                got.warning_undefined();
```
rennergade, line 1542: should this be a debug option?

tests/unit-tests/dylink_tests/deterministic/rdynamic_main.c
```diff
@@ -25,6 +27,8 @@ int main(void) {
         return 1;
     }
 
+    fork();
```
rennergade, line 30: does this actually tst the bug? i think we need to waitpid

rennergade (changes requested):
a few comments

also  lines around src/wasmtime/crates/wasmtime/src/runtime/instance.rs:1259 and linker.rs:1523 could possibly factored into one function

tests/unit-tests/dylink_tests/deterministic/rdynamic_main.c
qianxichen233: it tested the scenario, the library function is called within child. Without the fix, the library function in child would crash

src/wasmtime/crates/wasmtime/src/runtime/linker.rs
qianxichen233: I’m not sure we should remove these warnings entirely. An unresolved GOT entry is a strong signal that something could go wrong. The warning is triggered specifically when a library references a GOT entry that remains unresolved, which can lead to unexpected failures. These errors are often hard to diagnose from the surface since they may appear random, and people might not immediately connect them to an unresolved GOT entry.

PR #1157, opened 2026-05-04
dynamic loading: library dependency resolver

qianxichen233:
use dylink.0 Needed Section to automatically resolve and load library dependencies, so manually specify `--preload` is no longer needed


src/lind-boot/src/lind_wasmtime/execute.rs
```diff
@@ -765,94 +849,88 @@ fn load_library_module(
...
-    // Extract dylink metadata from the library.
-    // This includes table and memory requirements declared by the toolchain.
-    let dylink_info = match lib_module.dylink_meminfo() {
-        Some(info) => info,
-        None => return -(DylinkErrorCode::EDYLINKINFO as i32), // dylink section is not found
-    };
+    // Resolve all transitive dependencies and link them in dependency-first order
+    // before instantiating the requested library.
+    {
+        let environ_arc = main_module.data().lind_environ.clone().unwrap();
```
vidyalakshmir, line 860: Is this providing the path where the libraries are searched? Also, is there a way for the user to specify the path where library resides? (Like LD_LIBRARY_PATH?).

Yaxuan-w:
1. Need to resolve merge conflicts. 
2. Does this pass app tests too?

PR #1162, opened 2026-05-05
fdtables: bump highestneverusedfd in get_specific_virtual_fd

rennergade:
`get_unused_virtual_fd` (muthashmax fast path) hands out the next fd at `highestneverusedfd` and increments it. `get_specific_virtual_fd` inserted at the requested fd but never updated the counter, so a caller that reserves stdio (e.g. fds 0/1/2) followed by an `get_unused_virtual_fd` would silently overwrite those reserved slots with the new entry — observed in the IPC grate's preexec where stdio got hijacked by the first user pipe().

Other backends (`vanillaglobal`, `dashmaparrayglobal`, `dashmapvecglobal`) linear-scan for a vacant slot and don't have this issue.

### DIFF
--- a/src/fdtables/src/lib.rs
+++ b/src/fdtables/src/lib.rs
@@ -741,6 +741,51 @@ mod tests {
     }
 
     #[test]
+    // Regression: after reserving fds via get_specific_virtual_fd, a
+    // subsequent get_unused_virtual_fd must not hand back one of the
+    // reserved fds.
+    //
+    // In the muthashmax backend, get_specific_virtual_fd previously
+    // didn't bump `highestneverusedfd`, so the get_unused_virtual_fd
+    // fast path would re-issue fd 0, 1, 2... and silently clobber the
+    // reservations. This surfaced in the IPC grate's preexec reserving
+    // stdio: the user binary's first pipe() call would overwrite stdin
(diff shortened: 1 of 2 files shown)


Yaxuan-w (commented):
Can we also update other algo too?

rennergade:
> Can we also update other algo too?

this bug only exists in this one because of how its setup

  - dashmapvecglobal.rs — scans 0..FD_PER_PROCESS_MAX for the first is_none() slot (line 139). No counter, can't reuse occupied.                                                       
  - dashmaparrayglobal.rs — same scan pattern (line 137).                                                                                                                              
  - vanillaglobal.rs — uses Entry::Vacant to find an unoccupied slot (line 147). Same idea.

rennergade:
> > Can we also update other algo too?
> 
the CI failures have nothing to do with this fix, apparently some small subset of GHA runners have this option that we cant run with:

"1: compilation setting "has_avx512bitalg" is enabled, but not available on the host"

stupendoussuperpowers (dismissed):

stupendoussuperpowers:
Can we add a unit test for cases like these so that the drift between different implementation is better tracked?

stupendoussuperpowers (approved):

PR #1164, opened 2026-05-05
Expose fdtables backend selector via FDTABLES_IMPL

rennergade:
Adds passthrough Cargo features in rawposix and lind-boot so the fdtables impl picked by the fdtables crate's `default = ["dashmaparray"]` can be overridden from a single Makefile knob:

    make build                              # uses dashmaparray (default)
    make build FDTABLES_IMPL=muthashmax     # switches to muthashmax
    make lind-debug FDTABLES_IMPL=vanilla   # also propagates

Mechanically: every workspace crate that depends on fdtables now does so with `default-features = false`, otherwise Cargo's feature unification re-enables `dashmaparray` alongside whatever override the user picked, tripping the fdtables compile_error guard. lind-boot's own `default` feature picks `dashmaparray` so plain `cargo build` still works.

Verified all four impls compile via the override path.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,9 +222,13 @@ lint:
 	    src/fdtables/src/dashmapvecglobal.rs \
 	    src/fdtables/src/muthashmaxglobal.rs \
 	    src/fdtables/src/vanillaglobal.rs
+	# Note: --all-features can't be used here because it enables every
+	# fdtables-* impl simultaneously, which trips the fdtables impl-mutex
+	# compile_error guard. Enumerate the non-fdtables features explicitly
+	# and pin to the default fdtables impl.
 	cargo clippy \
 	    --manifest-path src/lind-boot/Cargo.toml \
-	    --all-features \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -40,10 +40,16 @@ sysroot: build-dir
 	./scripts/make_glibc_and_sysroot.sh $(if $(WITH_FPCAST),--with-fpcast)
 	$(MAKE) sync-sysroot
 
+# fdtables backend selector. One of: dashmaparray (default), dashmapvec,
+# muthashmax, vanilla. Threaded through lind-boot -> rawposix -> fdtables
+# via Cargo features.
```
Yaxuan-w, line 45: Could we have an example on usage

Yaxuan-w (dismissed):

stupendoussuperpowers:
Wouldn't be also need to update fdtables to have these features?

stupendoussuperpowers:
Okay that is done via a different PR already, makes sense

PR #1167, opened 2026-05-06
lind-boot: disable AVX-512 in wasmtime config for portable cwasm

### LINKED ISSUE #1163: Prevent from using GHA runners with avx512bitalg option
Every once in a while the full CI fails with:

```
1: compilation setting "has_avx512bitalg" is enabled, but not available on the host
```

which is not compatible with wasm.

Need to disable these runners in GHA

rennergade:
Cranelift's default is to auto-detect the host CPU's ISA features and bake them into the precompiled artifact.  This makes .cwasm files non-portable: a build runner with has_avx512bitalg produces a .cwasm that fails to load on a runner without it ("compilation setting 'has_avx512bitalg' is enabled, but not available on the host").

Force all AVX-512 cranelift flags to false so the compiled output caps out at AVX2, which is universally available on x86-64 GHA runners.

Closes #1163.

### DIFF
--- a/src/lind-boot/src/lind_wasmtime/execute.rs
+++ b/src/lind-boot/src/lind_wasmtime/execute.rs
@@ -922,6 +922,25 @@ fn make_wasmtime_config(backtrace: bool, enable_fpcast: bool) -> wasmtime::Confi
 
     wt_config.wasm_backtrace_details(details);
 
+    // Disable AVX-512 lanes so the precompiled .cwasm is portable across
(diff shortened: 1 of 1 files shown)


Uk-jake (approved):
I reviewed it — without a pinned runner, host CPU features can leak into the `.cwasm` and cause intermittent failures. Removing the native flags is the right fix since it makes the output portable regardless of the runner

Issue #942, opened 2026-03-13
switch to debug build for lind-boot by default in development


qianxichen233:
Currently make or make lind-boot always builds the release version of lind-boot by default in the development environment. This is not safe. It eliminates many debug checks (e.g., debug_assert and some Rust built-in checks like integer overflow checks) and can mask potential bugs.

We should use debug builds for development by default, and only use release builds for CI (worth discussion as well), performance measurements, or production-related tasks.

PR #1168, opened 2026-05-06
Build lind-boot in debug mode by default

DanielZ224:
## Summary

This PR keeps `make lind-boot` as the default release build and adds a new `make lind-boot-debug` target for building `lind-boot` in debug mode.

While testing the debug build, a Rust debug overflow panic was exposed in GOT memory relocation. This PR also makes that relocation use `wrapping_add`, preserving the existing release-build wrapping behavior explicitly.

## Changes

- Keep `make lind-boot` as the release build target
- Add `make lind-boot-debug` for debug builds
- Use `wrapping_add` for GOT memory relocation to avoid debug-build overflow panic

## Testing

- Ran `make lind-boot`
- Ran `make lind-boot-debug`
- Ran `hello.c` under debug `lind-boot`
- Ran file test harness with debug `lind-boot`

## Related Issue

Related to #942.


Yaxuan-w:
I'd like to have `make lind-boot` as default release version and `make lind-boot-debug` as debug version

rennergade:
> I'd like to have `make lind-boot` as default release version and `make lind-boot-debug` as debug version

I agree here as well.

DanielZ224:
Thanks, that makes sense. I’ll keep `make lind-boot` as the default release build and update the PR to add a separate `make lind-boot-debug` target for debug builds.

qianxichen233:
CI probably needs to run on debug build otherwise it cannot reflect the potential debug build specific issue

DanielZ224:
Updated Dockerfile.e2e so the artifacts target depends on both release and debug e2e stages. The debug stage uses make lind-boot-debug, which should allow CI to catch debug-build-specific regressions.

Yaxuan-w:
ci/cd related solved on #1248

PR #1169, opened 2026-05-07
propagate canonical exit status through exit_group and epoch_kill

qianxichen233:
### Problem

When a process exited with a non-zero code (e.g. faulthandler calling
`_exit(1)`), the parent's `waitpid()` could observe exit code 0 instead.
This is the root reason of python test worker bug observed in python tests.
This was caused by three separate but related issues:

**1. `exit_syscall` recorded the cage exit status**

`exit_syscall` (SYS_exit = 60) is called by glibc's `start_thread` when a
non-main pthread returns from its thread function. It was calling
`cage_record_exit_status(0)` even though a per-thread exit has no bearing on
the process-level exit code. `cage_record_exit_status` uses first-write-wins
semantics, so if SYS_exit(0) fired before `exit_group_syscall(1)`, it locked
`final_exit_status = Some(0)` and the subsequent `exit_group(1)` could not
overwrite it. `cage_finalize` then reported exit code 0 to the parent.

The glibc thread exit sequence is:
```
atomic_store_release(&pd->tid, 0)  // marks thread done, unblocks pthread_join
FUTEX_WAKE(&pd->tid, 1)            // wakes joiner
EXIT_SYSCALL(0)                    // very next instruction
```
`pthread_join` returns right after `FUTEX_WAKE`, so the joiner's return path
(asyncify rewind + multiple library frames) loses the race to thread A's
`cage_record_exit_status(0)`, which fires at the top of `exit_syscall` before
any asyncify unwind.

**2. `exit_group_syscall` passed its own `status_arg` to the exit trampoline**

Late threads calling `exit_group_syscall(0)` after the cage was already marked
`Exited(1)` passed `status_arg = 0` to the wasmtime exit trampoline. The last
thread's exit code determines the value returned to `OnCalledAction::Finish`,
so a late `exit_group(0)` could overwrite the intended code.

**3. `signal_handler` hardcoded exit code 0 for epoch-killed threads**

When `exit_group` or a fatal signal called `epoch_kill_all`, epoch-killed
threads entered `signal_handler` and called `ctx.exit_call(caller, 0, 0)`.
This hardcoded 0 was passed to `OnCalledAction::Finish` regardless of what
exit code had been recorded.

### Fix

- **`exit_syscall`**: remove the `cage_record_exit_status` call entirely. Only
  `exit_group_syscall` determines the process-level exit code.

- **`exit_group_syscall`**: compute `canonical_status_arg` by reading
  `cage.final_exit_status` at the point of calling the exit trampoline.
  Late threads calling `exit_group(0)` after the CAS winner recorded `Exited(1)`
  will use the authoritative code.

- **`signal_handler` (epoch_kill path)**: read `cage.final_exit_status` and
  pass the recorded exit code to `ctx.exit_call` instead of hardcoding 0.

### Test

`tests/unit-tests/process_tests/deterministic/exit_status_first_wins.c`

Forks a child that creates a pthread which returns NULL (triggering
`SYS_exit(0)` via glibc `start_thread`), then calls `exit(1)` after
`pthread_join`. Because `pthread_join` returns before `EXIT_SYSCALL` fires,
thread A's `cage_record_exit_status(0)` races with the main thread's
`exit_group(1)`. Thread A wins the race on the unfixed branch (asyncify rewind
overhead delays the main thread). Parent must observe exit code 1.

### DIFF
--- a/src/rawposix/src/sys_calls.rs
+++ b/src/rawposix/src/sys_calls.rs
@@ -1,6 +1,19 @@
 //! System syscalls implementation
 //!
 //! This module contains all system calls that are being emulated/faked in Lind.
+
+macro_rules! lind_log {
+    ($($arg:tt)*) => {{
+        use std::io::Write;
+        if let Ok(mut f) = std::fs::OpenOptions::new()
+            .create(true).append(true)
+            .open("/tmp/lind.log")
+        {
+            let _ = writeln!(f, $($arg)*);
+        }
+    }};
+}
+
 use crate::fs_calls::kernel_close;
 use cage::memory::vmmap::{VmmapOps, *};
 use cage::signal::signal::{convert_signal_mask, lind_send_signal, signal_check_trigger};
@@ -340,11 +353,15 @@ pub extern "C" fn exit_syscall(
         );
     }
 
-    let status = sc_convert_sysarg_to_i32(status_arg, status_cageid, cageid);
+    let _status = sc_convert_sysarg_to_i32(status_arg, status_cageid, cageid);
 
-    // Thread-only exit: just record status and trigger asyncify unwind.
-    // No CAS, no epoch_kill_all — the cage stays alive for other threads.
-    cage::cage_record_exit_status(cageid, ExitStatus::Exited(status));
+    // Thread-only exit: trigger asyncify unwind for this thread only.
+    // Do NOT call cage_record_exit_status here — SYS_exit is only called by
+    // non-main threads when their thread function returns (glibc start_thread),
+    // and a thread's individual exit code does not determine the cage's
+    // process-level exit status.  Calling it here would race with and
+    // overwrite the exit_group(1) recorded by e.g. the faulthandler thread,
+    // causing the cage to exit with code 0 instead of 1.
 
     // Call wasmtime to trigger asyncify unwind for this thread.
     // OnCalledAction handles lind_thread_exit + cage_finalize if last.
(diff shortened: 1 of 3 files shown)


rennergade:
One concern: the `Signaled → 1` mapping at `sys_calls.rs:444` and `signal.rs:56` reintroduces the same "last thread wins" non-determinism the PR is trying to eliminate, just in the signal-termination case instead of the explicit-exit case. The Terminate handler at `signal.rs:115` calls `exit_call(caller, 128 + signo, 0)` directly, but a sibling thread that gets epoch-killed and exits via the `thread_check_killed` branch will read `Signaled(_, _)` from `final_exit_status` and pass `1` instead — so whichever thread happens to call the trampoline last determines whether the wasmtime/OS-level exit code is `128 + signo` or `1`. `waitpid`'s `WIFSIGNALED` decoding still works because it goes through `encode_wait_status`, but anyone reading the process's shell-level `$?` sees a non-deterministic value. Both mapping sites should produce `128 + (signo & 0x7f)` to match what the Terminate path already passes.

qianxichen233:
> One concern: the `Signaled → 1` mapping at `sys_calls.rs:444` and `signal.rs:56` reintroduces the same "last thread wins" non-determinism the PR is trying to eliminate, just in the signal-termination case instead of the explicit-exit case. The Terminate handler at `signal.rs:115` calls `exit_call(caller, 128 + signo, 0)` directly, but a sibling thread that gets epoch-killed and exits via the `thread_check_killed` branch will read `Signaled(_, _)` from `final_exit_status` and pass `1` instead — so whichever thread happens to call the trampoline last determines whether the wasmtime/OS-level exit code is `128 + signo` or `1`. `waitpid`'s `WIFSIGNALED` decoding still works because it goes through `encode_wait_status`, but anyone reading the process's shell-level `$?` sees a non-deterministic value. Both mapping sites should produce `128 + (signo & 0x7f)` to match what the Terminate path already passes.

There are two different exit routine, one is for wasm module exit code (which propogates exit code via wasm semantic), and one is for waitpid semantic exit code (which records the exit code in the cage struct). the wasm module exit code currently isn't getting handled (e.g. the exit code of forked module is silently dropped). This is like thread exit code v.s. process exit code, where we currently only handled process exit code but not thread exit code. This PR basically removed recording of thread exit code (since it previously would overwrite process exit code) and force it to 1. This is a temporary solution before we add/fix the support for thread exit code (which will need to implement pthread_exit as well). The process exit code recorded on cage struct remains unchanged.

PR #1172, opened 2026-05-08
utimensat: implement syscall and fix stat time-field copy

### LINKED ISSUE #1171: Coreutils tests failing
Two independent bugs in the file-time path. Together they cause every coreutils test that depends on file timestamps (`touch/relative, touch/empty-file, touch/no-rights, touch/now-owned-by-other, touch/trailing-slash, cp/preserve-slink-time`, etc.) to fail.

Bug 1: `utimensat` system call not implemented. It returns 0 always

Bug 2 : `stat` doesn't copy time fields from libc::stat to user's StatData

```
pub fn convert_statdata_to_user(stat_ptr: &mut StatData, libc_statbuf: libc::stat) {
      stat_ptr.st_blksize = libc_statbuf.st_blksize as i32;
      stat_ptr.st_blocks  = libc_statbuf.st_blocks as u32;
      .....
      .....
      stat_ptr.st_uid     = libc_statbuf.st_uid;
  }
```

The destination StatData (defined at [src/sysdefs/src/data/fs_struct.rs:47-49](https://github.com/Lind-Project/lind-wasm/blob/9c416864b00a73933df8f583f1c844aa64930f9d/src/sysdefs/src/data/fs_struct.rs#L47)) has:
  
  ```
  pub st_atim: (u64, u64),
  pub st_mtim: (u64, u64),
  pub st_ctim: (u64, u64),
```
  …but they are never assigned. So even if Bug 1 were fixed and the host kernel held a real mtime, the cage program would still read 0.

### Reproduction
  Coreutils test `touch/relative`

 ```
TZ=UTC0 touch --date='2004-01-16 12:00 +0000' f
 touch --ref f --date='-5 days' f
 TZ=UTC0 ls -og --time-style=+%Y-%m-%d f
```
  Expected: 2004-01-11. Actual: 1970-01-01.

rennergade:
Two related bugs broke every coreutils test that depends on file timestamps (touch/relative, touch/empty-file, cp/preserve-slink-time, ...).

1) utimensat was unimplemented.  No constant in syscall_const.rs or
   lind_syscall_num.h; no handler in rawposix; not registered in
   SYSCALL_TABLE.  The glibc path used INLINE_SYSCALL_CALL with the
   raw Linux syscall number, which hit the dispatcher with no handler
   and produced garbage results.  This adds:
     - UTIMENSAT_SYSCALL = 280 in both syscall-number sources.
     - utimensat_syscall in fs_calls.rs that:
         * handles AT_FDCWD vs explicit dirfd,
         * dispatches to libc::futimens for the (fd, NULL path) form glibc uses to implement futimens(), * passes timespec[2] / NULL through verbatim so UTIME_NOW and UTIME_OMIT work without translation. - A glibc-side utimensat.c rewrite that uses MAKE_LEGACY_SYSCALL with TRANSLATE_GUEST_POINTER_TO_HOST on file and tsp64, matching chmod / openat / etc.  All callers (futimens, utime, futimesat) go through __utimensat64_helper, so a single rewrite covers them.

2) convert_statdata_to_user (typemap/filesystem_helpers.rs) copied every
   field except st_atim / st_mtim / st_ctim — so even with a working
   utimensat, stat() round-trips the timestamps as zero.  Add the three
   missing tuple assignments (sec, nsec).  Affects stat / lstat / fstat.

Test: tests/unit-tests/file_tests/deterministic/utimensat.c covers
  - utimensat(AT_FDCWD) + stat round-trip,
  - futimens + fstat round-trip,
  - UTIME_NOW + UTIME_OMIT semantics (atime advances, mtime preserved). The test passes on native Linux as a sanity check on the assertions.

Closes #1171.


src/rawposix/src/fs_calls.rs
```diff
@@ -3656,6 +3656,98 @@ pub extern "C" fn fchmod_syscall(
...
+        // portable form is utimensat with empty path + AT_EMPTY_PATH.
+        // We use the libc::futimens wrapper to keep this kernel-portable.
+        unsafe { libc::futimens(kernel_fd, times) }
+    } else {
+        let path = match sc_convert_path_to_host(path_arg, path_cageid, cageid) {
+            Ok(p) => p,
+            Err(e) => return syscall_error(e, "utimensat", "path conversion failed"),
+        };
+        if virtual_fd == AT_FDCWD {
+            unsafe { libc::utimensat(libc::AT_FDCWD, path.as_ptr(), times, flags) }
```
Yaxuan-w: Should we import from sysdefs for constants

vidyalakshmir (approved):

Issue #1173, opened 2026-05-09
Missing openat and fstatat syscalls — rm -r, cp -r, chmod -R, and other recursive coreutils   silently corrupt traversal


vidyalakshmir:
## Description

  Missing `openat` (Linux syscall 257) and `newfstatat / statx` (262 / 332) system call implementations in lind-wasm. Any program that uses gnulib's fts or the POSIX *at traversal API directly is silently broken - recursive removal, copy, and permission-change commands all appear to "succeed" partway, leave files in place, and then fail at the parent-directory step with confusing errors like `Directory not empty`.

  Found while triaging the coreutils correctness suite — rm/v-slash was the trigger.

PR #1174, opened 2026-05-09
Port *at family of fs syscalls to MAKE_LEGACY_SYSCALL (#1173)

rennergade:
Closes the broader form of #1173. Most of glibc's `*at` filesystem wrappers in sysdeps/unix/sysv/linux/ were on `INLINE_SYSCALL_CALL`, which expands to literal `0` in this WASM port — so they returned success without doing anything. fts traversal (used by every recursive coreutil) silently mis-stat'd entries, then handed bogus fds to fchmod and tripped EBADF on cleanup. ~117 of the 136 coreutils failures in the latest categorization CSV trace to this.

Glibc-side: route through MAKE_LEGACY_SYSCALL with translated guest pointers, mirroring the existing utimensat/openat/unlinkat ports.
- fchmodat.c, faccessat.c, fchownat.c (new file under sysdeps/unix/sysv/linux replacing the io/ ENOSYS stub), renameat.c, renameat2.c, chown.c, lchown.c, dl-openat64.c
- fstatat64.c: also fix the silently-stubbed fstatat64_time64_stat path that was returning success with an unfilled buffer.
- statx.c: skip the broken syscall path and use the fstatat-based statx_generic fallback.
- lstat64.c: route via NEWFSTATAT_SYSCALL with AT_SYMLINK_NOFOLLOW so it actually does lstat semantics. (Previously misrouted to XSTAT_SYSCALL — stat semantics — per a TODO note.)

RawPOSIX-side: add 8 new syscall handlers in fs_calls.rs modeled on unlinkat_syscall, register them in syscall_table.rs, and add constants for NEWFSTATAT_SYSCALL (262), FCHOWNAT_SYSCALL (260), RENAMEAT_SYSCALL (264), FCHMODAT_SYSCALL (268),
FACCESSAT_SYSCALL (269), RENAMEAT2_SYSCALL (316),
STATX_SYSCALL (332), CHOWN_SYSCALL (92), LCHOWN_SYSCALL (94) in both syscall_const.rs and lind_syscall_num.h.

Skipped for follow-up: execveat.c, fexecve.c (need exec adapter work), and ~50 unrelated INLINE_SYSCALL_CALL no-ops that don't affect coreutils (timer/sched/IPC/etc).


src/glibc/sysdeps/unix/sysv/linux/faccessat.c
```diff
@@ -16,61 +16,31 @@
...
-    return ret;
-
   if (flag & ~(AT_SYMLINK_NOFOLLOW | AT_EACCESS))
     return INLINE_SYSCALL_ERROR_RETURN_VALUE (EINVAL);
 
-  if ((flag == 0 || ((flag & ~AT_EACCESS) == 0 && ! __libc_enable_secure)))
-    return INLINE_SYSCALL (faccessat, 3, fd, file, mode);
-
-  struct __stat64_t64 stats;
-  if (__fstatat64_time64 (fd, file, &stats, flag & AT_SYMLINK_NOFOLLOW))
```
vidyalakshmir, line 43: Should these condition checks be retained? Or is the idea that everything shud be passed to rawposix?

src/glibc/sysdeps/unix/sysv/linux/faccessat.c
```diff
@@ -16,61 +16,31 @@
...
-    return -1;
-
-  mode &= (X_OK | W_OK | R_OK);	/* Clear any bogus bits. */
```
vidyalakshmir, line 46: mode is passed to rawposix.Is it ok to remove this?

src/glibc/sysdeps/unix/sysv/linux/fchmodat.c
```diff
@@ -18,88 +18,32 @@
...
-#else
-  if (flag == 0)
-    return INLINE_SYSCALL_CALL (fchmodat, fd, file, mode);
-
-  int r = INLINE_SYSCALL_CALL (fchmodat2, fd, file, mode, flag);
-  if (r != 0 && errno == ENOSYS)
-    return fchmodat_fallback (fd, file, mode, flag);
-  return r;
-#endif
+  return MAKE_LEGACY_SYSCALL (FCHMODAT_SYSCALL, "syscall|fchmodat",
```
vidyalakshmir, line 42: Similar comments here. The original function does check for multiple conditions and in some cases returns error, and other cases invokes other syscalls etc. Would it be okay to replace with a single call to rawposix?

src/glibc/sysdeps/unix/sysv/linux/faccessat.c
rennergade: host kernel will validate

vidyalakshmir (approved):

PR #1176, opened 2026-05-09
handle exec-ed module failure

qianxichen233:
when exec-ed module crashed, host will panic with the error message instead of treating the crash as exit code of 1 and propogate.

This PR adds the handling of failed exec-ed module execution, to match the behavior when wasm module crashed in fork/thread/signal routine


rennergade:
can we add a test for this?

qianxichen233:
we can, but it probably needs coordination of CI since it needs a dependent program to exec, see this commit: https://github.com/Lind-Project/lind-wasm/pull/1176/changes/131e3f491ade191d2a61d99b8db2d6ddae5f5da4.

I am not sure if there is easy way to add a new staged binary to automated_tests folder when running the CI

Issue #939, opened 2026-03-13
pointer misalignment panic in debug build of lind boot


qianxichen233:
There are 20 tests failing on debug build of lind-boot

Most of them are due to errors like `misaligned pointer dereference: address must be a multiple of 0x8 but is 0x786d2c02ecec`

My guess is that rust only does checks for the alignment when dereference the pointer in debug build, and the release build optimized it out

PR #1184, opened 2026-05-10
Fix debug runtime overflow and unaligned unwind metadata access

DanielZ224:
## Summary

This PR fixes debug-build runtime failures exposed by grate tests involving fork/thread execution.

The failures appeared in debug `lind-boot` runs as:

- `misaligned pointer dereference` in `lind-multi-process/src/lib.rs`
- `attempt to add with overflow` in `wasmtime/src/runtime/instance.rs`
- Grate test timeouts after the initial panic was avoided

The root cause was that debug builds enforce stricter runtime checks than release builds. Some asyncify unwind metadata was accessed through direct `u64` pointer dereferences, even though the unwind data address may not always be 8-byte aligned. In addition, GOT memory relocation used checked debug addition, which can panic on overflow even though release builds effectively use wrapping arithmetic.

## Changes

- Replaced direct `u64` pointer reads/writes on asyncify unwind metadata with:
  - `std::ptr::read_unaligned`
  - `std::ptr::write_unaligned`

- Updated GOT memory relocation arithmetic from:

```rust
let val = raw as u32 + memory_base;
```

to

```rust
let val = (raw as u32).wrapping_add(memory_base);
```

This makes debug behavior consistent with release behavior and prevents debug-only overflow panics during GOT relocation.

## Validation

Built debug lind-boot:

```bash
cargo build --manifest-path src/lind-boot/Cargo.toml \
  --no-default-features --features fdtables-dashmaparray

cp src/lind-boot/target/debug/lind-boot build/lind-boot
```

Ran the failing grate tests with debug lind-boot:

```bash
RUST_BACKTRACE=1 python3 scripts/harnesses/gratetestreport.py \
  --testfiles \
  tests/grate-tests/concurrent-request/race-test_grate.c \
  tests/grate-tests/concurrent-request/thread_race_grate.c \
  --timeout 600 \
  --debug \
  --generate-html
```

Result:

```text
[INFO] [1/2] concurrent-request/race-test_grate.c
[INFO] SUCCESS
[INFO] [2/2] concurrent-request/thread_race_grate.c
[INFO] SUCCESS
```

Also verified a reduced temporary reproduction case with debug lind-boot:

```text
tmp-debug-939/race-test-small_grate.c: SUCCESS
```

## Notes

The temporary reduced test files and generated reports were not included in this PR.

## Related Issue

Related to #939.


qianxichen233:
what’s causing these addresses to be unaligned in some cases? Is there a better way to fix the root cause, for example by ensuring the addresses are properly aligned, rather than allowing unaligned writes?

I don’t think we should blindly bypass debug-build errors through special handling. In general, it’s better to respect the debug check and fix the underlying issue, unless doing so would introduce significant overhead

DanielZ224:
Thanks for the comment. I rechecked the root cause instead of relying on unaligned u64 accesses.

The asyncify unwind header is a 32-bit layout: buf[0] stores the current unwind position and buf[4] stores the end pointer. The previous u64 writes were not just an alignment issue; they also wrote across both 32-bit fields and could corrupt the asyncify bounds state. I updated these sites to write/read the asyncify metadata as u32 fields.

For jmp_buf, the guest-provided address can be 4-byte aligned rather than 8-byte aligned in this environment, so I kept unaligned access only for the stored 64-bit hash there. The asyncify metadata itself is no longer handled as unaligned u64 data.

I also verified this with a debug build and reran the two grate regression tests:
- concurrent-request/race-test_grate.c
- concurrent-request/thread_race_grate.c

Both now pass.

qianxichen233:
> Thanks for the comment. I rechecked the root cause instead of relying on unaligned u64 accesses.
> 
> The asyncify unwind header is a 32-bit layout: buf[0] stores the current unwind position and buf[4] stores the end pointer. The previous u64 writes were not just an alignment issue; they also wrote across both 32-bit fields and could corrupt the asyncify bounds state. I updated these sites to write/read the asyncify metadata as u32 fields.
> 
> For jmp_buf, the guest-provided address can be 4-byte aligned rather than 8-byte aligned in this environment, so I kept unaligned access only for the stored 64-bit hash there. The asyncify metadata itself is no longer handled as unaligned u64 data.
> 
> I also verified this with a debug build and reran the two grate regression tests:
> 
> * concurrent-request/race-test_grate.c
> * concurrent-request/thread_race_grate.c
> 
> Both now pass.

The issue with asyncify buf size was already fixed before, it's also not reflected in your changed files as well. In your changed files, you are trying to replace all the write with write_unaligned. `write_unaligned` is only required when the address to write is unaligned. The justification needs to be mentioned here is why the address to be written (i.e. unwind_data_start_sys) is an unaligned address. FYI: `unwind_data_start_sys` is usually the start address of stack, which is not an user assigned address.

DanielZ224:
Updated the fix to avoid using unaligned access for the Asyncify metadata path. These metadata addresses are expected to be aligned, so I changed them back to regular aligned u32 loads/stores instead of bypassing the debug check with read_unaligned/write_unaligned.

I kept the separate jmp_buf hash handling as unaligned access because the guest-side jmp_buf address can be unaligned in practice. I also verified this with a debug lind-boot build and reran the original concurrent-request/race-test_grate.c and concurrent-request/thread_race_grate.c tests; both passed.

PR #1185, opened 2026-05-10
Honor GRATE_MEMORY_FLAG in mmap_syscall and brk_syscall

rennergade:
`mmap_syscall` used to do `useraddr = addr as u32` unconditionally, truncating the high 32 bits of any host pointer a grate had supplied via `GRATE_MEMORY_FLAG`. The truncated low bits then got re-translated through the calling cage's vmmap, landing the mapping at an arbitrary cage address. The cage's subsequent memcpy clobbered whatever lived there (stack, locals, fds) and downstream syscalls failed in confusing ways that didn't trace back to the mmap. `brk_syscall` had the inverse gap: it always interprets the address as a cage uaddr, so a grate that wanted to set the program break via a host sysaddr had no way to do it.

**Implementation note (subtle):** `GRATE_MEMORY_FLAG` is consumed glibc-side inside `make_threei_call` / `TRANSLATE_ARG_TO_HOST` — by the time the runtime sees the call, the flag bit has been stripped from `arg_cageid` and (if it was set) the arg is the already-resolved host pointer. The runtime therefore can't check the flag directly. It distinguishes the two cases by **value range**: cage linear memory occupies `[base, base+4GB]` on the host, and host bases are far above 4GB, so any arg above `u32::MAX` must be a pre-translated host sysaddr.

Adds two helpers in `typemap::datatype_conversion`:

- `sc_convert_addr_to_sys(arg, arg_cageid, cageid)`: returns the host sysaddr. `arg == 0` anchors to the calling cage's base (so `early_init_stack`-style `addr=0,MAP_FIXED` calls still work). `arg ≤ u32::MAX` is treated as a uaddr and translated through the calling cage's vmmap. `arg > u32::MAX` is used as-is (grate already translated it).
- `sc_convert_sys_to_user(sysaddr, cageid)`: inverse, for return-value bookkeeping and the brk case.

Wired into both branches of `mmap_syscall` (find_map_space hint resolution and the MAP_FIXED path) and into `brk_syscall` (sysaddr → uaddr when the arg looks like a host sysaddr). Adds `GRATE_MEMORY_FLAG` and `LIND_ARG_CAGEID_MASK` constants in `sysdefs/lind_platform_const.rs` to mirror glibc's `addr_translation.h`.

**Scope:** only `mmap` / `mmap64` (uaddr convention per glibc `mmap.c`) and `brk` (raw uaddr per glibc `brk.c`) need this. `munmap`, `mprotect`, `shmat`, `shmdt` already receive pre-translated sysaddrs from their glibc wrappers (via `TRANSLATE_GUEST_POINTER_TO_HOST`), so the runtime already uses them correctly regardless of the flag — leaving them untouched avoids double-translating the cage path.

## Test

`tests/grate-tests/simple-tests/mmap-flag.c` + `mmap-flag_grate.c`. The grate registers an mmap handler that forwards fd-backed mmaps to RawPOSIX with `GRATE_MEMORY_FLAG` on `arg1cage`, exercising the runtime's value-range path. Anonymous mmaps (including the cage's runtime-internal `instance.rs` setup mmap) are forwarded unchanged so they don't trip the FLAG path. The cage opens a file, mmaps it MAP_SHARED, writes/reads back a page, munmaps.

### DIFF
--- a/src/sysdefs/src/constants/lind_platform_const.rs
+++ b/src/sysdefs/src/constants/lind_platform_const.rs
@@ -32,6 +32,13 @@ pub const MAXFD: usize = 1024; // Maximum file descriptors per cage
 pub const MAX_LINEAR_MEMORY_SIZE: u64 = 0xFFFF_FFFF;
 /// Placeholder for unused syscall argument
 pub const UNUSED_ARG: u64 = 0xDEADBEEF_DEADBEEF;
+/// MSB of a syscall arg's cageid: signals that the arg should be treated as a
+/// host-side reference into the named cage's linear memory, not as a uaddr
+/// in the calling cage's memory. Mirrors `LIND_ARG_TRANSLATE_FLAG` in
+/// `src/glibc/lind_syscall/addr_translation.h`.
+pub const GRATE_MEMORY_FLAG: u64 = 1u64 << 63;
+/// Mask to recover the actual cageid by clearing `GRATE_MEMORY_FLAG`.
+pub const LIND_ARG_CAGEID_MASK: u64 = !GRATE_MEMORY_FLAG;
 /// Placeholder for unused cage/grate ID
 pub const UNUSED_ID: u64 = 0xCAFEBABE_CAFEBABE;
 /// Placeholder for unused syscall name
--- a/src/sysdefs/src/constants/lind_platform_const.rs
+++ b/src/sysdefs/src/constants/lind_platform_const.rs
@@ -32,13 +32,27 @@ pub const MAXFD: usize = 1024; // Maximum file descriptors per cage
 pub const MAX_LINEAR_MEMORY_SIZE: u64 = 0xFFFF_FFFF;
 /// Placeholder for unused syscall argument
 pub const UNUSED_ARG: u64 = 0xDEADBEEF_DEADBEEF;
-/// MSB of a syscall arg's cageid: signals that the arg should be treated as a
-/// host-side reference into the named cage's linear memory, not as a uaddr
-/// in the calling cage's memory. Mirrors `LIND_ARG_TRANSLATE_FLAG` in
+/// MSB of a syscall arg's cageid: requests guest-to-host address translation
+/// for that arg. Mirrors `LIND_ARG_TRANSLATE_FLAG` in
 /// `src/glibc/lind_syscall/addr_translation.h`.
 pub const GRATE_MEMORY_FLAG: u64 = 1u64 << 63;
 /// Mask to recover the actual cageid by clearing `GRATE_MEMORY_FLAG`.
 pub const LIND_ARG_CAGEID_MASK: u64 = !GRATE_MEMORY_FLAG;
+
+/// Return the actual cageid after clearing any syscall address-translation bit.
+pub const fn lind_arg_cageid(cageid: u64) -> u64 {
+    cageid & LIND_ARG_CAGEID_MASK
+}
+
+/// Return true when a syscall argument cageid requests address translation.
+pub const fn lind_arg_needs_translation(cageid: u64) -> bool {
+    cageid & GRATE_MEMORY_FLAG != 0
+}
+
+/// Mark a syscall argument cageid as requiring address translation.
+pub const fn lind_arg_with_translation(cageid: u64) -> u64 {
+    cageid | GRATE_MEMORY_FLAG
+}
 /// Placeholder for unused cage/grate ID
 pub const UNUSED_ID: u64 = 0xCAFEBABE_CAFEBABE;
 /// Placeholder for unused syscall name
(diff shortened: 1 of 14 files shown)


src/typemap/src/datatype_conversion.rs
```diff
@@ -260,6 +260,89 @@ pub fn sc_convert_to_u8_mut(arg: u64, arg_cageid: u64, cageid: u64) -> *mut u8 {
...
+/// on the host, and host bases are typically far above 4GB.
+///
+/// arg=0 is special-cased to "start of the calling cage's memory" — a NULL
+/// host pointer would be meaningless.
+///
+/// ## Returns
+/// - `Ok(sysaddr)` host system address.
+/// - `Err(Errno::EINVAL)` if the calling cage can't be looked up or its vmmap
+///   has no base address yet (only on the uaddr branch).
+pub fn sc_convert_addr_to_sys(arg: u64, arg_cageid: u64, cageid: u64) -> Result<usize, Errno> {
```
stupendoussuperpowers: I think we already have a `sc_convert_addr_to_host` ? We can just update that directly, they are not used anywhere else.

stupendoussuperpowers (dismissed):

src/typemap/src/datatype_conversion.rs
```diff
@@ -290,21 +320,14 @@ pub fn sc_convert_uaddr_to_host(uaddr: u64, addr_cageid: u64, _cageid: u64) -> u
...
-    if uaddr == 0 {
-        return uaddr;
-    }
-
     let cage = get_cage(addr_cageid).unwrap();
     let vmmap = cage.vmmap.read();
     let base_addr = vmmap.base_address.unwrap() as u64;
 
-    // Only convert to host if not already converted.
     if uaddr >= base_addr {
```
JustinCappos: I think this should check if the top bits match the base_addr, not this.

In practice, you should have a helper function for this to handle situations where we do not have 32 bit aligned 4GB regions.

src/threei/src/handler_table/hashmap_impl.rs
```diff
@@ -1,7 +1,6 @@
 use crate::threei_const;
-use std::collections::{hash_map::Entry, HashMap};
+use std::collections::HashMap;
 use std::sync::Mutex;
-use sysdefs::constants::lind_platform_const;
```
JustinCappos, line 4: Is this linter cleanup?

src/threei/src/handler_table/hashmap_impl.rs
```diff
@@ -73,7 +72,7 @@ pub fn _check_cage_handler_exists(cageid: u64) -> bool {
 /// ## Panics:
 ///     - If no entry exists for `self_cageid`.
 ///     - If no entry exists for `syscall_num`.
-pub fn _get_handler(self_cageid: u64, syscall_num: u64, target_cageid: u64) -> Option<(u64, u64)> {
+pub fn _get_handler(self_cageid: u64, syscall_num: u64, _target_cageid: u64) -> Option<(u64, u64)> {
```
JustinCappos: Are you making the linter happy, or...?

src/sysdefs/src/constants/lind_platform_const.rs
```diff
@@ -32,6 +32,13 @@ pub const MAXFD: usize = 1024; // Maximum file descriptors per cage
 pub const MAX_LINEAR_MEMORY_SIZE: u64 = 0xFFFF_FFFF;
 /// Placeholder for unused syscall argument
 pub const UNUSED_ARG: u64 = 0xDEADBEEF_DEADBEEF;
+/// MSB of a syscall arg's cageid: signals that the arg should be treated as a
+/// host-side reference into the named cage's linear memory, not as a uaddr
+/// in the calling cage's memory. Mirrors `LIND_ARG_TRANSLATE_FLAG` in
+/// `src/glibc/lind_syscall/addr_translation.h`.
+pub const GRATE_MEMORY_FLAG: u64 = 1u64 << 63;
```
JustinCappos: I don't really understand this.  Need to discuss

src/sysdefs/src/constants/lind_platform_const.rs
```diff
@@ -32,6 +32,13 @@ pub const MAXFD: usize = 1024; // Maximum file descriptors per cage
...
+/// Mask to recover the actual cageid by clearing `GRATE_MEMORY_FLAG`.
+pub const LIND_ARG_CAGEID_MASK: u64 = !GRATE_MEMORY_FLAG;
```
JustinCappos, line 40: Would it be better to make a macro that uses this?

src/threei/src/handler_table/hashmap_impl.rs
rennergade: yeah

Issue #1183, opened 2026-05-09
__libc_fcntl64 unconditionally translates arg as a guest pointer


rishabhBudhouliya:
fnctl64() translates the variadic argument as a pointer unconditionally that results an incorrect fd. https://linux.die.net/man/2/fcntl64 the third arg depends on cmd's value and we should translate it as an integer value or a pointer conditional upon cmd's value 

https://github.com/Lind-Project/lind-wasm/blob/0a12809cbf75c4bb5a9d0df6a7018c97d138b9b7/src/glibc/sysdeps/unix/sysv/linux/fcntl64.c#L47-L58

Coreutils test suite success numbers go from 162 -> 181 after fixing this issue

PR #1187, opened 2026-05-11
fcntl64: make variadic argument processing conditional to cmd

rishabhBudhouliya:
# Purpose
`src/glibc/sysdeps/unix/sysv/linux/fcntl64.c` function `__libc_fcntl64` (which is what `fcntl()` resolves to via `weak_alias` on lines 63 and 72) extracted the variadic third argument as `void *` and unconditionally ran it through `TRANSLATE_GUEST_POINTER_TO_HOST` regardless of `cmd`:

```c
arg = va_arg (ap, void *);
...
uint64_t host_arg = TRANSLATE_GUEST_POINTER_TO_HOST (arg);
return MAKE_LEGACY_SYSCALL(FCNTL_SYSCALL, ..., host_arg, ...);
```

Bug: https://github.com/Lind-Project/lind-wasm/issues/1183


rennergade:
@rishabhBudhouliya is there a test that confirms this fix

PR #1188, opened 2026-05-11
Added Tests for mmap address truncation and null mapping

harindham:
## Two memory safety tests added:

1. mmap_address_truncation.c: Exposes Address truncation and silent wrap-around when range overflows. 
2. mmap_null_address_mapping.c: Exposes null address mapping (related to the issue #1065)


Yaxuan-w (commented):
Those tests should put into regular folder since they "should" success eventually, but fails caused by bugs (that's why we need to put into skip_test_cases.txt to indicate such scenario)

PR #1195, opened 2026-05-14
Optimize grate call overhead

rennergade:
This reduces grate worker overhead for benchmarking and cheap grate calls.

  Changes:

  - Defaults grate worker pool size to 1 instead of eagerly creating MAX_GRATE_WORKERS.
  - Adds LIND_GRATE_WORKERS to opt into more workers when needed.
  - Clamps configured worker count to 1..=MAX_GRATE_WORKERS.
  - Caches each worker’s __stack_pointer global instead of resolving it on every call.
  - Avoids unconditional condvar wakeups on every grate call return/shutdown guard drop.


src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -235,7 +239,7 @@ struct GrateWorker<T> {
     /// Worker-local instance of the grate module.
     ///
     /// Calls submitted to this worker execute inside this instance.
-    instance: Instance,
+    _instance: Instance,
```
Yaxuan-w: We can just remove this field if not needed.

src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -280,6 +291,15 @@ fn worker_stack_top(cageid: u64, workerid: WorkerId) -> u32 {
     worker_stack_base(cageid, workerid) + GRATE_STACK_SLOT_SIZE
 }
 
+fn configured_grate_workers() -> usize {
```
Yaxuan-w, line 290: Worth commenting the default size is still MAX_GRATE_WORKER but users can configure by setting env var

src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -721,8 +747,8 @@ where
 /// mode, calls still use the same worker-pool abstraction, but entry is gated
 /// so that only one call runs at a time.
 ///
-/// This function performs eager worker creation so that the handler is ready to
-/// serve grate calls immediately after registration.
+/// This function performs eager worker creation for `LIND_GRATE_WORKERS`
+/// workers, defaulting to one worker to keep cage startup cheap.
```
Yaxuan-w: Current version not defaulting to 1, but configurable

Yaxuan-w (changes requested):
Minor comments. After resolving merge conflicts and comments this should be good to merge

Yaxuan-w (approved):

Issue #1161, opened 2026-05-05
investigate setjmp/longjmp implementation alternative


qianxichen233:
PostgreSQL uses setjmp/longjmp heavily for error handling. This is normally fine on native Linux because setjmp is cheap: it mostly saves a small amount of context and returns.

In lind, the current setjmp implementation relies on Asyncify. This makes setjmp much more expensive, because it may involve Asyncify stack-unwind/capture machinery.

The main issue is that setjmp is on the normal execution path and may be called very frequently. If every setjmp requires Asyncify machinery, applications like PostgreSQL can become extremely slow even when longjmp is rarely used.

Therefore, we need to investigate an alternative setjmp/longjmp implementation for lind where setjmp is cheap and the higher cost is paid only on longjmp.

One possible alternative that worth investigating:

LLVM/WebAssembly appears to support SJLJ lowering using WebAssembly exception handling, possibly via: `-mllvm -wasm-enable-sjlj` and Emscripten exposes a related mode with: `-sSUPPORT_LONGJMP=wasm`. This approach may implement longjmp by throwing a Wasm exception and unwinding to the matching setjmp point, avoiding Asyncify full unwind on every setjmp. Further investigation are needed, but this seems to be a promising alternative to current setjmp/longjmp implementation with much less setjmp overhead. It is known that we must at least upgrade wasmtime engine with new wasm exception handling feature enabled to make it work.

PR #1200, opened 2026-05-25
EH-based setjmp/longjmp, memory management refactor, pause/sigsuspend support

### LINKED ISSUE #899: Implement async sigsuspend
Context                                                                                                                                                                              
                                                                                                                                                                                     
The current sigsuspend implementation handles the common case where signals are already pending when sigsuspend is called — e.g., kill(getpid(), SIG) followed by sigsuspend(). It works by swapping the signal mask via sigprocmask, calling sched_yield to trigger the epoch callback (which delivers pending signals synchronously), then restoring the original mask.
                                                                                                                                                                                     
Limitation                                                                                                                                                                           
                                                                                                                                                                                     
Real sigsuspend blocks until a signal is delivered. The current implementation returns immediately after one sched_yield, so if no signal is pending at call time (e.g., waiting for a signal from another thread or timer), it returns -1/EINTR without any signal having been delivered.                                                                                
                                                                                                                                                                                     
Why we can't detect delivery from glibc:                                                                                                                               
                                                                                                                                                                                     
In Lind's epoch-based model, signal handlers run synchronously inside the Rust signal_handler() epoch callback, which runs inside the sched_yield host call. After the handler returns, sched_yield returns to glibc — but glibc has no way to know whether a handler actually ran during that call.                                                                
                                                                                                                                                                                     
Proposed fix                                                                                                                                                                         
                                                                                                                                                                                     
Add a Condvar to the Cage struct that sigsuspend can wait on. When lind_send_signal() queues a new signal in pending_signals, it notifies the condvar to wake any thread blocked in sigsuspend.

qianxichen233:
Closes https://github.com/Lind-Project/lind-wasm/issues/899 and https://github.com/Lind-Project/lind-wasm/issues/1161

This PR contains three closely related changes that were developed together because each one was required to make the others work correctly.

### 1. EH-based setjmp/longjmp

Replaces the previous asyncify-based setjmp/longjmp with an implementation built on wasm exception handling. The new default uses `-fwasm-exceptions -mllvm -wasm-enable-sjlj` (clang 18) together with the `saveSetjmp`/`testSetjmp`/`__wasm_longjmp` runtime in `src/glibc/setjmp/wasm_eh_setjmp.c`. The old asyncify path is preserved and still selectable via `LIND_ASYNCIFY_SETJMP=1`.

For full implementation details, design decisions, known limitations, and open tasks see `docs/internal/setjmp.md`.

It's measured that EH-based setjmp implementation is at least 100x times faster than old Asyncify-based setjmp, and postgres' pgbench can run much faster with it

### 2. Linear memory management refactor

The EH implementation exposed a latent bug: wasmtime allocates a GC heap for wasm exception objects using `Mmap::reserve()` followed by `make_accessible()`. In lind-wasm, `make_accessible` had been made a no-op (to prevent wasmtime from mprotecting wasm linear memory), so the GC heap remained `PROT_NONE` and the first `longjmp` caused a SIGSEGV inside JIT-compiled code.

The fix clarifies ownership of `mprotect`:
- `Mmap::make_accessible` is restored to call real `mprotect`. It is correct for host-internal allocations (GC heap, code memory, etc.) that wasmtime fully owns.
- PROT_NONE enforcement for wasm *linear* memory is moved to `attach_shared_memory` (called once per cage, before `init_vmmap`), and to the `ClonedMemory::New` arm in `new_child_linker` (fork children). Rawposix vmmap then takes ownership of those pages in a clean PROT_NONE state.

### 3. `pause` syscall and `sigsuspend` fix

A key test for EH setjmp is longjmp from a signal handler: the handler throws `__c_longjmp` through `signal_callback` and `pause()` with no Rust boundary in the unwind path. This required:
- Implementing the `pause` syscall in rawposix.
- Fixing `sigsuspend` to set the signal mask before blocking, so a signal queued before the call is delivered correctly.

These syscalls are also independently useful; the setjmp tests just exposed that they were missing or incomplete.

## Tests

- `tests/unit-tests/process_tests/deterministic/setjmp_edge.c` — comprehensive single-module edge cases
- `tests/unit-tests/process_tests/deterministic/test_crossmodule_longjmp.c` — cross-module EH tag sharing and signal path
- `tests/unit-tests/signal_tests/deterministic/test_sigsetjmp.c` — sigsetjmp/siglongjmp API
- `tests/unit-tests/dylink_tests/deterministic/longjmp_dlopen.c` — dlopen cross-module (skipped in main runner)
- `tests/unit-tests/memory_tests/fail/invalid_access_direct.c` — PROT_NONE enforcement (direct)
- `tests/unit-tests/memory_tests/fail/invalid_access_fork.c` — PROT_NONE enforcement (fork child)


rennergade:
A few issues:

- sigsetjmp can fail to compile for valid callers that include only <setjmp.h> I think you need to define SIG_BLOCK in setjmp.h or avoid the public dependency
- I don't think sigsuspend functions atomically at all (or if thats possible) the alarm test you do doesnt really expose that. But: If a pending signal is unblocked by the mask change, rawposix can trigger delivery around /tmp/lind-wasm-pr1200/src/rawposix/src/sys_calls.rs:1182 before pause() runs, after the handler returns, pause() can block forever.

rennergade (dismissed):

rennergade (approved):

Yaxuan-w (approved):

PR #1205, opened 2026-05-27
Add/Refactor lind logging system (`lind_log!` / `lind_debug_panic!`)

### LINKED ISSUE #1178: add logging system to lind
we can have a better logging system in lind for better debug support

conceptually, we can have a `lind_log` macro which supports three modes: 1. print the log into stdout, 2. print log to stderr and more importantly, 3. print log to a log file

3rd mode is important since some cross-testing environment, like cpython is using a native test runner to execute lind program one by one, and any debugging log would break the test runner expectation.

one good place to put the loggings are after the wasm module crashed. This is currently handled as an exit code of 1 and sliently propogated to the upstream process. It would be helpful if the reason why the program is crashed is logged somewhere since a wasm crash is likely to be a bug. sliently propogating the exit code is more likely served as a temporary workaround to make application running without interruption

Another related issue that can be fixed together is refactoring `lind_debug_panic` system so that it also supports multiple modes: 1. panic and exit 2. just logging 3. do nothing

qianxichen233:
Closes https://github.com/Lind-Project/lind-wasm/issues/1178

Introduces a unified diagnostic logging system for Lind with two macros:

- `lind_log!` — general diagnostic output, never panics, category-filtered
- `lind_debug_panic!` — soft panic for unexpected-but-survivable conditions; panics in dev/CI, logs or ignores in deployment

Both macros are **compile-time gated** by the `lind-logging` Cargo feature (enabled by default in `make lind-boot`, removable with `NO_LOGGING=1` for benchmarks). Runtime behavior is configured via environment variables:

```bash
LIND_LOG_OUTPUT=stderr          # stdout | stderr | none | /path/to/file
LIND_LOG_CATEGORIES=dylink      # default | dylink | threei | all | comma-separated
LIND_DEBUG_PANIC=log-only       # panic-and-exit | log-only | no-action
```

Default config writes to `$LIND_WASM_ROOT/LIND.log` (or `/tmp/LIND.log`), panics on `lind_debug_panic!`, and emits only `Default`-category logs. The test runner is configured to use `panic-and-exit` + `none` output so unexpected conditions surface as test failures without log noise.

This PR also migrates all existing `println!`/`eprintln!`/`lind_debug_panic` call sites across `cage`, `rawposix`, `threei`, `lind-multi-process`, `lind-dylink`, and `wasmtime` to the new macros, and removes the old `debug-dylink` compile-time feature.

See `docs/internal/logging.md` for full usage, configuration reference, and architecture details.


Yaxuan-w (commented):
What's the difference between this and lind_debug?

qianxichen233:
> What's the difference between this and lind_debug?

It's basically a standardlized logging system, we shouldnt rely on vanilia `eprinln` or `println` for logging as the project scales. It unified our existing lind_debug_panic mechanism and a bunch of `eprintln` logging in our codebases, with more configurations enabled. All the normal logging should now go through `lind_log!` macro rather than vanila `epinrln`.

It also enabled configurable settings for `lind_log!` and `lind_debug_panic!`, where `lind_debug_panic!` supports three modes:
1. panic and exit directly (the old `lind_debug_panic` behavior)
2. only logging, do not do anything
3. completely ignore

So does `lind_log!`:
1. log to stderr
2. log to stdout
3. log to specified files

All of these configuration are controlled by environment variables. This allows quick toggle of lind logging behavior for different scenarios, examples:

**Development — rich debug/logging information for better debugging experience:**
```bash
LIND_LOG_OUTPUT=stderr LIND_DEBUG_PANIC=panic-and-exit ./build/lind-boot program.wasm
```

**Deployment — best-effort continuation on unexpected conditions:**
```bash
LIND_LOG_OUTPUT=/var/log/lind.log LIND_DEBUG_PANIC=log-only ./build/lind-boot program.wasm
```

**CI — silence log noise but keep unexpected-condition detection as test failures:**
```bash
LIND_LOG_OUTPUT=none LIND_DEBUG_PANIC=panic-and-exit ./build/lind-boot program.wasm
```

**Maximum performance / benchmark — remove all logging code from the binary:**
```bash
make lind-boot NO_LOGGING=1
```

The purpose of `lind_log!` and `lind_debug_panic!` are designed to match below workflow:
```
Is continued execution possible?
│
├─ No, always → panic!
│
└─ Yes, or maybe →
       Is this condition expected during normal operation?
       │
       ├─ Yes → lind_log!
       │
       └─ No — something that should never happen →
              lind_debug_panic!
```

Yaxuan-w (approved):

PR #1217, opened 2026-05-29
Fix address truncation

harindham:
# Fix 32-bit integer overflow and interval boundary bugs in vmmap and syscall handling
 
## Purpose
 
Fixes silent 32-bit integer overflow in `addr + len` calculations causing pages to wrap to low memory causing `mprotect`/`munmap` to silently affect adjacent mappings.
 
## Files Changed
 
- `fs_calls.rs` — overflow guards on address arithmetic in `mmap`, `munmap`, `brk`, and `shmat` syscall handlers
- `vmmap.rs` — overflow check added in arithmetic throughout page range calculations. It might prevent the wrap around by clamping to the maximum within the range or it can throw an error where clamping to the max value is not ideal

### DIFF
--- a/skip_test_cases.txt
+++ b/skip_test_cases.txt
@@ -15,5 +15,4 @@ dylink_tests/deterministic/rdynamic_main.c
 dylink_tests/deterministic/rdynamic_lib.c
 process_tests/deterministic/fork_max_cages.c
 ci/deterministic/ci_intentional_failure_tmp.c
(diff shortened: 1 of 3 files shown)


Yaxuan-w (commented):
Need to pass linter. Run `make format` on the root of the repo. 

I'm not sure what's happening on the test case. Seems lind past but native failed?

Yaxuan-w:
Seems duplicated with #1234, so closed

PR #1239, opened 2026-06-06
umask syscall implementation

### LINKED ISSUE #1219: Coreutils test failure : umask() syscall is a stub
## Coreutils tests affected :
`chmod/equal-x, chmod/equals, chmod/usage, mkdir/parents, mkdir/perm,
  mkdir/special-1`

## Issue
umask() always returns -1 instead of storing/returning a real mask, so any chmod symbolic mode without a who part (e.g. =x, =u, -w) clears every bit and mkdir's resulting permissions are wrong

moogchi:
# Description:

## Problem:

umask syscall was just a stub that returned -1, which caused the coreutils test failure
The failed tests were:
- chmod/equal-x, chmod/equals, chmod/usage — symbolic chmod modes without a who part (e.g. =x) consult the umask to determine which permission bits to set; with umask broken these resolved incorrectly
- mkdir/parents, mkdir/perm, mkdir/special-1 — directory permissions came out wrong because the mask was never applied

## What changed:
- Added a per-cage umask field (AtomicU32, default 0o022) to the Cage struct
- Implemented umask_syscall in fs_calls.rs — atomically swaps the new mask in and returns the old one
- Applied the umask at file/directory creation time in open_syscall, openat_syscall, mkdir_syscall, and mknod_syscall via mode & !umask
- fork correctly inherits the parent's umask (POSIX-compliant)
- Updated the glibc umask.c stub to call into the Rust syscall via MAKE_LEGACY_SYSCALL
- Added a deterministic test covering return value, open, mkdir, and umask changes

## Fixes
#1219


src/glibc/sysdeps/unix/sysv/linux/umask.c
```diff
@@ -1,9 +1,14 @@
...
 
 mode_t
 umask (mode_t mask)
 {
-  return -1;
-}
+  return (mode_t) MAKE_LEGACY_SYSCALL (
+      UMASK_SYSCALL, "syscall|umask", (uint64_t) mask,
+      NOTUSED, NOTUSED, NOTUSED, NOTUSED, NOTUSED, TRANSLATE_ERRNO_ON);
+}
```
Yaxuan-w: Add new line

src/glibc/sysdeps/unix/sysv/linux/umask.c
```diff
@@ -1,9 +1,14 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <string.h>
+#include <sys/stat.h>
```
Yaxuan-w: Seems this file is not used. Should we remove?

src/rawposix/src/init.rs
```diff
@@ -287,6 +287,7 @@ pub fn rawposix_start(verbosity: isize) {
         exit_group_initiated: AtomicBool::new(false),
         is_dead: AtomicBool::new(false),
         grate_inflight: AtomicU64::new(0),
+        umask: AtomicU32::new(0o022),
```
Yaxuan-w: Can we have comments about default value here?

tests/unit-tests/file_tests/deterministic/umask.c
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if ((st.st_mode & 0777) != 0600) {
+        printf("FAIL: expected file perms 0600, got %04o\n", st.st_mode & 0777);
+        return 1;
+    }
+    printf("PASS: umask 0077 applied correctly on open (0600)\n");
+
+    printf("All umask tests passed.\n");
+    return 0;
+}
```
Yaxuan-w: Add new line

Yaxuan-w (changes requested):
left minor comments

Yaxuan-w:
As today's discussion, can you try turn this into a grate?

vidyalakshmir:
@Yaxuan-w @moogchi 
What is the decision on this? Are we waiting the grates implementation? If so, do we still keep this Rawposix implementation as a fall back mechanism?

Yaxuan-w:
> What is the decision on this? Are we waiting the grates implementation? If so, do we still keep this Rawposix implementation as a fall back mechanism?

Yeah, @moogchi is working on the grate implementation. Based on our previous discussion on the meeting, we'd prefer keeping grate first unless we found other reasons (ie: high performance overhead / etc.). Thanks for reminding, going to close this pr.

PR #1242, opened 2026-06-08
fix dynamically built grate

qianxichen233:
fix dynamically built grate so that grate can be dynamically built and run

also adjust the default stack space for grate worker to be 1MB rather than 8MB

### DIFF
--- a/src/sysdefs/src/constants/lind_platform_const.rs
+++ b/src/sysdefs/src/constants/lind_platform_const.rs
@@ -135,14 +135,45 @@ pub const FPCAST_FUNC_SIGNATURE: &str = "$fpcast_emu$";
 /// instance reserves the same number of worker stack slots in linear memory.
 pub const MAX_GRATE_WORKERS: usize = 32;
 
-/// Size in bytes of the usable stack region assigned to one grate worker.
+/// Default size in bytes of the usable stack region assigned to one grate worker.
 ///
 /// Each worker executes in its own `Store + Instance` context, but workers may
 /// still attach to the same underlying linear memory. Therefore, every worker
-/// must be given a disjoint stack slot inside the shared stack arena.
+/// must be given a disjoint stack slot inside the shared stack arena. This is the
+/// usable portion of that per-worker slot.
 ///
-/// This constant specifies the usable portion of that per-worker slot.
-pub const GRATE_STACK_SLOT_SIZE: u32 = 8 * 1024 * 1024;
+/// Grate workers only ever run grate *handler* code (syscall interposition), which
+/// is shallow compared to arbitrary user programs, so the default is modest (1 MiB).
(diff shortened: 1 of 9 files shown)


src/sysdefs/src/constants/lind_platform_const.rs
```diff
@@ -135,14 +135,45 @@ pub const FPCAST_FUNC_SIGNATURE: &str = "$fpcast_emu$";
...
+    *GRATE_STACK_SLOT_SIZE_CACHED.get_or_init(|| {
+        std::env::var(GRATE_STACK_SIZE_ENV)
+            .ok()
+            .and_then(|v| v.trim().parse::<u32>().ok())
+            .filter(|n| *n > 0)
+            // Keep slot boundaries page-aligned so the arena layout stays well-formed.
+            .map(|n| (n + 4095) & !4095)
+            .unwrap_or(DEFAULT_GRATE_STACK_SLOT_SIZE)
+    })
+}
```
Yaxuan-w, line 176: `LIND_GRATE_STACK_SIZE` page rounding can overflow and produce a zero/small stack slot

src/sysdefs/src/constants/lind_platform_const.rs
```diff
@@ -135,14 +135,45 @@ pub const FPCAST_FUNC_SIGNATURE: &str = "$fpcast_emu$";
...
+/// usable portion of that per-worker slot.
+///
+/// Grate workers only ever run grate *handler* code (syscall interposition), which
+/// is shallow compared to arbitrary user programs, so the default is modest (1 MiB).
+/// Override at runtime with the `LIND_GRATE_STACK_SIZE` environment variable; see
+/// [`grate_stack_slot_size`].
+pub const DEFAULT_GRATE_STACK_SLOT_SIZE: u32 = 1024 * 1024;
+
+/// Environment variable that overrides the per-worker grate stack slot size (bytes).
+pub const GRATE_STACK_SIZE_ENV: &str = "LIND_GRATE_STACK_SIZE";
```
Yaxuan-w, line 152: Not sure if making this configurable through CLI makes more sense

Yaxuan-w (commented):
Some design suggestions:

1. Use a stable runtime module identity instead of `module.name()`: The Wasm name section is optional and may be removed by stripping, precompilation, or toolchain differences. Introduce a Lind-level logical module ID and use it consistently for global snapshots, GOT state, and memory-base lookup.
2. Unify the dynamic-linking replay for threads and grate workers: `build_dylink_child_store` duplicates logic already present in the pthread creation path. Both paths should use one shared helper for rebuilding the linker, function table, GOT, preload modules, dlopen modules, TLS, and global snapshots. The callers should handle only thread/cage-specific bookkeeping.

Issue #1216, opened 2026-05-29
CI/CD Roadmap


Yaxuan-w:
Based on today’s discussion with @vidyalakshmir and @Uk-jake, here are the proposed near-term CI/CD tasks and decisions that need follow-up.

## 1. Consolidate lind-wasm Images

Currently we maintain two image variants. However, the E2E workflow rebuilds the environment even when using the release image.

We can remove redundant image maintenance and simplify the workflow.

## 2. Configure App Repository Testing to Use GCP

GitHub Actions has fairly restrictive disk limits, which can become a bottleneck for our testing workflows.

Since we have access to GCP credits, it may be beneficial to offload App repository testing workloads to GCP-backed runners/resources.

### Proposed action

* Investigate GitHub Actions + GCP integration for App repository testing.
* Confirm available GCP credits and expected usage budget with Justin.

### Question

* What is our available GCP budget and how much testing capacity can it realistically support?

## 3. Pilot GCP-Based Testing with Example Grate

Compared with the App repository, the Grates repository has lower disk requirements.

At the same time, both repositories share similar CI requirements:

* Pulling a lind-wasm image
* Running integration/testing workflows

Therefore, Example Grate may be a good candidate for an initial deployment experiment.

### Proposed action

* Start with Example Grate as a proof-of-concept.
* Evaluate GCP-based execution there first.

### Decision criteria

* If GCP quota/credits are sufficient
    * Deploy GCP-backed testing for Example Grate.
* If GCP quota/credits are limited
    * Keep Grates on standard GitHub Actions runners.
    * Continue using per-PR testing.

## 4. Determine App Repository Testing Frequency

We need to decide how frequently App repository testing should run.

Topics for discussion

* Per-PR only?
* Scheduled testing (daily/nightly)?
* Hybrid approach (per-PR + periodic full regression)?

vidyalakshmir:
Looks good!

Uk-jake:
Thank you for the summary. I'll start by working on the lind-wasm image consolidation and the Grate test pipeline. I'll keep you updated on any progress:)

Yaxuan-w:
Discussion moved to #1342

PR #1248, opened 2026-06-11
ci: refactor e2e, release workflow

Uk-jake:
### Description

Refactored `e2e.yml` and `release.yml` 

### Changes

1. `e2e.yml` — triggered on PR open/update

- Runs on every PR open and update
- Workflow: build → test → report

2. `release.yml` — triggered on push to main

- Runs when a PR is merged into main
- Builds the release image from the latest code and pushes it to Docker Hub

**Differences from the previous workflow**

- The release image was pushed on a weekly schedule.
- `e2e.yml` was triggered on both PR open/update and main push, but changes merged into main were never reflected in the release image at that point.

Related: #1216

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release on Docker Hub
 
 on:
   workflow_dispatch:
-  schedule:
+  push:
     branches: [main]
 
 # Serialize releases on the same ref.
@@ -12,8 +12,6 @@ concurrency:
   cancel-in-progress: false
 
 permissions: {}
-
-name: Test and release on Docker Hub
 
 jobs:
   # Release job: build the 'release' image and publish to Docker Hub.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,25 @@
-name: Test and release on Docker Hub
+name: Release on Docker Hub
 
 on:
   workflow_dispatch:
   schedule:
-  - cron: "0 8 * * 1" # Run at 08:00, every Monday
+    branches: [main]
+
+# Serialize releases on the same ref.
+# Do NOT cancel in-progress runs: a cancelled run could abort a push and leave Docker Hub in a partial state.
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
 
 permissions: {}
 
+name: Test and release on Docker Hub
+
 jobs:
-  docker:
+  # Release job: build the 'release' image and publish to Docker Hub.
+  release:
+    permissions:
+      contents: read
     runs-on: ubuntu-22.04
     environment: dockerhub
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,14 +1,25 @@
-name: Test and release on Docker Hub
+name: Release on Docker Hub
 
 on:
   workflow_dispatch:
   schedule:
-  - cron: "0 8 * * 1" # Run at 08:00, every Monday
+    branches: [main]
```
tashabits, line 6: should this be under a push trigger instead of schedule?

The PR description says this should run when changes land on main, but this currently puts `branches: [main]` under `schedule`. My understanding is that `schedule` needs a `cron` entry, while branch filters belong under `push`.

.github/workflows/release.yml
```diff
@@ -1,14 +1,25 @@
...
+
+# Serialize releases on the same ref.
+# Do NOT cancel in-progress runs: a cancelled run could abort a push and leave Docker Hub in a partial state.
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
 
 permissions: {}
 
+name: Test and release on Docker Hub
```
tashabits: was this second top-level name: left in accidentally?

There’s already a workflow name at the top of the file, so I think this duplicate should probably be removed.

tashabits (changes requested):

.github/workflows/release.yml
Uk-jake: My mistake. I updated with a new commit. Thanks!

Yaxuan-w (commented):
> 2. `release.yml` — triggered on push to main
> 
> - Runs when a PR is merged into main
> - Builds the release image from the latest code and pushes it to Docker Hub

What will happen if there's a second commit to main while the previous build is running? Will the previous build be canceled or it will wait for it to finish?

Uk-jake:
> > 2. `release.yml` — triggered on push to main
> > 
> > 
> > * Runs when a PR is merged into main
> > * Builds the release image from the latest code and pushes it to Docker Hub
> 
> What will happen if there's a second commit to main while the previous build is running? Will the previous build be canceled or it will wait for it to finish?

My understanding is that due to the code below, the second build will wait in the queue until the first build completes.

```
concurrency:
  group: release-${{ github.ref }}
  cancel-in-progress: false
```

Uk-jake:
@rennergade @Yaxuan-w @vidyalakshmir

Do we need to add the release image push cycle to weekly? 

It's configured to deploy on every commit to main because a weekly release would delay reflecting the latest updates immediately. I can configure the setup, whether we want a weekly release or a commit-based release on the main branch.(or both)

Yaxuan-w:
> Do we need to add the release image push cycle to weekly?


We only want release image once a new release version has been published

Uk-jake:
> We only want release image once a new release version has been published

Does a new release mean whenever there are changes to the main branch, or is there a specific criteria for creating a new release?

I think it would also be a good to decide on our tagging strategy based on that release criteria.

Yaxuan-w (approved):

vidyalakshmir (approved):

tashabits:
LGTM now, the two things I called out have been fixed

tashabits (approved):

PR #1268, opened 2026-06-21
Implement copy_file_range syscall

### LINKED ISSUE #1240: python test failure : copy_file_range not implemented
Implement `copy_file_range` system call. This is used in python tests.

DanielZ224:
## Purpose

This PR implements support for the Linux copy_file_range syscall in Lind-Wasm. The syscall was previously unavailable through Lind: the glibc-side copy_file_range() implementation returned ENOSYS, which caused Python-related tests that rely on file-copy behavior to fail.

The implementation wires copy_file_range through the full Lind syscall path:

```c
glibc copy_file_range()
  -> MAKE_LEGACY_SYSCALL(COPY_FILE_RANGE_SYSCALL)
  -> Lind syscall trampoline
  -> RawPOSIX syscall table
  -> copy_file_range_syscall()
  -> host SYS_copy_file_range
```

## Related Issue

Fixes #1240.

## Changes

* Added COPY_FILE_RANGE_SYSCALL with the Linux x86_64 syscall number 326.
* Added the syscall number to both:
    * src/sysdefs/src/constants/syscall_const.rs
    * src/glibc/lind_syscall/lind_syscall_num.h
* Replaced the existing glibc copy_file_range ENOSYS stub with a Lind MAKE_LEGACY_SYSCALL wrapper.
* Added copy_file_range_syscall in RawPOSIX.
* Registered copy_file_range_syscall in the RawPOSIX syscall dispatcher table.
* Preserved support for:
    * explicit input/output offset pointers
    * NULL offset pointers
    * errno translation through the existing Lind syscall wrapper path

## Testing

Tested locally with:

```c
cargo fmt --check --all --manifest-path src/wasmtime/Cargo.toml
make sysroot
make lind-boot
```

I also manually verified the syscall with small Lind-compiled C programs covering:

1. Basic file copy through copy_file_range.
2. Nonzero flags returning EINVAL.
3. Explicit off_in and off_out pointer behavior, including offset updates and copied content validation.

Observed outputs included:

```c
copy_file_range ok
copy_file_range flags test ok
copy_file_range offsets test ok
```

## Notes for Reviewers

The previous glibc implementation intentionally returned ENOSYS, so RawPOSIX was never reached when guest programs called copy_file_range(). This PR updates both sides of the syscall path: the glibc wrapper and the RawPOSIX dispatcher/handler.

The RawPOSIX implementation currently delegates to the host SYS_copy_file_range after translating Lind virtual file descriptors to host file descriptors. This matches the existing style used by nearby file syscalls such as pread, pwrite, and lseek, which similarly translate descriptors and then invoke the host libc/syscall layer.

### DIFF
--- a/src/glibc/lind_syscall/lind_syscall_num.h
+++ b/src/glibc/lind_syscall/lind_syscall_num.h
@@ -140,6 +140,7 @@
 #define PWRITEV_SYSCALL 296
 #define PRLIMIT64_SYSCALL 302
 #define GETRANDOM_SYSCALL 318
+#define COPY_FILE_RANGE_SYSCALL 326
 
 /* Lind-specific syscalls (not part of the Linux syscall table) */
 #define REGISTER_HANDLER_SYSCALL 1001
--- a/src/glibc/sysdeps/unix/sysv/linux/copy_file_range.c
+++ b/src/glibc/sysdeps/unix/sysv/linux/copy_file_range.c
@@ -18,16 +18,30 @@
 
 #include <errno.h>
 #include <unistd.h>
+#include <stdint.h>
+#include <syscall-template.h>
+#include <lind_syscall_num.h>
+#include <addr_translation.h>
 
 ssize_t
 copy_file_range (int infd, __off64_t *pinoff,
                  int outfd, __off64_t *poutoff,
                  size_t length, unsigned int flags)
 {
-  // BUG: we currently cannot support this syscall
-  //      so instead of letting it crash directly,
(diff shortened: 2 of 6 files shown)


Yaxuan-w (changes requested):
Can we also have a unit test for this syscall?

DanielZ224:
@Yaxuan-w Added a unit test for copy_file_range covering basic copy behavior, explicit offset updates, and the nonzero-flags error path.

Yaxuan-w (approved):

drapl0n (approved):

PR #1284, opened 2026-06-24
Update toolchain version in doc

Yaxuan-w:
Update toolchain version in doc


Uk-jake (approved):
Thank you for changing it.

Also, I think the "Docker Hub Release workflow" and "End-to-End Testing" in the docs might need some updates as well. I'll try to take a look and fix them

PR #1290, opened 2026-06-29
Fold E2E PR report sections by harness

Yaxuan-w:
Update the E2E PR comment renderer to fold each harness section from its `h2`.

This makes report comments consistent across harnesses instead of only folding sections that happen to contain `h3 + table` markup.


Uk-jake (approved):
it works well

PR #1295, opened 2026-07-02
docs: add Lind-Wasm threat model

rennergade:
## Summary

Adds a threat model document for Lind-Wasm to the internal docs, plus the actor-topology figure it references.

- `docs/internal/threat-model.md` — actors, trust relationships (TCB vs. trusted policy surface), attacker model, the attacker×victim matrix with per-cell reasoning, security properties (P1–P6), and out-of-scope items.
- `docs/images/doc-images/actor-topology.png` — Figure 1, the actor topology.
- Registered the page under **Internal Documentation** in `mkdocs.yml` nav so it's linked in the published site.

## Notes

- Image path follows the existing `docs/images/doc-images/` convention; the relative link resolves from `docs/internal/`.
- Content was drafted with assistance and is intended as a starting point for discussion/review — feedback on the trust model and matrix reasoning welcome.


JustinCappos (approved):
Most of this is for Lind, not specifically Lind-Wasm.  I don't know if it's worth making that clearer here.  I do know have a large cleanup project once we start to make that transition.

docs/internal/threat-model.md
```diff
@@ -0,0 +1,265 @@
...
+- Call any host import with chosen arguments, including `register_handler`,
+  `make_syscall`, and `copy_handler_table_to_cage`, subject to 3i's interposition
+  and scoping.
+- Call the syscall API with arbitrary inputs.
+- Rewrite its own handler table, try to copy in a different one, or try to drive
+  3i operations against peer or ancestor cages (scoping denies these).
+
+It cannot, by the Wasm memory model rather than any policy check, address host
+memory, issue raw Linux syscalls, alter JIT output, or reach another cage's
+memory.
```
F-Wilke, line 108: Does it make sense to include this: 
Wasm runtimes also prohibit access to virtual files within the /proc/self directory. This prevents cages from accessing host information such as secrets through the likes of /proc/self/mem.

docs/internal/threat-model.md
```diff
@@ -0,0 +1,265 @@
...
+**Case 4: the runtime as attacker (KO).** A memory-safety bug in Wasmtime, 3i, or
+RawPOSIX that an attacker rides into the runtime compromises every cage and grate
+(the three share fate, Section 4). Even then the attacker is just an unprivileged
+Linux process. Nothing Lind adds sits in the kernel or hardware, so the worst its
+own code can cost you is whatever an unprivileged process can do. That ceiling is
+the main resilience argument for the design.
+
+**Case 5: cage or grate against the runtime or kernel (U).** Neither has a path
+downward. The runtime's state is host memory it cannot address (Section 4), and
+the only way up is a runtime bug, which is Case 4. So these cells are unaffected.
```
F-Wilke, line 193: Further up you mention that a cage can cause a fault in RawPOSIX:

_Either it travels down tagged with its originating cage and RawPOSIX faults if it falls outside that cage's address space, or the grate pulls the bytes in with the
copy-from-cage call, which checks the same provenance before copying._

Does this mean a cage can cause the runtime to crash, or will this fault be contained?

My observation is that it kills the whole runtime (see attached code and log).

[faultlog.txt](https://github.com/user-attachments/files/29650895/faultlog.txt)
[hello.c](https://github.com/user-attachments/files/29650896/hello.c)

docs/internal/threat-model.md
```diff
@@ -0,0 +1,265 @@
...
+   forwarding.
+
+In between, **3i** routes a cage's syscalls through optional policy components
+called *grates* before they reach RawPOSIX. Affecting the host or another cage
+means defeating both layers.
+
+## 2. Methodology
+
+We do not sort components into trusted and untrusted, because that hides the fact
+that a grate is trusted by the cages it isolates and by nothing else. Instead we
```
vidyalakshmir, line 23: >> because that hides the fact that a grate is trusted by the cages it isolates and by nothing else

Probably a better way to say is, 


We do not sort components into trusted and untrusted; there is no global notion of trust. Instead, each component is trusted by a specific set of other components — a grate, for example, is trusted by the cages it isolates and by nothing else.

docs/internal/threat-model.md
Yaxuan-w: Yeah I have same confusion here. 

Section 5 says a compromised cage can call host imports and syscall APIs with arbitrary inputs, including make_syscall, register_handler, and copy_handler_table_to_cage. That is a direct input path into 3i/RawPOSIX/Wasmtime.

Also Fredrik's comment showing malformed pointer behavior can kill the whole runtime... I'm not sure should we make this more precise here..?

docs/internal/threat-model.md
```diff
@@ -0,0 +1,265 @@
...
+
+## 6. The matrix
+
+Columns are the **attacker**, rows the **victim**. Each cell is what the attacker
+can do to the victim.
+
+| Victim \ Attacker | Cage A | Grate B | Cage C | Grate D | Runtime | Kernel+HW |
+| --- | --- | --- | --- | --- | --- | --- |
+| **Cage A** | N/A | CC | RE | RE | KO | KO |
+| **Grate B** | FE | N/A | RE | RE | KO | KO |
```
Yaxuan-w, line 130: I think Grate B only has RE against unrelated Cage C / Grate D is not precise enough.. Actually a grate can copy from arbitrary cage IDs (from the compromised-grate section and our current implementation)

docs/internal/threat-model.md
```diff
@@ -0,0 +1,265 @@
...
+`make_syscall()` → 3i → grate stack → RawPOSIX → kernel, and RawPOSIX validates
+pointer arguments against the calling cage's vmmap before forwarding.
+
+**P3: directional confinement.** Syscalls run through the grate stack toward
+RawPOSIX, and a cage cannot reach past, around, or below the grates that mediate
+it.
+
+**P4: handler-table integrity.** The 3i logic and per-cage handler tables are
+host memory, unreachable from any cage (Section 4), so routing cannot be tampered
+with even if the 3i policy logic has a bug. A cage influences routing only
```
Yaxuan-w, line 216: > routing cannot be tampered
with even if the 3i policy logic has a bug

This claim might be too strong. The routing correctness needs to depend on correct 3i scoping checks

docs/internal/threat-model.md
Yaxuan-w: > Wasm runtimes also prohibit access to virtual files within the /proc/self directory

This feature is provided by wasi API, which is not used in our lind-wasm implementation. lind-wasm provides fs isolation by enforce chroot at initialization -- only gives cage/grates access to `lindfs/` -- but doesn't prohibit access to /proc/self inside lindfs.

docs/internal/threat-model.md
```diff
@@ -0,0 +1,265 @@
...
+It cannot, by the Wasm memory model rather than any policy check, address host
+memory, issue raw Linux syscalls, alter JIT output, or reach another cage's
+memory.
+
+Out of the attacker's control: the kernel and hardware, the Wasmtime JIT and its
+output, the 3i logic and handler tables, RawPOSIX state, and other cages' memory.
+
+### Compromised grate
+
+A grate that is compromised can observe, change, drop, or fake results for any
```
vidyalakshmir, line 115: In earlier section, it is mentioned
`A  grate is trusted only by its own descendant cages, and only to enforce their policy.`

Does having a grate compromised affect the trust relationship in anyway?

Or is trust and a component getting compromised not related?

vidyalakshmir:
>> Most of this is for Lind, not specifically Lind-Wasm. I don't know if it's worth making that clearer here. I do know have a large cleanup project once we start to make that transition.

I think wasm related is specifically the P6 in Section 8 - Security Properties. In general, any SFI can guarantee cross-cage memory isolation etc.

vidyalakshmir:
One thing not discussed is about kernel attack surface reduction or if attacks targeting kernel is made harder through Lind design. Some thoughts on this.. 

There are two cases here : 1) In a conventional Linux environment, an attacker who finds a memory-safety bug in the application — a buffer overflow, a use-after-free, a format string vulnerability — can exploit it to gain controlled execution within the process and then immediately leverage that foothold to issue arbitrary syscalls with attacker-chosen arguments, targeting known vulnerabilities in the kernel's syscall handlers. 2) Even without fully compromising the application's control flow, an attacker who can influence what data the application processes — through a malicious file, a crafted network packet, or attacker-controlled input that flows into a syscall argument — can steer the application into making a syscall that triggers a kernel bug without the application itself being "exploited" in the traditional sense. In both cases, the kernel's attack surface is the full Linux syscall table, reachable with arbitrary argument values, and a single kernel vulnerability reachable through this path can result in privilege escalation, arbitrary memory corruption across all processes, or a full system compromise.

In Lind, both attack paths are disrupted by the same structural property: a cage's only route to the OS is through lind_syscall → 3i → RawPOSIX → kernel. Even if an attacker fully exploits an application cage and achieves arbitrary Wasm code execution, they cannot issue raw kernel syscalls directly — every syscall is intercepted by RawPOSIX, which validates all pointer arguments against the cage's vmmap, enforces unconditional policy rules regardless of argument values, and forwards to the kernel only the subset of operations it deems well-formed and cage-appropriate. A syscall argument deliberately crafted to trigger a specific kernel bug must pass through this validation layer first. Critically, this defence holds even when the application cage is fully compromised — because the interposition is not inside the application but in the trusted host code the cage cannot modify or bypass.

docs/internal/threat-model.md
rennergade: This seems like a bug to me yes? We shouldn't be able to kill the runtime like this.

rennergade:
> One thing not discussed is about kernel attack surface reduction or if attacks targeting kernel is made harder through Lind design. Some thoughts on this..
> 
> There are two cases here : 1) In a conventional Linux environment, an attacker who finds a memory-safety bug in the application — a buffer overflow, a use-after-free, a format string vulnerability — can exploit it to gain controlled execution within the process and then immediately leverage that foothold to issue arbitrary syscalls with attacker-chosen arguments, targeting known vulnerabilities in the kernel's syscall handlers. 2) Even without fully compromising the application's control flow, an attacker who can influence what data the application processes — through a malicious file, a crafted network packet, or attacker-controlled input that flows into a syscall argument — can steer the application into making a syscall that triggers a kernel bug without the application itself being "exploited" in the traditional sense. In both cases, the kernel's attack surface is the full Linux syscall table, reachable with arbitrary argument values, and a single kernel vulnerability reachable through this path can result in privilege escalation, arbitrary memory corruption across all processes, or a full system compromise.
> 
> In Lind, both attack paths are disrupted by the same structural property: a cage's only route to the OS is through lind_syscall → 3i → RawPOSIX → kernel. Even if an attacker fully exploits an application cage and achieves arbitrary Wasm code execution, they cannot issue raw kernel syscalls directly — every syscall is intercepted by RawPOSIX, which validates all pointer arguments against the cage's vmmap, enforces unconditional policy rules regardless of argument values, and forwards to the kernel only the subset of operations it deems well-formed and cage-appropriate. A syscall argument deliberately crafted to trigger a specific kernel bug must pass through this validation layer first. Critically, this defence holds even when the application cage is fully compromised — because the interposition is not inside the application but in the trusted host code the cage cannot modify or bypass.

I do think some of this has merit, I'll try to integrate some here. What I think we want to avoid is a "popular paths" style argument in surface reduction, since at this point its really just "we didnt implement these syscalls", even though something like that could be done with a grate.

docs/internal/threat-model.md
qianxichen233: a cage should be designed to not crash host, if that happens, this is a bug that should be fixed immediately. We shouldn't rely on current lind implementation to define/test threat model as it could have bugs, threat model is a more high level design stuff where the implementation should follows

Issue #1250, opened 2026-06-12
Rewrite RawPOSIX internal docs for clarity and current repo layout


rennergade:
## Problem

The current RawPOSIX internal docs read stale and overstate the component's guarantees.

Specific issues:
- `docs/internal/rawposix.md` describes RawPOSIX as fully "POSIX-compliant" and says it should behave the same as native Linux. That is stronger than the rest of the docs and likely stronger than the implementation guarantees.
- The opening text is hard to parse, e.g. "provide an in-process OS while isolating them."
- The page describes a standalone RawPOSIX repository layout (`src/`, `syscalls.rs`, `cage.rs`, `tests/`, `docs/`) instead of orienting readers around the current `lind-wasm/src/rawposix` tree.
- The page does not clearly explain how RawPOSIX relates to cages, 3i, Wasmtime, and host Linux syscalls in the current architecture.

## Desired outcome

Rewrite `docs/internal/rawposix.md` so it is accurate, current, and useful to contributors.

Suggested structure:
- What RawPOSIX is in Lind-Wasm
- What it is responsible for
- What it is not responsible for
- How syscalls flow from lind-glibc / 3i into RawPOSIX and then to host Linux
- How per-cage state is represented and isolated
- Where the relevant source files live in the current repo
- How RawPOSIX is tested

## Notes

Avoid overclaiming exact POSIX compatibility. Prefer wording like "POSIX-style", "POSIX-compatible where implemented", or "aims to preserve POSIX behavior for supported calls" unless a stronger guarantee is intentionally documented.

rishabhBudhouliya:
@Yaxuan-w Can I work on this, at least for an initial draft which can be reviewed and refined?

Yaxuan-w:
> Can I work on this, at least for an initial draft which can be reviewed and refined?

Yeah

PR #1296, opened 2026-07-03
rewrite rawposix.md so it matches the current codebase and architecture

rishabhBudhouliya:
# Purpose
Rewrite docs/internal/rawposix.md so it matches the current codebase and architecture.

Addresses https://github.com/Lind-Project/lind-wasm/issues/1250

# What changed
- Compatibility wording is now calibrated: "aims to preserve POSIX behavior for the calls it implements", matching the framing already used in docs/index.md, with syscall_table.rs named as the authoritative list of supported calls.
- Added explicit "responsible for" and "not responsible for" sections covering the boundaries with Wasmtime/lind-boot, 3i, grates, and lind-glibc.
- Added a "life of a syscall" section tracing read() from lind-glibc through the lind::make-syscall import, the lind-common host function, and 3i dispatch into fs_calls.rs and host libc, with file references for each hop.
- Documented per-cage state as it exists today: the Cage struct in src/cage, fd virtualization in src/fdtables, and the RawCallFunc handler ABI with its per-argument cage IDs and convert-early convention.
- Source layout section now shows the real src/rawposix/src tree and the sibling crates it depends on (cage, fdtables, threei, typemap, sysdefs).
- Testing section points at the actual harness (scripts/test/harnesses/wasmtestreport.py over tests/unit-tests/) instead of the nonexistent src/RawPOSIX/src/tests/ path.

### DIFF
--- a/docs/internal/rawposix.md
+++ b/docs/internal/rawposix.md
@@ -1,45 +1,91 @@
----
-id: RawPOSIX
----
+# RawPOSIX
 
-# Introduction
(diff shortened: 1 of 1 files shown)


JustinCappos (approved):
Really nice documentation refresh here!

Yaxuan-w (approved):

PR #1297, opened 2026-07-05
Fix setjmp/longjmp table matching to survive raw jmp_buf save/restore

qianxichen233:
buf[0] previously stored the jmp_buf's own fixed address as the key testSetjmp matches against. For a jmp_buf that lives in a long-lived buffer and gets saved/restored as raw bytes by an unwind-protect-style helper (memcpy(&saved, &buf, sizeof(buf)) and back), the address never changes, so every registration on that buffer was indistinguishable. A caller that restores an older snapshot before re-throwing therefore kept matching its own already-fired registration, causing an unbounded catch/rethrow loop that silently exhausted the null-collected GC heap.

Replace the address-based key with a fresh per-call monotonic token stored in buf[0], and stop clearing matched table entries (which broke legitimate repeated longjmps to a still-live buffer with no intervening setjmp). Tokens make raw byte restoration meaningful again: an outer, still-live registration's token correctly falls through the inner frame's table and resolves in the outer frame's own table, matching native register/stack snapshot restoration semantics.


Yaxuan-w:
Can you trigger another ci run?

src/glibc/setjmp/wasm_eh_setjmp.c
```diff
@@ -45,22 +62,24 @@ __attribute__((visibility("default"))) int getTempRet0(void)
...
+ * Communicates the (possibly new) table size via setTempRet0().
  *
  * The initial table is allocated by the caller (malloc(40), size=4).
  */
 __attribute__((visibility("default"))) int *saveSetjmp(int *buf, int label_id, int *table, int size)
 {
-    /* Store self-reference so testSetjmp can match on buf[0]. */
-    buf[0] = (int)(uintptr_t)buf;
+    int token = ++g_setjmp_token;
+    if (token == 0) token = ++g_setjmp_token; /* skip the 0 sentinel on wraparound */
```
Yaxuan-w: In C, signed integer overflow is UB, so the program has already entered UB before we get to the if (token == 0) sentinel check. Maybe we can consider `uint32_t` in this case?

Yaxuan-w (commented):
Is that possible to have a unit test case for this?

src/glibc/setjmp/wasm_eh_setjmp.c
qianxichen233: theoritically this should never happen, gc heap overflow would happen before token overflow, as roughly ten million setjmp invocation would exhaust gc heap with even 4GB size (maximum possible size), an overflow of token would take 2 billion invocation

Yaxuan-w: Then it would be helpful to include these in the comments because the code explicitly handles this scenario

rishabhBudhouliya:
@qianxichen233 @Yaxuan-w I can confirm that bash exits cleanly after building lind-wasm with this change: 
```
lind@580bfae3d29e:~/lind-wasm$ lind-wasm --enable-fpcast bin/bash -c 'trap "echo CLEANUP" EXIT; exit 0'; echo "exit=$?"
  CLEANUP
  exit=0
lind@580bfae3d29e:~/lind-wasm$ lind-wasm --enable-fpcast bin/bash -c 'trap "exit 0" EXIT; exit 0'; echo "exit=$?"
  exit=0
```
I also ran the coreutils test with the setjmp flag modifications and it seems to be running successfully now. I still don't have the entire numbers in but most of them seem to be passing.

Yaxuan-w (approved):

PR #1305, opened 2026-07-07
fix dockerfile.dev - update version of libtinfo5

Uk-jake:
Cause
Ubuntu released a security update on security.ubuntu.com that replaced the libtinfo5_6.3-2ubuntu0.1_amd64.deb package. Dockerfile.dev was still pulling the removed version, causing the pipeline to fail.

Fix
Updated Dockerfile.dev to download the updated package version.

### DIFF
--- a/Docker/Dockerfile.dev
+++ b/Docker/Dockerfile.dev
@@ -76,9 +76,9 @@ RUN groupadd --gid 1000 ${USERNAME} && \
     echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
 
 # --- Install libtinfo5 as root ---
-RUN wget http://security.ubuntu.com/ubuntu/pool/universe/n/ncurses/libtinfo5_6.3-2ubuntu0.1_amd64.deb && \
-    apt install ./libtinfo5_6.3-2ubuntu0.1_amd64.deb
-
+RUN wget http://security.ubuntu.com/ubuntu/pool/universe/n/ncurses/libtinfo5_6.3-2ubuntu0.2_amd64.deb && \
+    apt install ./libtinfo5_6.3-2ubuntu0.2_amd64.deb
+    
 # --- Switch to user ASAP ---
 USER ${USERNAME}
 WORKDIR /home/${USERNAME}


Docker/Dockerfile.dev
```diff
@@ -76,9 +76,9 @@ RUN groupadd --gid 1000 ${USERNAME} && \
     echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
 
 # --- Install libtinfo5 as root ---
-RUN wget http://security.ubuntu.com/ubuntu/pool/universe/n/ncurses/libtinfo5_6.3-2ubuntu0.1_amd64.deb && \
-    apt install ./libtinfo5_6.3-2ubuntu0.1_amd64.deb
-
+RUN wget http://security.ubuntu.com/ubuntu/pool/universe/n/ncurses/libtinfo5_6.3-2ubuntu0.2_amd64.deb && \
+    apt install ./libtinfo5_6.3-2ubuntu0.2_amd64.deb
```
tashabits, line 80: This change fixes the current build failure by updating the `libtinfo5` .deb URL from `6.3-2ubuntu0.1` to `6.3-2ubuntu0.2`, but it preserves the same brittle pattern. Ubuntu removed the old package file from `security.ubuntu.com`, which is why the previous URL now returns 404. The same thing can happen again the next time Ubuntu publishes a security update and supersedes `6.3-2ubuntu0.2`.

Uk-jake: I agree. For now, I modified the code to quickly resolve the issue. I found that a fast fix was necessary because the dev image hasn't been updated for two weeks due to this problem. I will create a new issue to address this. If you have any ideas for a more fundamental solution, could you please share them?

Yaxuan-w: If we decide to use current solution, please also document in the comments of this file

Uk-jake: @Yaxuan-w 
This issue is blocking dev image updates. Is this preventing you from working because the dev image is outdated?  I was wondering if resolving this issue is urgent.

Yaxuan-w: Not actually blocking but cause some inconvenience when using containers.. I'd prefer merge a quick fix for now and move the discussion to the issue

Uk-jake: I opened an issue regarding this issue. I'd appreciate it if you could approve the current PR.

https://github.com/Lind-Project/lind-wasm/issues/1310

Yaxuan-w (approved):

PR #1318, opened 2026-07-16
Add cross-cage file descriptor brute-force test

lionne99:
Adds tests for ISO-002 cross-cage file descriptor isolation.

Changes:
- Add cross_cage_fd_bruteforce.c general unit test to verify that a cage cannot access another cage's file descriptors by bruteforcing fd numbers. All invalid accesses return `EBADF`.
- Add cross_cage_fd_isolation fdtables inner test in lib.rs to verify fd isolation between independent cages, including separate virtual fd mappings and rejecting translation of fds opened by another cage.


src/fdtables/src/lib.rs
```diff
@@ -239,6 +239,40 @@ mod tests {
         );
     }
 
+    #[test]
+    // ISO-002 fd-lib inner test: two cages have independent fd tables, so one
```
Yaxuan-w, line 243: > Cross-cage fd: cage A guesses/brute-forces fd numbers to touch cage B's descriptors

maybe this means trying getting B's fd by A's cageid

tests/unit-tests/file_tests/deterministic/cross_cage_fd_bruteforce.c
Yaxuan-w, line 1: Same on this test. Fork two cages and get B's fd by A's id. This should return EBADF.

Yaxuan-w (changes requested):
Good start! Left some clarifications on the test purpose.

tests/unit-tests/file_tests/deterministic/cross_cage_fd_bruteforce.c
```diff
@@ -0,0 +1,70 @@
...
+        assert(errno == EBADF);
+
+        assert(write(to_child[1], "x", 1) == 1);
+
+        close(to_parent[0]);
+        close(to_child[1]);
+
+        int status;
+        assert(waitpid(pid, &status, 0) == pid);
+        assert(WIFEXITED(status) && WEXITSTATUS(status) == 0);
```
Yaxuan-w: Unreachable parent logic

Yaxuan-w: Please move the parent logic into an else branch and keep the waitpid() there so child failures are propagated.

tests/unit-tests/file_tests/deterministic/cross_cage_fd_bruteforce.c
```diff
@@ -0,0 +1,70 @@
...
+    }
+    
+
+    printf("cross_cage_fd_bruteforce test: PASS\n");
+    return 0;
+}
```
Yaxuan-w: Add new line

tests/unit-tests/file_tests/deterministic/cross_cage_fd_bruteforce.c
```diff
@@ -0,0 +1,70 @@
...
+#include <unistd.h>
+#include <sys/wait.h>
+
+
+int main() {
+    for(int i = 3; i < 1024; i++) {
+        int r = fcntl(i, F_GETFD);
+        assert(r == -1);
+        assert(errno == EBADF);
+    }
```
Yaxuan-w: This scan runs before the fork and before cage B creates any fd, so it only checks that the initial process has no open fds in that range. It does not exercise cross-cage brute forcing. The scan should happen in the parent after the child has created its B-only descriptors, while excluding descriptors legitimately inherited or owned by the parent.

For example:
1. create cage A and cage B
2. cage B creates some fds only belongs to B, ie: 500..509
3. B notifies A that creation is done
4. A tries fcntl/read/write/close these fd numbers
5. All attempts in A **must** return EBADF (need to verify the errno)
6. B verifies that these fds are still valid (usable)

src/fdtables/src/lib.rs
```diff
@@ -239,6 +239,44 @@ mod tests {
...
+        assert_eq!(entry_b.underfd, 222);
+        assert_ne!(entry_a.underfd, entry_b.underfd);
+
+        let fd_a2 = get_unused_virtual_fd(cage_a, 0, 333, false, 0).unwrap();
+        assert_eq!(translate_virtual_fd(cage_a, fd_a2).unwrap().underfd, 333);
+        assert!(translate_virtual_fd(cage_b, fd_a2).is_err());
+
+        const B_ONLY_FD: u64 = 900;
+        get_specific_virtual_fd(cage_b, B_ONLY_FD, 0, 555, false, 0).unwrap();
+        get_specific_virtual_fd(cage_b, B_ONLY_FD, 0, 555, false, 0).unwrap();
```
Yaxuan-w: duplicate lines?

src/fdtables/src/lib.rs
```diff
@@ -239,6 +239,44 @@ mod tests {
...
+
+        assert!(translate_virtual_fd(cage_a, B_ONLY_FD).is_err());
```
Yaxuan-w: Also need to verify the errno==EBADF

Yaxuan-w (commented):
Thanks for the revision. This is moving in the right direction! The overall test design is close, and I’ve left a few comments on the remaining issues. Please let me know if anything is unclear or if you’d like to discuss the expected test flow.

tests/unit-tests/file_tests/deterministic/cross_cage_fd_bruteforce.c
Yaxuan-w: This is good to merge after resolving this comment

Yaxuan-w (approved):

PR #1331, opened 2026-07-24
Pin nightly version to 2026-07-16

Yaxuan-w:
## Summary

Yesterday's newest rust nightly version breaks lind-boot build. Pin the nightly version to the latest working version instead.

## Related issues

None

## Testing

`make test` /  `make lind-boot`

## Compatibility and security impact

None

## Checklist

- [x] I have read the
      [contribution guidelines](https://github.com/Lind-Project/lind-wasm/blob/main/CONTRIBUTING.md).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -19,8 +19,7 @@ security implications, or performance impact. Write "None" when not applicable.
 
 ## Checklist
 
-- [ ] I have read the
-      [contribution guidelines](https://github.com/Lind-Project/lind-wasm/blob/main/CONTRIBUTING.md).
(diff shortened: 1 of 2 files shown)


Uk-jake (approved):
Is there a specific reason need to use nightly instead of stable(just curiosity)?

I don't know exactly where Rust is used in lind-wasm. I think pinning a nightly version is a temporary solution. Switching to stable might actually solve this issue

Yaxuan-w:
> Is there a specific reason need to use nightly instead of stable(just curiosity)?

Some features used in lind-wasm only exist in nightly 

> I don't know exactly where Rust is used in lind-wasm. I think pinning a nightly version is a temporary solution. Switching to stable might actually solve this issue

Almost all lind-wasm runtime is written in rust. But yeah, we should check if there's a more general solution for this

PR #1333, opened 2026-07-27
Add ISO-003 cross-cage vmmap escape tests

lionne99:
Adds ISO-003 tests to verify that invalid memory ranges cannot escape the cage memory boundaries.

Changes
Added cross_cage_vmmap_escape.c unit test:
- Verifies that writing from a fully unmapped region returns EFAULT
- Verifies that writing across a mapped/unmapped page boundary is rejected
- Verifies that extremely large lengths that could overflow range calculations are rejected
- Uses a pipe canary check to ensure failed writes do not partially modify the destination

Added vmmap unit tests:
- Added coverage for rejecting address range calculations with overflowing lengths
- Added coverage for calculate_page_range returning None when the requested range overflows

### DIFF
--- a/src/cage/src/memory/vmmap.rs
+++ b/src/cage/src/memory/vmmap.rs
@@ -2503,4 +2503,48 @@ mod tests {
         let result = vmmap.calculate_page_range(0, 1);
         assert_eq!(result, Some((0, 1)), "1 byte should still be 1 page");
     }
+    #[test]
+    // ISO-003: a length large enough to overflow the range-arithmetic must be rejected
+    fn test_check_addr_write_negative_length_overflow_rejected() {
+        let mut vmmap = Vmmap::new();
+        vmmap.set_base_address(0);
+        vmmap.start_address = 0;
+        vmmap.end_address = 1000;
+
+        vmmap
+            .add_entry_with_overwrite(
+                10,
+                10,
+                PROT_READ | PROT_WRITE,
+                PROT_READ | PROT_WRITE,
+                0,
+                MemoryBackingType::Anonymous,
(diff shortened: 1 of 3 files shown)


src/cage/src/memory/vmmap.rs
```diff
@@ -2503,4 +2503,48 @@ mod tests {
         let result = vmmap.calculate_page_range(0, 1);
         assert_eq!(result, Some((0, 1)), "1 byte should still be 1 page");
     }
+    #[test]
+    // ISO-003: a length large enough to overflow the range-arithmetic must be rejected
+    fn test_check_addr_write_negative_length_overflow_rejected() {
+        let mut vmmap = Vmmap::new();
+        vmmap.set_base_address(0);
```
Yaxuan-w: Can we use `test_vmmap()` here?

tests/unit-tests/memory_tests/deterministic/vmmap_escape.c
Yaxuan-w, line 1: Seems we only have one cage in this test, but naming as "cross_cage*". Need to change to a more precise name

Yaxuan-w (changes requested):
Thanks for the PR! The tests are well-designed overall. Few things:

1. Worth also testing `read(fd, bad_buf, len) = host` (writing into cage memory).
2. Worth documenting in a comment: Test 2/3 assume native Linux returns a full EFAULT (rather than a POSIX-permitted partial write) for a straddling buffer on a pipe. Not all kernel has same behavior on this case

Yaxuan-w (approved):

PR #1334, opened 2026-07-28
feat: add linkat

### LINKED ISSUE #1332: feat: linkat not working
### Before submitting

- [x] I searched existing issues and discussions for similar proposals.

### Problem or use case

[linkat in glibc](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/sysdeps/unix/sysv/linux/i386/i686/linkat.c) is not using make_legacy_syscall which will always return success. 

### Proposed solution

Need to update with our API and also add test file

### Alternatives considered

_No response_

### Additional context

_No response_

Yaxuan-w:
## Summary

[linkat in glibc](https://github.com/Lind-Project/lind-wasm/blob/main/src/glibc/sysdeps/unix/sysv/linux/i386/i686/linkat.c) is not using make_legacy_syscall which will always return success. 

Update linkat in glibc to use make_legacy_syscall

## Related issues

Resolves #1332 

## Testing

New test case added in this PR

## Checklist


qianxichen233 (commented):
it probably makes sense to have the original `link` syscall directly calls the new `linkat` syscall for two reasons:
1. it reduces the number of trusted code
2. I think linux kernel also have an internal function that's called by both `link` and `linkat`

DanielZ224 (commented):
Thanks for implementing the missing `linkat` syscall path and adding regression tests. The syscall-number plumbing, glibc wrapper, syscall-table registration, and the basic success/error cases look reasonable. The tests also verify the resulting filesystem state rather than only checking the return value, which is important for the original issue.

However, I have two blocking concerns about the non-`AT_FDCWD` path handling.

## 1. Raw guest paths are passed directly to the host kernel

For a non-`AT_FDCWD` dirfd, the implementation translates the virtual fd to an underlying host fd, but then passes the raw guest pathname directly to `libc::linkat()`.

This appears to bypass Lind's sandbox-aware path translation and confinement logic. In particular:

- An absolute guest path would be resolved from the host root because the host kernel ignores `dirfd` for absolute paths.
- A relative path containing `..` may escape above the directory represented by the underlying fd.
- The `oldpath_cageid` / `newpath_cageid` values do not appear to be used in these branches.

Could this reuse an existing sandbox-aware `*at` path-resolution helper, or introduce a shared helper that resolves a guest path relative to a virtual dirfd without exposing the raw host path?

The same concern applies to both `oldpath` and `newpath`.

## 2. Absolute paths should ignore the corresponding dirfd

`linkat()` should only validate and use a dirfd when the associated pathname is relative. When the pathname is absolute, the corresponding dirfd must be ignored.

The current implementation appears to translate and validate the dirfd before determining whether the pathname is absolute. As a result, a call such as:

```c
linkat(-1, "/absolute/source", -1, "/absolute/destination", 0);
```

may incorrectly return `EBADF`, even though both dirfds should be ignored.

Could the implementation first determine whether each path is absolute, and only translate the corresponding dirfd for relative paths?

## Suggested regression coverage

Please consider adding tests for:

- A real `olddirfd` with a relative `oldpath`.
- Both `olddirfd` and `newdirfd` being real directory descriptors.
- Absolute paths combined with intentionally invalid dirfds.
- `..` traversal and other sandbox-confinement cases.
- The default `flags == 0` symlink behavior in addition to `AT_SYMLINK_FOLLOW`.

I also agree with the existing suggestion to share the trusted implementation between `link` and `linkat`, for example by implementing `link` through the equivalent `linkat(AT_FDCWD, ..., AT_FDCWD, ..., 0)` path. This would reduce duplicated security-sensitive path-resolution and hard-link logic.

qianxichen233 (approved):

PR #1338, opened 2026-08-03
test(iso4-path): Add path confinement tests

lionne99:
Adds ISO-004 test coverage to verify that cage paths remain confined to the virtual root

Changes:
Added 3 `normpath()` unit tests covering:
- Absolute paths being resolved from the virtual root
- Excess `..` components being clamped at the root
- Normal `..` path resolution working correctly

- Added test-only `dashmap` and `parking_lot` dependencies to support the test cage setup

- Added `symlink_confinement.c` end-to-end test to verify that symlinks cannot be used to escape path confinement

### DIFF
--- a/src/typemap/Cargo.toml
+++ b/src/typemap/Cargo.toml
@@ -14,5 +14,3 @@ default = ["fast"]
 fast = []
 secure = []
 
-[dev-dependencies]
-dashmap = "5.1"
--- a/src/typemap/Cargo.toml
+++ b/src/typemap/Cargo.toml
@@ -13,3 +13,7 @@ sysdefs = { path = "../sysdefs" }
 default = ["fast"]
 fast = []
 secure = []
+
+[dev-dependencies]
+dashmap = "5.1"
+parking_lot = "0.12"
--- a/src/typemap/Cargo.toml
+++ b/src/typemap/Cargo.toml
@@ -16,4 +16,3 @@ secure = []
 
 [dev-dependencies]
 dashmap = "5.1"
-parking_lot = "0.12"
--- a/src/typemap/src/path_conversion.rs
+++ b/src/typemap/src/path_conversion.rs
@@ -264,7 +264,7 @@ mod tests {
         let cageid = 1500;
         make_test_cage(cageid, "/");
 
-        let result = normpath(PathBuf::from("/etc/passwd"), cageid);
+        let result = normpath(PathBuf::from("/etc/../etc/passwd"), cageid);
 
         assert_eq!(result, PathBuf::from("/etc/passwd"));
     }
--- a/src/typemap/src/path_conversion.rs
+++ b/src/typemap/src/path_conversion.rs
@@ -226,7 +226,7 @@ mod tests {
         assert_eq!(path_without_trailing_slashes("///"), "/");
     }
 
-    use dashmap::DashMap;
+    use cage::DashMap;
 
     // Builds a minimal cage and registers it under 'cageid', so
     // normpath(path, cageid) can find it via cage::get_cage()
--- a/src/typemap/src/path_conversion.rs
+++ b/src/typemap/src/path_conversion.rs
@@ -225,4 +225,69 @@ mod tests {
         assert_eq!(path_without_trailing_slashes("/"), "/");
         assert_eq!(path_without_trailing_slashes("///"), "/");
     }
+
+    use dashmap::DashMap;
+
+    // Builds a minimal cage and registers it under 'cageid', so
+    // normpath(path, cageid) can find it via cage::get_cage()
+    // Every field besides 'cwd' is irrelevant to path logic, they're
+    // filled with empty/default values to satisfy the struct
+    fn make_test_cage(cageid: u64, cwd: &str) {
+        cage::cagetable_init();
+
+        let test_cage = cage::Cage {
+            cageid,
+            parent: cageid,
+            cwd: cage::RwLock::new(cage::Arc::new(PathBuf::from(cwd))),
+            rev_shm: cage::Mutex::new(Vec::new()),
+            signalhandler: DashMap::new(),
+            sigset: cage::AtomicU64::new(0),
+            pending_signals: cage::RwLock::new(vec![]),
+            epoch_handler: DashMap::new(),
+            os_tid_map: DashMap::new(),
+            main_threadid: cage::RwLock::new(0),
+            interval_timer: cage::IntervalTimer::new(cageid),
+            zombies: cage::RwLock::new(vec![]),
+            child_num: cage::AtomicU64::new(0),
+            vmmap: cage::RwLock::new(cage::Vmmap::new()),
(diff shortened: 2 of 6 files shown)


src/typemap/src/path_conversion.rs
```diff
@@ -225,4 +225,69 @@ mod tests {
...
+    //ISO-004: an absolute path must always be rebuilt from the virtual root
+    // never passed through some other branch unmodified
+    fn normpath_confines_absolute_path_to_virtual_root() {
+        let cageid = 1500;
+        make_test_cage(cageid, "/");
+
+        let result = normpath(PathBuf::from("/etc/passwd"), cageid);
+
+        assert_eq!(result, PathBuf::from("/etc/passwd"));
+    }
```
vidyalakshmir, line 270: This assertion passes whether `normpath` rebuilds the path or does nothing at all. To distinguish the latter case, and test if path is rebuilt, use something like `/etc/../etc/passwd`

tests/unit-tests/file_tests/deterministic/symlink_confinement.c
```diff
@@ -0,0 +1,45 @@
...
+        assert(link_n == direct_n);
+        assert(memcmp(direct_buf, link_buf, (size_t)direct_n) == 0);
+
+        close(direct_fd);
+        close(link_fd);
+    }
+
+    unlink("evil_link");
+
+    printf("symlink_confinement test: PASS\n");
```
vidyalakshmir, line 62: This test asserts that opening via `evil_link` behaves identically to opening `/etc/passwd` directly. if confinement were totally broken and both opens hit the host's real /etc/passwd, the two would still match each other and the test would still PASS.

src/typemap/Cargo.toml
```diff
@@ -13,3 +13,7 @@ sysdefs = { path = "../sysdefs" }
 default = ["fast"]
 fast = []
 secure = []
+
+[dev-dependencies]
+dashmap = "5.1"
+parking_lot = "0.12"
```
vidyalakshmir: where is `parking_lot` dependency used

vidyalakshmir (commented):
Also, this PR does check the path confinement to a virtual root. Just wondering if the chroot to lindfs/ filesystem should also be tested here. @Yaxuan-w

tests/unit-tests/file_tests/deterministic/symlink_confinement.c
Yaxuan-w: Yeah I agree with this.

tests/unit-tests/file_tests/deterministic/symlink_confinement.c
```diff
@@ -0,0 +1,64 @@
...
+    unlink("evil_link");
+
+    errno = 0;
+    /*
+    NOTE: /lind/README.md is specific to this dev-conatainer's mount
+    layout(repo checked out at /lind, matching LINDFS_ROOT's hardcoded
+    assumption in sysdefs). If this runs somewhere that mounts the repo 
+    differently, this path may not exist at all, in which case this 
+    check would accidentally pass via ENOENT.
+    Re-verify this path is valid if test is run in a new environment.
```
Yaxuan-w: The solution could be to pin the expected output instead of deriving it from an unconfined run. Adding:

```sh
tests/unit-tests/file_tests/deterministic/expected/symlink_confinement.output
```

makes [`get_expected_output()`](https://github.com/Lind-Project/lind-wasm/blob/main/scripts/test/harnesses/wasmtestreport.py#L576) return early and skip the native compile+run entirely. Failure detection is preserved: the lind run's return code still goes through [`handle_return_code()`](https://github.com/Lind-Project/lind-wasm/blob/main/scripts/test/harnesses/wasmtestreport.py#L460), so a real escape is still reported as a Runtime Failure regardless of the expected text.

Yaxuan-w (changes requested):

src/typemap/Cargo.toml
```diff
@@ -13,3 +13,6 @@ sysdefs = { path = "../sysdefs" }
 default = ["fast"]
 fast = []
 secure = []
+
+[dev-dependencies]
+dashmap = "5.1"
```
Yaxuan-w: Could you make this import in Cage lib public so that we don't need to re-dependent here? This can help us avoid version mismatch between typemap and cage.

Yaxuan-w (changes requested):

tests/unit-tests/file_tests/deterministic/symlink_confinement.c
```diff
@@ -0,0 +1,86 @@
...
+
+        close(direct_fd);
+        close(link_fd);
+    }
+
+    unlink("evil_link");
+
+    errno = 0;
+    /*
+    NOTE: /lind/README.md is specific to this dev-conatainer's mount
```
vidyalakshmir: container //typo

vidyalakshmir: Opening `/lind/README.md` from inside the cage resolves to `<LINDFS_ROOT>/lind/README.md`, which doesn't exist. This would result in a return value of `ENOENT` which would pass this test, but that doesn't necessarily test policy enforcement. All it says is that name did not resolve to anything inside the virtual filesystem. Also, doing a generic test here would be problematic when the test is run natively.


A suggested solution is to have a file within the virtual filesystem and outside either.

File `/tmp/lind/sentinel.txt` will contain the text "LIND_HOST_ONLY"

and File `$LINDFS_ROOT/tmp/lind/sentinel` will contain the text "LIND_CAGE_ONLY".

So when you run it within lind-wasm, the following three conditions can hold,

1. Reads LIND_CAGE_ONLY means confinement works, and test passes
2. Reads LIND_HOST_ONLY means there is an escape and test fails
3. If open fails, then failure

vidyalakshmir (changes requested):

vidyalakshmir:
Posting review comments here again.. 

Opening `/lind/README.md` from inside the cage resolves to `<LINDFS_ROOT>/lind/README.md`, which doesn't exist. This would result in a return value of `ENOENT` which would pass this test, but that doesn't necessarily test policy enforcement. All it says is that name did not resolve to anything inside the virtual filesystem. Also, doing a generic test here would be problematic when the test is run natively.


A suggested solution is to have a file within the virtual filesystem and outside either.

File `/tmp/lind/sentinel.txt` will contain the text "LIND_HOST_ONLY"

and File `$LINDFS_ROOT/tmp/lind/sentinel` will contain the text "LIND_CAGE_ONLY".

So when you run it within lind-wasm, the following three conditions can hold,

1. Reads LIND_CAGE_ONLY means confinement works, and test passes
2. Reads LIND_HOST_ONLY means there is an escape and test fails
3. If open fails, then failure 

@Yaxuan-w Are these tests also compared against native execution? If so, I wonder if we really need to do it. Because testing confinement only applies to sandboxed execution and not native, so there is no behavior to compare against.

PR #1343, opened 2026-08-10
fix: validate guest pointer in debug-panic host function

### LINKED ISSUE #1056: Debug panic reads raw host pointer without validation
`src/wasmtime/crates/lind-common/src/lib.rs:234-237`

The debug function reads from a raw pointer without base addition or bounds checking, bypassing normal syscall address translation.

rishabhBudhouliya:
## Summary

`src/wasmtime/crates/lind-common/src/lib.rs` registered `lind`/`debug-panic`
as `CStr::from_ptr(arg)` on a u64 straight from the guest. glibc pre-translated
the pointer (`lind_debug.c:18`), so the host dereferenced a guest-controlled
host address: arbitrary host-memory read, unbounded walk on an unterminated
string, and a `.to_str().unwrap()` panic on non-UTF-8. The read happened even
in `NO_LOGGING=1` builds, where the result is discarded.

## Fix

- glibc passes a guest pointer; the host adds the base itself.
- New read_guest_cstr bounds the offset to 32 bits, so base + offset always lands inside the cage's own memory, then calls check_addr_read before dereferencing. The memory size can't serve as the check.
- Reads are capped at 1024 bytes and at the end of the starting page; from_utf8_lossy instead of unwrap.
- `lind_debug_str` uses the same reader.

## Related issues

Should close: https://github.com/Lind-Project/lind-wasm/issues/1056

## Testing

Minimal repro — a module importing `debug-panic` directly and passing `0xdeadbeef`:

| | unpatched | patched |
|---|---|---|
| exit | 139 (SIGSEGV) | 0 |
| output | — | `FROM GUEST: <unreadable message pointer 0xdeadbeef>` |

Also checked: valid guest pointer still delivers its string; `0x1122334455667788`
rejected; real glibc call site prints `FROM GUEST: sched_getparam called but not
supported!`; default `PanicAndExit` still panics with the message (exit 101);
`lind_debug_str` verified under a `lind_debug` build.

Full `make build` + unit tests:
- math, filesystem, memory, process — 164 passed, 2 failed
- signals, networking, dynamic-linking — 77 passed, 0 failed

## User behavior related:
A message straddling a page boundary is truncated at the boundary. String literals are fixed at link time, so this is deterministic per build rather than intermittent, and it affects only diagnostic text on already-fatal paths.

## Compatibility and security impact

**Security:**. `lind`/`debug-panic` previously dereferenced a guest-controlled host address with no base addition and no bounds check, so any module could import it directly and make the runtime read arbitrary
host memory (contents land in the log) or segfault. It also walked past the end of
an unterminated string and panicked the host via `.to_str().unwrap()` on non-UTF-8.

**Performance:**  Validation costs exactly one check_addr_read per call — one vmmap lock acquisition — on paths that are already fatal or debug-only.

## Checklist

- [ ] I added or updated tests where appropriate.
- [ ] I formatted the code and ran the relevant linters.

### DIFF
--- a/src/wasmtime/crates/lind-common/src/lib.rs
+++ b/src/wasmtime/crates/lind-common/src/lib.rs
@@ -1,11 +1,12 @@
 #![allow(dead_code)]
 
 use anyhow::Result;
-use cage::memory::check_addr_write;
+use cage::memory::{check_addr_read, check_addr_write};
 use rand::Rng;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
 use sysdefs::constants::Errno;
+use sysdefs::constants::fs_const::PAGESIZE;
 use sysdefs::constants::lind_platform_const;
 use sysdefs::constants::lind_platform_const::{UNUSED_ARG, UNUSED_ID};
 use sysdefs::lind_debug_panic;
@@ -13,7 +14,6 @@ use threei::threei::{
     copy_data_between_cages, copy_handler_table_to_cage, make_syscall, register_handler,
 };
 use threei::threei_const;
-use typemap::path_conversion::get_cstr;
 use wasmtime::{AsContext, AsContextMut, AsyncifyState, Caller};
 use wasmtime_lind_dylink::DynamicLoader;
 use wasmtime_lind_multi_process::{LindHost, get_memory_base, get_memory_base_and_size};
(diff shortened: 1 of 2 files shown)


qianxichen233:
ideally, we should never need/have to do this kind of checks. debug-panic are used for debugging purpose only, in release build (not --release build, the actual lind release version) all the debug panic functions should not be present in the binary at all and hence there isn't any way for an attacker to invoke this API since this API is supposed to disappear in release/deploy build in the first place. If we are talking about debug build, then I think these checks are not really **necessary** as well since it's just for debugging purpose.

That's to say, this PR can still be merged as the temporary fix, since the above plan is not really implemented either (i.e. release flag gated debug panic API) but the latter one should be our long term solution

rishabhBudhouliya:
@qianxichen233 Agreed on the direction and release approach. A build with the import isn't registered is better. 

However, when I was looking at the usage of debug-panic, I saw it outside the #ifdef at couple of places like 
the six `sched_*` stubs, the generic `syscall()` entry, the `clone-internal.c:52` fallback, and the setjmp-table OOM
at `wasm_eh_setjmp.c:95`. 

So, gating the import out of release also needs a decision on what those sites do instead (abort? return an errno?)

qianxichen233:
> @qianxichen233 Agreed on the direction and release approach. A build with the import isn't registered is better.
> 
> However, when I was looking at the usage of debug-panic, I saw it outside the #ifdef at couple of places like the six `sched_*` stubs, the generic `syscall()` entry, the `clone-internal.c:52` fallback, and the setjmp-table OOM at `wasm_eh_setjmp.c:95`.
> 
> So, gating the import out of release also needs a decision on what those sites do instead (abort? return an errno?)

My thinking is to make the body of lind_debug_panic wrapped inside `#ifdef`, then compiler should optimize out these invocation automatically. Though have the actual invoking site wrapped inside `#ifdef` is also a reasonable choice

rishabhBudhouliya:
@qianxichen233  I looked at what would happen if we wrap the body in #ifdef. The default panic behavior is PanicAndExit (sysdefs/src/logging.rs:175), so these call sites abort the process today. With an empty body they would return their fallback values instead, and nothing would report the problem:

- sched_getp.c:10 returns -1 without setting errno
- clone-internal.c:53 returns 0, which the caller reads as a successful child
- wasm_eh_setjmp.c:96 returns the un-grown table after a failed realloc

So I think the release-gated version needs two entry points instead of one #ifdef. One would be an unconditional abort that takes no message, used for the unimplemented-feature markers. No guest pointer crosses the boundary there, so there is nothing to validate. The other would be the string version, gated behind LIND_DEBUG. Filed as #1358 .

This PR stands as the temporary fix and CI is green. Could you or @Yaxuan-w submit a formal approval? It is currently blocked on review.

src/wasmtime/crates/lind-common/src/lib.rs
```diff
@@ -242,6 +242,45 @@ fn add_syscall_to_linker<
...
+/// panic the host.
+fn read_guest_cstr<
+    T: LindHost<T, U> + Clone + Send + 'static + std::marker::Sync,
+    U: Clone + Send + 'static + std::marker::Sync,
+>(
+    caller: &mut Caller<'_, T>,
+    offset: u64,
+) -> Result<String, Errno> {
+    // Bounds `off` so `base + off` below cannot wrap.
+    if offset > u32::MAX as u64 {
```
Yaxuan-w: The upper bound is here but there's no lower one, and `0` is a reachable, *readable* offset — so a NULL message pointer now passes validation instead of being reported.

`lind_debug_panic(NULL)` sends `(uint64_t)(uintptr_t)NULL == 0`. For a statically linked main module, `instance.rs:570` calls `init_vmmap` and then mmaps `[start_addr = 0, start_addr + required_memory_size)` as `PROT_READ|PROT_WRITE`, so `check_addr_read(cageid, base + 0, len)` succeeds. The host reads guest byte 0 — normally NUL — and logs `FROM GUEST: ` with an empty message, which hides the fact that the caller passed a null pointer.

Worth noting the old code had the same blind spot in the other direction: `__lind_translate_ptr_to_host` special-cases `p == NULL` to `0ULL`, and the host then did `CStr::from_ptr(0)`. So this is still an improvement — just an easy one to finish:

```suggestion
    if offset == 0 || offset > u32::MAX as u64 {
```

That routes NULL to the existing `<unreadable message pointer 0x0>` branch.

rishabhBudhouliya: Thanks Alice for the suggestion and a detailed explanation, great catch. I will fix this.

rishabhBudhouliya: @Yaxuan-w i have addressed the review comment

Yaxuan-w (approved):

qianxichen233 (approved):

PR #1349, opened 2026-08-12
test(cage,fdtables): add CONC-001..005 concurrency and isolation coverage

### LINKED ISSUE #1304: tests: concurrency, signals, and lifecycle test coverage
This sub-issue tracks concurrency and lifecycle tests for Lind-Wasm. The goal is to verify that cage creation/destruction, fd-table operations, cage-table operations, resource accounting, and concurrent syscall execution preserve runtime invariants.

### Test Matrix

| ID | Test | Oracle | How to achieve |
|----|------|--------|----------------|
| CONC-001 | Cage spawn/destroy stress: rapid create/destroy while cages issue syscalls | No cage-table race, no leaked host resource | Cage lib inner test suite |
| CONC-002 | Concurrent cage-table + fd-table + FS-metadata access | Zero data races | General unit tests |
| CONC-003 | Cage-table and fd refcount operations | All interleavings preserve invariants | General unit tests |
| CONC-004 | fd refcount conservation under concurrent dup/close/fork | No double-free, no leak | General unit tests |
| CONC-005 | Per-cage resource exhaustion: fd flood, unbounded memory growth, syscall flood; confirm one cage cannot starve others or the host | Limits enforced per cage | General unit tests |

### Progress

| ID | Status | Notes |
|----|--------|-------|
| CONC-001 | Not started | |
| CONC-002 | Not started | |
| CONC-003 | Not started | |
| CONC-004 | Not started | |
| CONC-005 | Not started | |

Shounak-Ghosh:
## Summary

Implements the full CONC-001..005 test matrix from #1304, which asks us to verify
that cage creation/destruction, fd-table operations, cage-table operations,
resource accounting, and concurrent syscall execution preserve runtime invariants.

**This PR contains tests only.** An earlier revision also carried the 11 bug fixes
these tests surfaced; per review feedback those are now split out into one issue
and one PR per bug.

Also adds a `rust-unit-tests` CI job for the `cage` and `fdtables` crates, whose
unit tests have never run in CI. The `cage` suite does not currently compile: two
tests in `vmmap.rs` bind a `Vmmap` without `mut` and then mutate it. Both are
one-word corrections inside `#[cfg(test)] mod tests`, so they touch test code
rather than runtime code, and they are here as the job cannot run
without them.

## Related issues

Closes #1304.

Mapping to the issue's matrix (the issue lists CONC-005 as one row covering three
resources; it is implemented here as three separate tests, one per resource):

| Row | Oracle (from #1304) | Implemented as |
|---|---|---|
| CONC-001 | No cage-table race, no leaked host resource | `src/cage/src/cage.rs`: `conc_001_cage_spawn_destroy_stress`, `conc_001_finalize_waits_for_grate_inflight` (matches the issue's "Cage lib inner test suite") |
| CONC-002 | Zero data races | `conc_002_cage_fd_fs_stress.c` + 4 in-crate `conc_002_*` tests |
| CONC-003 | All interleavings preserve invariants | `conc_003_cage_fd_refcounts.c` + 4 in-crate `conc_003_*` tests |
| CONC-004 | No double-free, no leak | `conc_004_dup_close_fork_refcounts.c` + 2 in-crate `conc_004_*` tests |
| CONC-005 | Limits enforced per cage | `conc_005_fd_exhaustion_isolation.c`, `conc_005_memory_pressure_isolation.c`, `conc_005_syscall_flood_isolation.c` + 3 in-crate `conc_005_*` tests |

The C and Rust suites are deliberately complementary, not redundant: the C side
is a black-box differential test (the same binary is run natively and under lind,
and stdout is diffed), while the in-crate Rust side asserts exact refcount
equalities that the public C/POSIX interface cannot observe at all. Where the C
test can only watch a pipe reach EOF, the Rust mirror reads the count straight
out of a registered close handler.

## Bugs found

Writing these tests surfaced **11 defects**, plus one missing capability and one
behavioural inconsistency that needs a team decision. Per review feedback none of
the fixes are in this PR. Each is filed as its own issue and will be submitted as
its own PR after discussion. The full set of fixes is preserved verbatim on branch
**`test-conc-bug-fixes`**, where each is paired with the test that found it.

Three of the 11 are guest-triggerable, including one that lets any cage terminate
the entire runtime; those are called out in their respective issues.

Issue list: *(to be filled in once filed)*

## Currently ignored / skipped

Eight tests cannot pass on current `main` and are marked `#[ignore]` or listed in
`skip_test_cases.txt`. **The CI job is green as submitted.** Each is annotated in
place with the condition that unblocks it, so the fix PRs can find them by grep.

### Rust: `#[ignore]` (5)

| Test | Behaviour on unfixed `main` |
|---|---|
| `cross_cage_fd_isolation` | Fails; **pre-existing on `main`**, not from this PR |
| `get_specific_virtual_fd_rejects_fd_at_max` | **Panics** on an out-of-bounds index |
| `conc_002_shared_underfd_refcount_race` | **Deadlocks** |
| `conc_002_copy_fdtable_vs_concurrent_churn` | **Deadlocks** |
| `conc_003_dup2_overwrite_refcount_conservation` | **Deadlocks** |

The three deadlocks are not a judgement call: the refcount corruption panics one
worker thread, and the surviving workers then block forever on the round barrier.
Unfixed, these hang the test binary rather than reporting a failure, so `#[ignore]`
is required rather than merely tidy.

### C: `skip_test_cases.txt` (3)

| Test | Why |
|---|---|
| `conc_004_dup_close_fork_refcounts.c` | One assertion only (`dup2(BADFD, BADFD)`); the rest of the file passes today |
| `conc_005_fd_exhaustion_isolation.c` | Needs three fd-exhaustion fixes; running it unfixed also leaks host descriptors |
| `conc_005_memory_pressure_isolation.c` | Its `setrlimit()` reaches a `lind_debug_panic!` and **takes the whole runtime down**; unfixed it does not fail, it kills the harness |

Each skipped file's header comment explains what blocks it and what to remove when
the fix lands.

## Testing

```bash
# Rust in-crate suites (same invocation as the new CI job)
cargo test --manifest-path src/cage/Cargo.toml -- --test-threads=1
cargo test --manifest-path src/fdtables/Cargo.toml -- --test-threads=1

# C tests, native + wasm, stdout diffed by the harness
python3 scripts/test/harnesses/wasmtestreport.py \
  --testfiles tests/unit-tests/process_tests/deterministic/conc_002_cage_fd_fs_stress.c \
              tests/unit-tests/process_tests/deterministic/conc_003_cage_fd_refcounts.c \
              tests/unit-tests/process_tests/deterministic/conc_005_syscall_flood_isolation.c \
  --compile-flags -pthread -lpthread
```

Results on this branch:

- `cage`: **38/38 pass**, including both `conc_001_*` tests.
- `fdtables`: **45 pass, 0 fail, 6 ignored** (the 5 above plus one pre-existing
  `#[ignore]`), and a second binary at 24/24.
- Re-run 25× (`fdtables`) and 15× (`cage`) consecutively with **zero** failures or
  hangs, since several of these tests are probabilistic by construction and a single
  green run proves little.
- `cargo fmt --check` clean on both crates.

Every C test is written to be harness-safe: exactly one line of stdout, and no
pids, clocks, addresses, fd numbers, sizes, or errno values are ever printed or
compared, since those legitimately differ between the native and lind runs.
Diagnostics go to fd 2, which the harness surfaces only on a nonzero exit.

### CI

Adds a `rust-unit-tests` job to `lint.yml` running both crates' tests.
`--test-threads=1` is mandatory: both crates' tests mutate process-global state
(`CAGE_MAP`; `FDTABLE`/`FDCOUNT`/`CLOSEHANDLERTABLE`) and serialize on hand-rolled
mutexes rather than real test isolation.

One thing worth knowing about the existing suite: `refresh()` clears `FDTABLE` and
`CLOSEHANDLERTABLE` but **not** `FDCOUNT`, so refcount state leaks between tests in
one binary. At least one pre-existing test (`test_dup_close`, a `#[should_panic]`)
depends on that leakage and fails when run in isolation. Every test added here uses
a dedicated `fdkind` and a disjoint `underfd` window specifically so it never
aliases another test's leftovers, but the ordering fragility in the older tests is
worth a follow-up.

## Compatibility

No production code changes, so **no guest-visible behaviour changes**. The only
non-test files touched are `.github/workflows/lint.yml` (new job) and
`skip_test_cases.txt`.

## Notes for reviewers

Suggested reading order:

1. **CI**: `.github/workflows/lint.yml`, and the ignore/skip set above.
2. **Tests**: the six `conc_*.c` files and the in-crate `mod tests` additions.

Each ignored or skipped test states, at its definition, the condition that unblocks
it. Those are worth reading alongside the corresponding issue.

## Checklist

---

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -61,3 +61,28 @@ jobs:
       #   env:
       #     GH_TOKEN: ${{ github.token }}
       #   run: zizmor -v .github/workflows #Running with -v to show all passes, will halt if any fail
+
+  rust-unit-tests:
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+
+      # Deliberately no `rustup default` override here: this uses the
+      # toolchain pinned in rust-toolchain.toml, unlike the `lint` job above
+      # which pins an older nightly for clippy compatibility (see #242).
+      - name: Cage crate unit tests
+        run: cargo test --manifest-path src/cage/Cargo.toml -- --test-threads=1
+
+      # --test-threads=1 is mandatory: fdtables' tests share process-global
+      # state (FDTABLE / FDCOUNT / CLOSEHANDLERTABLE) and serialize on a
+      # hand-rolled TESTMUTEX that several tests take only after calling
+      # refresh(), which wipes that same state. This runs against the
+      # default `dashmaparray` feature only, which is what the rest of the
+      # build uses (see FDTABLES_IMPL in the top-level Makefile).
+      - name: fdtables crate unit tests
+        run: cargo test --manifest-path src/fdtables/Cargo.toml -- --test-threads=1
--- a/src/fdtables/src/lib.rs
+++ b/src/fdtables/src/lib.rs
@@ -188,6 +188,10 @@ mod tests {
 
     use std::collections::HashSet;
 
+    use std::sync::atomic::{AtomicU64, Ordering};
+
+    use std::sync::{Arc, Barrier};
+
     // I'm having a global testing mutex because otherwise the tests will
     // run concurrently.  This messes up some tests, especially testing
     // that tries to get all FDs, etc.
@@ -240,6 +244,11 @@ mod tests {
     }
 
     #[test]
+    // Pre-existing failure on main, not introduced here: this asserts EBADF but
+    // every backend's translate_virtual_fd returns EBADFD, while
+    // get_specific_virtual_fd returns EBADF for the same class of error. Which
+    // side is wrong is a team decision, not a fix.
+    #[ignore = "pre-existing: translate_virtual_fd returns EBADFD where this expects EBADF; which side is wrong needs a team decision"]
     // ISO-002 fd-lib inner test: two cages have independent fd tables, so one
     // cage cannot reach another cage's descriptors
     fn cross_cage_fd_isolation() {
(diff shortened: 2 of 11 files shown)


Yaxuan-w (commented):
Thanks for doing all the work here. These are valuable! I think some changes will need discussion (especially for the glibc and vmmap bug fixes). That would be great to:

- Only include test in this PR (put test into skip list if they cannot be passed on current version)
- Open up issues for **each bug** you found
- **Submit PR per bug** after the discussion with the team

.github/workflows/lint.yml
```diff
@@ -60,4 +60,29 @@ jobs:
...
+        run: cargo test --manifest-path src/cage/Cargo.toml -- --test-threads=1
+
+      # --test-threads=1 is mandatory: fdtables' tests share process-global
+      # state (FDTABLE / FDCOUNT / CLOSEHANDLERTABLE) and serialize on a
+      # hand-rolled TESTMUTEX that several tests take only after calling
+      # refresh(), which wipes that same state. This runs against the
+      # default `dashmaparray` feature only, which is what the rest of the
+      # build uses (see FDTABLES_IMPL in the top-level Makefile).
+      - name: fdtables crate unit tests
+        run: cargo test --manifest-path src/fdtables/Cargo.toml -- --test-threads=1
```
Yaxuan-w, line 88: Add new line

tests/unit-tests/process_tests/deterministic/conc_005_syscall_flood_isolation.c
```diff
@@ -0,0 +1,406 @@
...
+     * by another cage's syscall volume would never get here, and the
+     * harness's 30s timeout would report it. */
+    assert(xwaitpid(pb, &status) == pb);
+    assert(WIFEXITED(status));
+    expect_exit0(status, "B");
+    assert(sh->b_progress == TARGET);
+
+    /* Wind the flood down and confirm A itself was healthy throughout --
+     * a flooder that had died early would have made the test vacuous. */
+    sh->stop = 1;
```
Yaxuan-w, line 372: If any assertion between releasing A and `sh->stop = 1` fires, the parent aborts without stopping the flood, and A's threads spin forever (no other exit condition). Natively that leaves an orphan on 5 cores holding the stdout pipe, so the harness reports `Native_Timeout`; under lind the crash path waits on A until the 30s timeout. Every failure of this test therefore shows up as a timeout instead of the diagnostics written to fd 2.

Probably need to set `sh->stop = 1` (or `kill(pa, SIGKILL)`) before any abort path, e.g. in `expect_exit0`'s failure branch or via `atexit`

src/fdtables/src/lib.rs
```diff
@@ -1713,4 +1722,2223 @@ mod tests {
...
+                Err(_) => TESTMUTEX.clear_poison(),
+            }
+        }
+    }
+
+    /// Must be called *after* refresh(): refresh() clears
+    /// CLOSEHANDLERTABLE, so a one-time (e.g. std::sync::Once) handler
+    /// registration would silently lose the handlers on the next test that
+    /// calls refresh().
+    fn c2_setup() {
```
Yaxuan-w, line 1813: These bookkeeping mutexes will poison-cascade on the first real failure. Several asserts hold the guard while panicking...

src/fdtables/src/lib.rs
```diff
@@ -1713,4 +1722,2223 @@ mod tests {
...
+    ///
+    /// In-crate mirror of what
+    /// tests/unit-tests/process_tests/deterministic/conc_002_cage_fd_fs_stress.c
+    /// drives from the C side: fork() interleaved with concurrent fd-table
+    /// churn in the forking cage.
+    ///
+    /// Ignored rather than merely failing: the desync panics a worker and the
+    /// rest block forever on the round barrier, so on an unfixed tree this
+    /// DEADLOCKS the test binary rather than reporting a failure.
+    #[ignore = "deadlocks until copy_fdtable_for_cage snapshots and increments under one guard"]
```
Yaxuan-w, line 2313: seems this test doesn't have barrier

src/fdtables/src/lib.rs
```diff
@@ -1713,4 +1722,2223 @@ mod tests {
...
+    }
+
+    #[test]
+    /// A stable cage with a known fd-table, plus victim cages the main
+    /// thread creates and destroys underneath 4 reader threads. Readers
+    /// translate only in the stable cage and observe victims exclusively
+    /// through check_cage_exists(), which does not separately assert-then-
+    /// unwrap a FDTABLE.get() (Test 4 drives the analogous race in
+    /// copy_fdtable_for_cage specifically).
+    fn conc_002_translate_isolation_under_cage_removal() {
```
Yaxuan-w, line 2057: Seems TOCTOU not triggered by this test

Yaxuan-w (changes requested):
Need another round of review after fixing these issues

PR #1357, opened 2026-08-17
fix(exec): reuse the cage's linear memory instead of mapping a second 4 GiB

### LINKED ISSUE #1354: bug: execve allocates a second 4 GiB linear memory instead of replacing the first
### Before submitting

- [x] I searched existing issues for duplicates.
- [x] This report does not disclose a security vulnerability.

### Problem

 `execve` allocates a second 4.06 GiB (`0x104000000`) reservation (instead of replacing a cage's linear memory) releasing the old only after the exec'd program finishes. A plain `execv` with no fork costs two reservations; `fork()`+`execve()` costs three.

Separately, an exec'd cage's reservation is never released on cage exit or for the process lifetime.

We were trying to run gcc on enarx when we saw this issue: 
 `gcc hello.c -o hello` (driver → cc1 → as → ld) allocates 7 reservations, releases
  3, and peaks at **5 simultaneous = 20.31 GiB** of address space for at most 2 live
  cages. 16.25 GiB is still mapped at exit.

and saw this error: 
```
thread 'lind-fork-3' panicked at lind-multi-process/src/lib.rs:457:26:
failed to create child linker: mmap failed to allocate 0x104000000 bytes
Caused by: Out of memory (os error 12)
```

### Steps to reproduce

One needs to strace a repo script to be able to witness the redundant allocation via mmap

1. Two programs
```c
/* repro_parent.c */
#include <unistd.h>
#include <sys/wait.h>
int main(void) {
  char *argv[] = {"/repro_child", NULL};
  if (fork() == 0) { execv("/repro_child", argv); _exit(127); }
  wait(NULL);
  return 0;
}
/* repro_child.c */
#include <stdio.h>
int main(void) { printf("child\n"); return 0; }
```

2. Build static and stage:

./scripts/bin/lind_compile -s repro_parent.c
./scripts/bin/lind_compile -s repro_child.c
sudo cp repro_parent.cwasm lindfs/repro_parent
sudo cp repro_child.cwasm  lindfs/repro_child
sudo chmod 755 lindfs/repro_parent lindfs/repro_child

3. Trace the reservations:
sudo strace -f -qq -e trace=mmap,munmap -o /tmp/t.log ./build/lind-boot /repro_parent
grep 4362076160 /tmp/t.lo

### Environment

- lind-wasm main @ 067007a4e
- Docker securesystemslab/lind-wasm-dev, host Ubuntu 24.04, kernel 6.11, x86_64
- SGX figures from TriSeal open-futex-fs-syscalls; SGX1+SGX2, EPC 2,122,313,728 B

### Logs and additional context

```shell

```

Yaxuan-w:
### Problems

1. `execve` replaces a cage's program image but kept allocating a brand-new 4 GiB linear-memory reservation for the new image. Because the exec-ed program runs nested inside the pre-exec `Store`'s host call stack, the old reservation stays mapped until it finishes so a plain `execv` peaked at two reservations and `fork()`+`execve()` at three.
2. `load_main_module` registers a grate worker pool in the process-wide `GRATE_POOL`, and each worker owns a `Store` attached to the cage's linear memory. `cleanup_grate_handler` is never called from anywhere (`get_runtime_cleanup_funcptr` has zero call sites), so that entry and the 4 GiB behind it survived for the lifetime of the host process.

### Changes:

- `attach_shared_memory` takes an optional existing `SharedMemory`. On the exec path the cage's current reservation is passed in and reused instead of calling `SharedMemory::new`; the grow delta is now computed from `mem.size()` rather than `combined_min`, since a reused memory is already at 65536 pages.
- When reusing, the whole 4 GiB is re-established as a fresh anonymous mapping (`MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED`, `PROT_NONE`)
- `execve_call` looks the cage's memory up in its own linker.
- `execute_with_lind` unregisters the cage's grate handler once the program has finished. 

`execv` goes from 2 reservations to 1, `fork()`+`execve()` from 3 to 2, and a cage's address space is actually returned when it exits.

## Related issues

Closes #1354

## Checklist

### DIFF
--- a/src/wasmtime/crates/lind-multi-process/src/lib.rs
+++ b/src/wasmtime/crates/lind-multi-process/src/lib.rs
@@ -7,7 +7,8 @@ use std::ffi::c_void;
 use std::ptr::NonNull;
 use sysdefs::constants::fs_const::{MAP_ANONYMOUS, MAP_FIXED, MAP_PRIVATE, PROT_NONE};
 use sysdefs::constants::lind_platform_const::{
-    UNUSED_ARG, UNUSED_ID, UNUSED_NAME, unset_stack_arena_base,
+    LIND_MEMORY_IMPORT_MODULE, LIND_MEMORY_IMPORT_NAME, UNUSED_ARG, UNUSED_ID, UNUSED_NAME,
+    unset_stack_arena_base,
 };
 use sysdefs::constants::syscall_const::{EXEC_SYSCALL, EXIT_SYSCALL, FORK_SYSCALL};
 use sysdefs::constants::{Errno, MAX_SHEBANG_DEPTH, MMAP_SYSCALL};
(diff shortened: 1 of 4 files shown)


src/wasmtime/crates/lind-multi-process/src/lib.rs
```diff
@@ -2451,21 +2488,76 @@ pub fn attach_shared_memory<
...
     Ok(())
 }
 
+/// Return the `(module, name)` pair under which `module` imports its shared linear
+/// memory, or `None` if it does not import one.
+///
+/// Both `attach_shared_memory` (which defines the memory in a cage's linker) and
+/// the exec path (which looks the memory back up so it can be reused) need this
+/// name, so it is derived in one place.
+fn shared_memory_import_name(module: &Module) -> Option<(String, String)> {
```
qianxichen233, line 2549: `shared_memory_import_name` seems like a overengineer for me, the memory in lind is fixed to be env::memory, do the search is meaningless in most of the case and is an non-necessary overhead

qianxichen233 (commented):
also there is a problem when reuse the linear memory on exec when the process is multithreaded, since other threads are still using the memory. In native, on exec, all other threads in the process should be terminated. It can be fixed later if you want (but should definitely put an issue), this scenario seems not touched by our existing applications yet

rishabhBudhouliya:
I had a concern about what happens once a grate's `execute_with_lind` returns back? Does subsequent interposed syscalls from another live child survive or abort the entire runtime?

qianxichen233 (approved):

PR #1378, opened 2026-09-12
lind overlay for the Rust libc crate

### LINKED ISSUE #123: fix nested address translation in dispatcher

rishabhBudhouliya:
## Summary

Rust programs built for lind use the `wasm32-wasip1` target. The Rust libc crate picks its wasi module for that target, which describes wasi-libc. However, lind links the binary against lind-glibc. Struct layouts, open flags and errno values don't match causing syscall failures.

Some observed behavior when trying to run in-toto in lind-wasm:

- metadata(): len=34359742464, is_file=false (st_size read from the wrong offset)
- read_dir(): four entries with empty names (d_name read from the wrong offset)
- File::create(): "os error 2" (wasi O_CREAT bit passed to a Linux open)

I was wondering why grates never have this problem and it turns out they do file I/O through raw 3i calls. The imfs grate has a workaround comment for the same bug. Would be interesting to hear more about how the grates designed around this issue.

## Testing

scripts/rust/make_lind_libc.sh copies libc 0.2.189 from the cargo registry into build/lind-libc and applies scripts/rust/lind-libc-wasi.patch. The patch changes one file, src/wasi/mod.rs: the stat and dirent layouts as measured under lind.

A crate uses it with: cargo --config 'patch.crates-io.libc.path="build/lind-libc"'

## Checklist

- [ ] I have read the [contribution guidelines](https://github.com/Lind-Project/lind-wasm/blob/main/CONTRIBUTING.md).
- [ ] I added or updated tests where appropriate.
- [ ] I ran the relevant tests locally and they pass.
- [ ] I formatted the code and ran the relevant linters.


scripts/rust/lind-libc-wasi.patch
Yaxuan-w, line 1: Seems this file is a `git diff` result? What's the purpose of this file?

rishabhBudhouliya: Yes Alice, it is a diff. make_lind_libc.sh copies the unmodified libc 0.2.189 crate from the cargo registry into build/lind-libc and applies this file to src/wasi/mod.rs. The patch changes the struct stat and struct dirent layouts, the 32-bit `off_t/ino_t/nlink_t/blkcnt_t`, and Linux `O_*/AT_*/DT_*/errno` values. We could think of an alternative you mentioned on #1381, a proper wasm32-lind Rust target that would remove the need for this file entirely.

Yaxuan-w: Oh thanks for clarification and this makes sense to me. I was curious on whether this git diff file is the supposed file format for our usage

Yaxuan-w (approved):

PR #1385, opened 2026-09-17
improve fork performance by avoiding copying unnescessary grate poll stack

qianxichen233:
Each stack in grate pool used to always assigned with RW permission at start even if they aren't used. This is a huge allocation. During fork, all the RW memories will be copied entirely including those unused grate stacks (8MB * grate_worker_number). This is a huge amount of unnescessary work.

This PR make grate worker's stack always assigned with PROT_NONE at start, so that fork will skip these slots as they are not used. Only when the grate worker is activated, the worker's stack is mmap-ed into RW and can be potentially copied at fork.

speed up fork 10-100 times faster depending on different scenarios


src/wasmtime/crates/lind-3i/src/lib.rs
```diff
@@ -711,6 +714,31 @@ where
 
     let stack_base = worker_stack_base(cageid, worker_id);
     let stack_top = worker_stack_top(cageid, worker_id);
+
+    // The slot starts PROT_NONE (see new_started_impl_with_lind); activate
+    // just this worker's own slot now. Leave the guard page before it
+    // PROT_NONE so stack overflow still traps.
```
Yaxuan-w, line 720: > so stack overflow still traps

The 4K guard in front of each slot never enters the vmmap but our vmmap walks gaps in ascending order starting from address 0. This comment seems doesn't hold bc a worker stack overflow now silently clobbers whatever the user just mmapped there.

src/sysdefs/src/constants/lind_platform_const.rs
```diff
@@ -285,3 +285,34 @@ pub fn fork_stack_arena_base_for_child(
         .ok_or("parent grate stack arena base not initialized")?;
     init_stack_arena_base(child_grate_id, parent_base)
 }
+
+/// Tracks, per cage, whether a grate handler has ever been registered.
+/// Lets `fork` skip rebuilding a worker pool for cages that never use grates.
+pub static GRATE_HANDLER_REGISTERED: OnceLock<RwLock<Vec<bool>>> = OnceLock::new();
```
Yaxuan-w, line 291: Doesn't need to be pub

src/lind-boot/src/lind_wasmtime/execute.rs
```diff
@@ -745,9 +751,6 @@ fn load_main_module(
         };
         let host = store.data().clone();
 
-        // initialize the grate pool for later use in grate calls and
-        // other syscalls that require re-entry into wasmtime runtime.
-        init_grate_pool();
         unregister_grate_handler(cageid);
```
Yaxuan-w, line 754: We probably need to keep `unregister_grate_handler(cageid)` unconditional and clear the flag in the branch that does not register.

If a cage with a registered handler execs into a module that does not export `pass_fptr_to_wt`, the stale handler stays in `GRATE_POOL`. Its 32 worker Stores keep the old module and old linear memory alive, and a grate call targeting that cage id runs into the replaced image. `GRATE_HANDLER_REGISTERED[cageid]` is also never cleared, so later forks rebuild a pool for nothing.

Yaxuan-w (changes requested):

PR #1386, opened 2026-09-17
Fix release image usable as a base for app builds

Uk-jake:
## Summary
Running the apps' tests in the pipeline requires a base image. 

I tested the apps' tests using the dev image as the base image, but most tests failed. I guess the dev image is prebuilt with `make lind-debug`  so apps built against it do not run correctly. 

we decided to use the release image as a base image, but it needs to be fixed. The current release image has only sysroot, but several app build scripts need lind_debug.o, which is built in the build tree.

## Changes
1. Add `src/glibc/build/lind_debug.o`
Several app build scripts in `lind-wasm-apps` (zlib, gnulib, openssl, bash, libtirpc) link `$LIND_WASM_ROOT/src/glibc/build/lind_debug.o` directly. This object lives in the glibc build tree, not the installed sysroot, so the current `release` stage.

2. Add the `/home/lind/lind-wasm` symlink
It makes `/home/lind/lind-wasm/lindfs`resolve to `/lindfs` .

## Related issues
- Build CICD pipeline(https://github.com/Lind-Project/lind-wasm-apps/issues/275) 

## Testing
Ran three apps from a clean build tree, mounting the apps source into the container.

App | Result
-- | --
sed |  7/7 passed
curl |  25/25 passed
git |  36/36 passed

## Checklist

- [X] I have read the [contribution guidelines](https://github.com/Lind-Project/lind-wasm/blob/main/CONTRIBUTING.md).
- [X] I added or updated tests where appropriate.
- [X] I ran the relevant tests locally and they pass.
- [X] I formatted the code and ran the relevant linters.


Yaxuan-w (approved):

Docker/Dockerfile.e2e
```diff
@@ -52,6 +52,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
         python3 \
         python3-pip \
         sed \
+        sudo \
```
vidyalakshmir: Is there a use case for `sudo` in release builds?

Uk-jake: When running app tests with release image as a base image, sudo is required. I checked the app test script, and it calls commands using sudo directly

vidyalakshmir (approved):

